PR #82, opened 2020-10-07
bug: Force decoder_length to be integer type in TimeSeriesDataset

emigre459:
This ensures that it can be used directly or used to derive an integer value that can then be used to slice Tensors without error

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -278,7 +278,7 @@ class TimeSeriesDataSet(Dataset):
             ), "encoder_length is a protected column and must not be present in data"
             if "encoder_length" not in self.time_varying_known_reals and "encoder_length" not in self.reals:
                 self.static_reals.append("encoder_length")
-            data["encoder_length"] = 0.0  # dummy - real value will be set dynamiclly in __getitem__()
+            data["encoder_length"] = 0  # dummy - real value will be set dynamiclly in __getitem__()
 
         # validate
         self._validate_data(data)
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/data/timeseries.py
```diff
@@ -278,7 +278,7 @@ def __init__(
             ), "encoder_length is a protected column and must not be present in data"
             if "encoder_length" not in self.time_varying_known_reals and "encoder_length" not in self.reals:
                 self.static_reals.append("encoder_length")
-            data["encoder_length"] = 0.0  # dummy - real value will be set dynamiclly in __getitem__()
```
jdb78, line 281: Good catch!

pytorch_forecasting/data/timeseries.py
```diff
@@ -869,16 +869,18 @@ def __getitem__(self, idx: int) -> Tuple[Dict[str, torch.Tensor], torch.Tensor]:
...
+                    self.max_prediction_length,
+                    sequence_length - self.min_encoder_length,
+                ))
+
         encoder_length = sequence_length - decoder_length
         assert (
             decoder_length >= self.min_prediction_length
         ), "Decoder length should be at least minimum prediction length"
         assert encoder_length >= self.min_encoder_length, "Encoder length should be at least minimum encoder length"
+        assert isinstance(encoder_length, int), f"encoder_length = {encoder_length}, must be of integer type"
```
jdb78, line 883: I suggest we do the assert in the `__init__` method. Let's do it for encoder length and decoder_length, minimum and maximum.

pytorch_forecasting/data/timeseries.py
```diff
@@ -869,16 +869,18 @@ def __getitem__(self, idx: int) -> Tuple[Dict[str, torch.Tensor], torch.Tensor]:
             sequence_length >= self.min_prediction_length
         ), "Sequence length should be at least minimum prediction length"
         # determine prediction/decode length and encode length
-        decoder_length = min(
-            time[-1] - (self.min_prediction_idx - 1),
-            self.max_prediction_length,
-            sequence_length - self.min_encoder_length,
-        )
+        decoder_length = int(min(
```
jdb78, line 872: I would rather not cast this here because the user could still pass floats and then they are unpredictably converted/rounded to integers. Better to check at initialization that passed values are of the expected format.

jdb78 (changes requested):
Thanks a lot for taking this on! I suggest we check for integers at the beginning to warn the user as early as possible and simplify debugging.

jdb78:
Closing in favour of #95

PR #125, opened 2020-10-27
Removed duplicate forward pass for decoder LSTM

julioasotodv:
There were two forward passes for the decoder LSTM, which waste (actually a lot) of CPU/GPU cycles.

### DIFF
--- a/pytorch_forecasting/models/temporal_fusion_transformer/__init__.py
+++ b/pytorch_forecasting/models/temporal_fusion_transformer/__init__.py
@@ -506,14 +506,6 @@ class TemporalFusionTransformer(BaseModel, CovariatesMixin):
             (hidden, cell),
         )
 
-        decoder_output, _ = rnn.pad_packed_sequence(decoder_output, batch_first=True)
-
-        # run local decoder
-        decoder_output, _ = self.lstm_decoder(
-            embeddings_varying_decoder,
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/models/temporal_fusion_transformer/__init__.py
```diff
@@ -506,14 +506,6 @@ def forward(self, x: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
             (hidden, cell),
         )
 
-        decoder_output, _ = rnn.pad_packed_sequence(decoder_output, batch_first=True)
```
jdb78, line 509: I think this line is still required

jdb78 (changes requested):
Thanks for catching that! Indeed a waste of resources to run it twice.

pytorch_forecasting/models/temporal_fusion_transformer/__init__.py
julioasotodv: Yep, removed it by error :)

jdb78:
Think there is an issue with the linter package that is used - will fix it at a later point

julioasotodv:
So, apparently the current linter does dot work with forks :) but it does not complain

PR #137, opened 2020-11-03
BaseModel.predict() properly moves inputs to device

JakeForsey:
### Fixes

- Fixes issue bug when trying to make predictions with model / data that are not on the same device

https://stackoverflow.com/questions/59560043/what-is-the-difference-between-model-todevice-and-model-model-todevice


jdb78:
Thanks for spotting this!

PR #150, opened 2020-11-12
Add the optimizer options and default to BaseModel doc string

JakeForsey:
Working with this package has been a pleasure!

The only major snag I've had so far is selecting a sensible LR, I did not realise that the optimiser was ranger, this might have helped me a bit.


jdb78:
Thanks!! The docs can certainly improved. A very welcome addition!

PR #153, opened 2020-11-13
Feature/support gru cells in deepar

JakeForsey:
- DeepAR now supports GRU cells

This helped a lot on my problem, if the implementation does not fit your design philosophy let me know, I'm more than happen to make changes.


jdb78:
What a great addition! Could you add some more docstrings to the module to explain what is going on? I think it ready to merge otherwise.

JakeForsey:
Done

jdb78 (approved):

PR #160, opened 2020-11-17
Feature/beta distribution metric

JakeForsey:
Attempt at implementing Beta distribution for data in range [0, 1].

I could use some thoughts on:
 * Changes to GroupNormalizer and TorchNormalizer
 * New Normalizer

NOTE: Test is not yet finished (need a normalizer that makes sense....)

### DIFF
--- a/pytorch_forecasting/metrics.py
+++ b/pytorch_forecasting/metrics.py
@@ -837,32 +837,27 @@ class LogNormalDistributionLoss(DistributionLoss):
 
 
 class BetaDistributionLoss(DistributionLoss):
-    """
-    Log-normal loss, e.g. for data in the range [0.0, 1.0]
-
-    Requirements for original target normalizer:
-        * not centered normalization
-        * coerced to be positive
-        * not normalized in log space
-    """
-
+    """Loss function for DeepAR"""
     distribution_class = distributions.Beta
-    distribution_arguments = ["concentration0", "concentration1"]
+    distribution_arguments = ["mean", "shape"]
 
     def map_x_to_distribution(self, x: torch.Tensor) -> distributions.Beta:
-        return self.distribution_class(concentration0=x[..., 0], concentration1=x[..., 1])
+        mean = x[..., 0]
+        shape = x[..., 1]
+        return self.distribution_class(concentration0=(1 - mean) * shape, concentration1=mean * shape)
 
     def rescale_parameters(
         self, parameters: torch.Tensor, target_scale: torch.Tensor, transformer: BaseEstimator
     ) -> torch.Tensor:
-        assert transformer.coerce_positive, "Beta distribution is only compatible with strictly positive data"
+        # assert transformer.coerce_positive, "Beta distribution is only compatible with strictly positive data"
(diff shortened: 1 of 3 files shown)


jdb78:
Couple of thoughts here:
* It might make sense to reparametrize as by mean and shape beta as α = μν, β = (1 − μ)ν where the original concentrations are parameters α and β.
* Normalization should probably be a deterministic/not-learned transformation to scale data to 0-1. As the beta distribution as bounds, the user should know upfront how to normalize. This is could be done before feeding to the timeseries dataset by the user. Then, you can use the standard target normalizers - no need to modify them.
* Finally, using the target scale (mean, std), in the metric's `rescale_parameters()` method, you could try μ_new= sigmoid(μ + mean) and ν_new = softplus(ν) * μ_new * (1- μ_new) / std**2  (just a quick suggestion, there might be better ways to to calculate μ_new and ν_new). From  μ_new and ν_new, you can calculate α and β to create the distribution from which you can sample.
As always, [the wikipedia article on the Beta distribution](https://en.wikipedia.org/wiki/Beta_distribution) is helpful for reparametrizations.

I hope this is helpful!

JakeForsey:
I've done the re-parameterisation and my model is training well. 

I'm not sure how to write a proper tests for it, in a bit over my head.

pytorch_forecasting/metrics.py
```diff
@@ -835,3 +835,37 @@ def rescale_parameters(
...
+    def rescale_parameters(
+        self, parameters: torch.Tensor, target_scale: torch.Tensor, transformer: BaseEstimator
+    ) -> torch.Tensor:
+        assert transformer.coerce_positive, "Beta distribution is only compatible with strictly positive data"
+        assert (
+            not transformer.log_scale
+        ), "Beta distribution is not compatible with log transformation - use LogNormal"
+        assert not transformer.center, "Beta distribution is not compatible with centered data"
+
+        scaled_mean = torch.sigmoid(parameters[..., 0] + target_scale[..., 0].unsqueeze(1))
```
jdb78: I wonder if it works even better if you take `torch.logit(target_scale[..., 0])` instead of `target_scale[..., 0]`.

tests/test_metrics.py
```diff
@@ -14,6 +14,7 @@
     LogNormalDistributionLoss,
     NegativeBinomialDistributionLoss,
     NormalDistributionLoss,
+    BetaDistributionLoss,
```
jdb78: Could you run black and isort (best is to use the pre-commit hooks to ensure correct formatting)?

jdb78 (changes requested):
Looks really good! Thanks a lot for all the hard work! Just minimal formatting changes and a suggestion on the parametrization. How are you training? Maybe you could write some documentation as to how it can be used by a user and what they should take consider.

jdb78:
Integrated the beta distribution loss - should work now

jdb78 (approved):

JakeForsey:
This is awesome, much better than what I was messing around with.

PR #173, opened 2020-11-24
Bugfix/basemodel handles non tensor raw predictions

JakeForsey:
DeepAR output contains a string value which can not be concatenated with torch.cat(). This MR resolves this by only concatenating values which are torch.Tensors

### DIFF
--- a/pytorch_forecasting/models/base_model.py
+++ b/pytorch_forecasting/models/base_model.py
@@ -653,7 +653,8 @@ class BaseModel(LightningModule):
             output = torch.cat(output, dim=0)
         elif mode == "raw":
             output_cat = {}
-            for name in output[0].keys():
+            tensor_keys = [key for key, item in output[0].items() if isinstance(item, torch.Tensor)]
+            for name in tensor_keys:
                 output_cat[name] = torch.cat([out[name] for out in output], dim=0)
             output = output_cat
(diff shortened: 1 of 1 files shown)


jdb78:
We could instead use different functions for concatenations? Maybe np.concat for non-tensors?

jdb78:
Great addition, btw!

JakeForsey:
> We could instead use different functions for concatenations? Maybe np.concat for non-tensors?

Like this?

```
            for name in output[0].keys():
                v0 = output[0][name]
                if isinstance(v0, torch.Tensor):
                    output_cat[name] = torch.cat([out[name] for out in output], dim=0)
                elif isinstance(v0, np.ndarray):
                    output_cat[name] = np.concatenate([out[name] for out in output])
                else:
                    output_cat[name] = [out[name] for out in output]
```

Not sure if along which axis I should be concatenating numpy arrays

jdb78:
Yes. This could work. Think by default along `axis=0`.

JakeForsey:
> Yes. This could work. Think by default along `axis=0`.

isinstance(v0, np.ndarray) is not enough to ensure that np.concatenate works (zero-dimensional arrays and strings both through errors) so this is falling back to list comprehension.

PR #184, opened 2020-12-01
DeepAR correctly calculates target position

JakeForsey:
I think this might resolve https://github.com/jdb78/pytorch-forecasting/issues/182#issue-753749841

### DIFF
--- a/pytorch_forecasting/models/deepar/__init__.py
+++ b/pytorch_forecasting/models/deepar/__init__.py
@@ -193,9 +193,7 @@ class DeepAR(AutoRegressiveBaseModelWithCovariates):
 
     @property
     def target_position(self):
-        variables = self.hparams.time_varying_reals_encoder  # todo: support categorical targets
-        pos = variables.index(self.hparams.target) + len(self.hparams.static_reals)
-        return pos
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/models/deepar/__init__.py
```diff
@@ -194,7 +194,7 @@ def construct_input_vector(
     @property
     def target_position(self):
         variables = self.hparams.time_varying_reals_encoder  # todo: support categorical targets
-        pos = variables.index(self.hparams.target)
+        pos = variables.index(self.hparams.target) + len(self.hparams.static_reals)
```
jdb78: This might not yield the right result because the variables are not necessarily directly mapping to the dataset inputs. What will always work is `self.hparams.x_reals.index(self.hparams.target)`.

jdb78 (changes requested):
Thanks a lot for detecting this - makes sense that this did not work before.

jdb78 (approved):

JakeForsey:
> Thanks a lot for detecting this - makes sense that this did not work before.

Its a shame! I will have to go back to predicting t+n = t0 :D

PR #203, opened 2020-12-14
Updated example ar.py

JustinNeumann:
### New features

- Actually uses the DeepAR model now with minimal changes to the old version of ar.py


jdb78:
Thanks for this contribution, @JustinNeumann. Could you run the pre-commit hooks? `pre-commit install` and `pre-commit run --all-files`. This way we can fix the remaining non-passing checks.

jdb78 (approved):
Looks really good! Thanks for adding the example

JustinNeumann:
Thanks! I can now celebrate my first pull request 💯

PR #223, opened 2020-12-28
Fixed self.norm_ in GroupEncoder in fit method

reumar:
I think you forgot to add an underscore to self.norm_ atribute of the GroupEncoder.


jdb78:
Creat catch!

jdb78 (approved):

PR #234, opened 2021-01-02
Fix marry vs merry

rustyconover:
### Description

Fixes a verb in the documentation.


jdb78 (approved):
Thanks for spotting the typo!

PR #268, opened 2021-01-18
Handling `return_x=True` in case of multi-target prediction

snumumrik:
- Handling multi-target prediction case with return_x=True.
- Concatenation of multi-target prediction output by target instead of batch. (little bug fix)


jdb78:
Thanks a lot for this PR! Will give it a go and actually write a test for the `predict` output.

PR #549, opened 2021-06-05
TCN of pytorch-forecasting

wangbingnan136:


pytorch_forecasting/models/tcn/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+                if name in self.encoder_variables or name in self.static_variables
+            ],
+            dim=-1,
+        )
+        prediction = self.network(network_input.permute(0,2,1))
+        
+
+        # We need to return a dictionary that at least contains the prediction and the target_scale.
+        # The parameter can be directly forwarded from the input.
+        return dict(prediction=prediction, target_scale=x["target_scale"])
```
jdb78: Could you use the new way to return de-normalized predictions?
```suggestion
        prediction = self.transform_output(prediction, target_scale=x["target_scale"])
        return self.to_network_output(prediction=prediction)
```

pytorch_forecasting/models/tcn/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        self.network = nn.Sequential(*layers)
+
+    def forward(self, x):
+        return self.network(x)
+
+
+
+
+class TCN(BaseModelWithCovariates):
+    def __init__(
```
jdb78: Could you add a docstring that also cites the paper on which the network is based?

jdb78 (commented):
Great addition to the package! Thanks a lot for all the work!

In order to get this PR over the line:

Could you add some documentation:
- [ ] docstrings
- [x] model documentation in the models.rst
- [x] model mentioning in Readme.md
- [x] link the tutorial in the tutorials.rst

add some tests (see, for example, test_nbeats.py):
- [ ] integration test
- [ ] pickle test

wangbingnan136:
Ok,I will do that as soon as I can!

wangbingnan136:
I have modified TCN according to your suggestion,thanks a lot!

wangbingnan136:
I'm sorry.I do not know how the checks above works? Can you give me some complete examples of code testing?So that I can do the test myself.

jdb78:
In the directory `tests/test_models` are multiple files that test a network. I suggest to pretty much copy paste the `test_nbeats.py`. as this is very similar to what you are trying to do.

jdb78:
There seems to be an import error: https://github.com/jdb78/pytorch-forecasting/pull/549/checks?check_run_id=2783133836

wangbingnan136:
I have fixed the import error

jdb78:
@wangbingnan136: How do we approach the multi-horizon forecasting issue with TCNs? Is it supposed to be re-current or a multi-horizon forecast? In case of the later, it is a bit unclear to me how to leverage covariates in the decoder.

wangbingnan136:
> @wangbingnan136: How do we approach the multi-horizon forecasting issue with TCNs? Is it supposed to be re-current or a multi-horizon forecast? In case of the later, it is a bit unclear to me how to leverage covariates in the decoder.

Tcn can only support the output mode of **vector output**, 

image

which I think is what you mean "re-current," that is, TCN only has an encoder structure and no decoder structure. We use the latent vector obtained by the encoder to directly connect to a fully connected layer to complete the forecasting task. 

Wavenet supports the structure of encoder+decoder.

jdb78:
@wangbingnan136 Could you check the implementation? I am not really sure it is correct. Adding a skip connection (optional in the paper) dramatically improved the performance. Could you check it and confirm the implementation is correct before we merge?

wangbingnan136:
Ok,I will see

wangbingnan136:
> @wangbingnan136 Could you check the implementation? I am not really sure it is correct. Adding a skip connection (optional in the paper) dramatically improved the performance. Could you check it and confirm the implementation is correct before we merge?

I'm a little confused,but I have implemented the skip connection in here:

image

wangbingnan136:
I have implemented the skip connection function according to https://github.com/philipperemy/keras-tcn/blob/master/tcn/tcn.py

and test the data in tutorial:
image

The performance of model becomes better.

jdb78:
@wangbingnan136 , could you revert the version to after the lock-file fix (best to use `git pull` before continuing the work) and use the pre-commit hooks?

I believe the main issue is that M-TCN uses a slightly modified architecture and currently, the model uses convolutions along features and not time.

wangbingnan136:
> @wangbingnan136 , could you revert the version to after the lock-file fix (best to use `git pull` before continuing the work) and use the pre-commit hooks?
> 
> I believe the main issue is that M-TCN uses a slightly modified architecture and currently, the model uses convolutions along features and not time.

Ok,I have reverted the version.

wangbingnan136:
image

I think the model uses convolutions along time.

StatMixedML:
@wangbingnan136 Many thanks for making the TCN model available. However, I am not sure which version of the following you are actually implementing:

- [Probabilistic forecasting with temporal convolutional neural network](https://www.sciencedirect.com/science/article/abs/pii/S0925231220303441): Preferred one since this outputs and trains on quantile losses
- [A Multi-Horizon Quantile Recurrent Forecaster](https://arxiv.org/pdf/1711.11053.pdf): Available in GluonTS 
- [Multivariate Temporal Convolutional Network: A Deep Neural Networks Approach for Multivariate Time Series Forecasting](https://www.mdpi.com/2079-9292/8/8/876)

May I ask you to be more specific on which one you are using.

wangbingnan136:
> @wangbingnan136 Many thanks for making the TCN model available. However, I am not sure which version of the following you are actually implementing:
> 
> * [Probabilistic forecasting with temporal convolutional neural network](https://www.sciencedirect.com/science/article/abs/pii/S0925231220303441): Preferred one since this outputs and trains on quantile losses
> * [A Multi-Horizon Quantile Recurrent Forecaster](https://arxiv.org/pdf/1711.11053.pdf): Available in GluonTS
> * [Multivariate Temporal Convolutional Network: A Deep Neural Networks Approach for Multivariate Time Series Forecasting](https://www.mdpi.com/2079-9292/8/8/876)
> 
> May I ask you to be more specific on which one you are using.

Actually it is almost the same as https://github.com/philipperemy/keras-tcn.
I think the related paper is this one:https://arxiv.org/pdf/1803.01271.pdf 《An Empirical Evaluation of Generic Convolutional and Recurrent Networks for Sequence Modeling》

StatMixedML:
> Actually it is almost the same as https://github.com/philipperemy/keras-tcn.
> I think the related paper is this one:https://arxiv.org/pdf/1803.01271.pdf 《An Empirical Evaluation of Generic Convolutional and Recurrent Networks for Sequence Modeling》

Ok I see, many thanks for clarifying!

jdb78:
Actually, seeing this, we should probably switch to **Probabilistic forecasting with temporal convolutional neural network** as it has a proper decoder (The quantile loss is trivial and would be supported anyways).

wangbingnan136:
> Actually, seeing this, we should probably switch to **Probabilistic forecasting with temporal convolutional neural network** as it has a proper decoder (The quantile loss is trivial and would be supported anyways).

I see.I may need some time to study the network structure of the tcn in this paper。

StatMixedML:
> Actually, seeing this, we should probably switch to **Probabilistic forecasting with temporal convolutional neural network** as it has a proper decoder (The quantile loss is trivial and would be supported anyways).

@jdb78 I would also opt for this version. Would love to be part of supporting the implementation.

Issue #571, opened 2021-06-18
[QUESTION] Custom Loss flow on Nbeats


pnmartinez:
Hi.

Trying to implement a **custom loss function** patching the default `MASE()` object (which inherits from `MultiHorizonMetric()` and `Metric()`).

I am debugging the `loss` computation to hunt down this little bug, but **I cannot see where to retrieve the actual scalar value logged into Tensorboard.**

Where I am:

- I see that it is the `_update_losses_and_lengths()` method in the `MultiHorizonMetric()` is the one **reducing** **the loss tensor ([batch_size, pred_len]) to a scalar ([])**,
- but **when I print the value** at the end of that function, they are always **huge values** (~ 10⁷, 10^9) instead of the value logged, of order ~10^0 or 10^1 at most.

Getting kind of crazy right here, so:

- Could any of you please **point me to the next steps in the logic** after  `_update_losses_and_lengths()`?

jdb78:
Can you post some code? Maybe even contribute the loss function as a PR?

pnmartinez:
Hi @jdb78 ,

PR done. You can see there that **changes are minimal**, only adding some extra steps to the `loss()` in the `MASE()` object.

I also elaborate a bit in the PR description: 
- **my end problem** is that I randomly get `RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn` while training.
- **That's the reason I am tracking down** how all the codebase deals with the tensors, reductions, etc.

Cheers!

jdb78:
Did the comments on the PR help? If your first value is 0, you divide by 0.

pnmartinez:
> Did the comments on the PR help? If your first value is 0, you divide by 0.

Sorry Jan, but I can't find any comments on the PR.

If you mean the `first_value` column tensor, there's not a single 0.0 in the dataset (nor negative values). Values are small, in the order of 10⁻², but do not go closer to 0, ever.

jdb78:
Do you have a collab notebook? Somewhere, clearly there is something causing the computation to yield `nan`. I suggest to go through it with a debugger and find out.

TimothyLiuu:
Hello! Have you solve the problem? @pnmartinez

PR #576, opened 2021-06-21
Custom MASE loss

pnmartinez:
### Description

Submitting the **problematic custom MASE loss mentioned** on #571. 

The question is about **the loss reduction** from tensor to scalar: there are so many intermediate calls on different API levels that I am not able to retrieve manually the loss values I am getting.

My end purpose is to **fix this bug**, which I am getting on random epochs while training (sometimes I can train 1 epoch before crash, others 100..):
```py
```

### Changelog

- `metrics.py/MASE`   -->  `metrics.py/CustomMASE`.

### DIFF
--- a/pytorch_forecasting/metrics.py
+++ b/pytorch_forecasting/metrics.py
@@ -799,10 +799,9 @@ class RMSE(MultiHorizonMetric):
         return loss
 
 
-class MASE(MultiHorizonMetric):
+class CustomMASE(MultiHorizonMetric):
     """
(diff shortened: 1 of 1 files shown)


jdb78:
If your std is not very stable, MASE can lead very quickly to exploding / nan gradients. @pnmartinez , have you been able to fix the issue?

jdb78:
Closing as used for bug fixing. @pnmartinez , feel free to re-open.

PR #672, opened 2021-09-13
fix infinite loop in TimeDistributedEmbeddingBag

eavae:
### Description

This PR ...

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


jdb78:
Good catch. Thank you!

Issue #680, opened 2021-09-18
MultiNormalizer always raise NotFittedError, even when it was actually fitted


chefPony:
- PyTorch-Forecasting version:
- PyTorch version: 1.9.0
- Python version: 3.9
- Operating System: OSX

### Expected behavior

When all normalisers embedded in a MultiNormalizer are fitted also the MultiNormalizer itself should be considered fitted, and hence do not raise a NotFittedError when passed to check_is_fitted function.

### Actual behavior

MultiNormalizer always raise NotFittedError

### Code to reproduce the problem

```
import pandas as pd
import numpy as np
from sklearn.utils.validation import check_is_fitted
from pytorch_forecasting.data import TimeSeriesDataSet, MultiNormalizer, TorchNormalizer


train_data = pd.DataFrame({
    "time_idx": np.concatenate([np.arange(0, 15, 1), np.arange(0, 15, 1)]),
    "group": np.concatenate([np.zeros((15, ), dtype=int), np.ones((15, ), dtype=int)]),
    "feat_0": np.random.rand(30),
    "y_0": np.random.rand(30),
    "y_1": np.random.rand(30),
})

train_data.group = train_data.group.astype(str)

train_ts_dataset = TimeSeriesDataSet(
    train_data, time_idx="time_idx", target=["y_0", "y_1"], group_ids=["group"], 
    max_encoder_length=3, min_encoder_length=3, max_prediction_length=2, min_prediction_length=2,
    time_varying_known_reals=["feat_0"], 
    target_normalizer= MultiNormalizer([TorchNormalizer(), TorchNormalizer()]),
    add_target_scales=True
)

check_is_fitted(train_ts_dataset.target_normalizer.normalizers[0])    #OK
check_is_fitted(train_ts_dataset.target_normalizer.normalizers[1])    #OK

check_is_fitted(train_ts_dataset.target_normalizer)  #KO raises not fitted error

---------------------------------------------------------------------------
NotFittedError                            Traceback (most recent call last)
<ipython-input-37-21b62a367f65> in <module>
----> 1 check_is_fitted(train_ts_dataset.target_normalizer)

~/Library/Caches/pypoetry/virtualenvs/pytorch-forecasting-P4jsLfeu-py3.9/lib/python3.9/site-packages/sklearn/utils/validation.py in inner_f(*args, **kwargs)
     61             extra_args = len(args) - len(all_args)
     62             if extra_args <= 0:
[...]

PR #681, opened 2021-09-18
- add fitted_ attribute to MultiNormalizer when fit is called

chefPony:
### Description

This PR solves issue #680 

### Checklist

- [x] Linked issues (if existing) #680 
- [x] Added/modified tests


jdb78:
Good catch. Thank you!

PR #732, opened 2021-11-05
Fix inattention mutation to `x_cont`.

eavae:
### Description

This PR fixed an issue about inattention mutation with `x_cont` that caused original data changed.

The mutation happened here:

```python
input_vector[..., self.target_positions] = torch.roll(
    input_vector[..., self.target_positions], shifts=1, dims=1
)
```

It caused `x_cont` changed. And the changes keep exists from `encode` to `decode`. So, the shift happen twice. 

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Release Notes
 
+## v0.9.2 UNRELEASED
+
+### Added
+
+### Fixed
+
+- Fix inattention mutation to `x_cont`. 
+
+### Contributors
+
+- eavae
+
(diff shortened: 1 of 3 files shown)


jdb78:
Great catch @eavae. Do you want to add an entry to the changelog and add your name / user name?

jdb78 (dismissed):

eavae:
@jdb78 CHANGELOG updated.

jdb78 (approved):

PR #740, opened 2021-11-09
Fixed some typos.

bsircelj:
What confused me a bit is the formula for Quantile loss. The actual implemented method uses the correct formula, only the one in the documentation is wrong, so I fixed it.

### DIFF
--- a/pytorch_forecasting/metrics.py
+++ b/pytorch_forecasting/metrics.py
@@ -652,7 +652,7 @@ class QuantileLoss(MultiHorizonMetric):
     """
     Quantile loss, i.e. a quantile of ``q=0.5`` will give half of the mean absolute error as it is calculated as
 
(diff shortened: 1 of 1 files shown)


jdb78:
is it not correct, as q * (**y-y_pred**), (1-q) * (**y_pred-y**)?

jdb78:
But agree, confusing as different in implementation. Maybe you want to replace `(1-q) * (y_pred-y)` with `(q-1) * (y-y_pred)`?

bsircelj:
Ah you are right. I didn't notice that the y and y_pred terms are switched. The code I was checking is few lines below:
```
            errors = target - y_pred[..., i]
            losses.append(torch.max((q - 1) * errors, q * errors).unsqueeze(-1))
```

PR #744, opened 2021-11-11
Update metrics.py

stllfe:
Fixing the signature for the base metric class

### Description

This PR just adds a tiny `self` argument to the base `Metric.update` method. It's still getting overridden in all the other metrics, but still, IDEs give the annoying warning for overriding the base signature.


jdb78 (approved):
Good find!

PR #783, opened 2021-12-02
Fix pytorch warning when converting lists of ndarrays

Seon82:
### Description

This PR removes a new pytorch warning that occurs when converting a list of ndarrays to a tensor, and which happens on every batch if target_scale is a ndarray.

Full warning here:
```UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at  ../torch/csrc/utils/tensor_new.cpp:201.)```

The change shouldn't have any negative effect if `[batch[0]["target_scale"]` isn't an `ndarray`.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -1654,7 +1654,7 @@ class TimeSeriesDataSet(Dataset):
                     scale = torch.tensor([batch[0]["target_scale"][idx] for batch in batches], dtype=torch.float)
                 target_scale.append(scale)
         else:  # convert to tensor
-            target_scale = torch.tensor([batch[0]["target_scale"] for batch in batches], dtype=torch.float)
+            target_scale = torch.from_numpy(np.array([batch[0]["target_scale"] for batch in batches], dtype=np.float))
 
         # target and weight
         if isinstance(batches[0][1][0], (tuple, list)):

### CODE REVIEW

File: pytorch_forecasting/data/timeseries.py [hunk 1]
```diff
@@ -1654,7 +1654,7 @@ def _collate_fn(
                     scale = torch.tensor([batch[0]["target_scale"][idx] for batch in batches], dtype=torch.float)
                 target_scale.append(scale)
         else:  # convert to tensor
-            target_scale = torch.tensor([batch[0]["target_scale"] for batch in batches], dtype=torch.float)
+            target_scale = torch.tensor(np.array([batch[0]["target_scale"] for batch in batches]), dtype=torch.float)
```
jdb78: Could you use `torch.from_numpy`? This should allow it to share memory and speed things up.

jdb78 (changes requested):
Great addition. Missed this one.

pytorch_forecasting/data/timeseries.py
Seon82: Done

jdb78 (dismissed):

jdb78:
Fixed by #817

PR #806, opened 2021-12-21
Improve performance of __getitem__ of TimeSeriesDataSet

denix56:
### Description

Pandas DataFrame is quite slow in comparison to numpy due to additional checks.
By replacing it with np.recarray I was able to improve performance by 5-10%.
Recarray allows us to have nice attribute access as in pandas, while improving performance.
The raw numpy arrays are a bit faster than recarray, however the difference is not as big as between pandas and recarray.
I have tested on [Demand Forecasting](https://pytorch-forecasting.readthedocs.io/en/latest/tutorials/stallion.html) with gpu=1, 0 workers and pin_memory=True.

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -1247,7 +1247,7 @@ class TimeSeriesDataSet(Dataset):
             len(df_index) > 0
         ), "filters should not remove entries all entries - check encoder/decoder lengths and lags"
 
-        return df_index
+        return df_index.to_records(index=False)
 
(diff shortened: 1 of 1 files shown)


jdb78:
I am tempted to merge this. Think we should run the example notebooks also because things might change there - even if only visual.

jobs-git:
any news on this?

fkiraly (changes requested):
It appears the changes in this PR break internal API assumptions in other methods, e.g., `get_groups` - so it cannot be merged in its current state.

Still worth to keep open as long as we are reworking for v2.

PR #816, opened 2022-01-02
Implement logger adapter to handle different loggers (TensorBoard and Wandb for now)

### LINKED ISSUE #89: Expand figure logging for multiple logging platforms
Following up on the discussion from issue #79, there is a need to extend the logging capabilities for keeping track of training figures (e.g. showing attention, forecast quantiles, etc.) for loggers beyond tensorboard (e.g. W&B). The `self.logger.experiment.add_figure()` lines that exist in `models.base_model` seem to be the route of the issue, as not every logger platform has an `add_figure()` method for its `experiment` (or `Run` in W&B's case) objects.

This is a to-do item for now. Can currently be circumvented (at least in the case of using W&B) by setting `log_interval=-1` in the instantiation for `TemporalFusionTransformer` object.

mikcnt:
### Description

This PR resolves #89.
* I created a base logger class (`loggers.ForecastingLoggerBase`) to handle different loggers (e.g., TensorBoard, Wandb) using the same syntax. This inherits from the PyTorch Lightning logger base class, and implements three new methods used across the PyTorch Forecasting project: `add_figure`, `add_embeddings`, and `log_gradient_flow`.
* I implemented two new logger classes for TensorBoard and Wandb based on `loggers.ForecastingLoggerBase`, and inheriting from their PyTorch Lightning counterparts (respectively `pytorch_lightning.loggers.TensorBoardLogger` and `pytorch_lightning.loggers.WandbLogger`).
* I modified the tests, replacing all the `pytorch_lightning.loggers.TensorBoardLogger` instances with the newly created `loggers.ForecastingTensorBoardLogger`.

As for the Wandb logger, `loggers.ForecastingWandbLogger`, I was wondering if someone could guide me on how to structure its tests.

### Checklist

- [x] Added/modified tests

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -293,7 +293,7 @@ unicode_backport = ["unicodedata2"]
 name = "click"
 version = "8.0.3"
 description = "Composable command line interface toolkit"
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=3.6"
 
@@ -385,6 +385,18 @@ python-versions = "*"
 test = ["flake8 (==3.7.8)", "hypothesis (==3.55.3)"]
 
 [[package]]
+name = "configparser"
+version = "5.2.0"
+description = "Updated configparser from Python 3.8 for Python 2.6+."
+category = "main"
+optional = false
+python-versions = ">=3.6"
+
+[package.extras]
+docs = ["sphinx", "jaraco.packaging (>=8.2)", "rst.linker (>=1.9)", "jaraco.tidelift (>=1.4)"]
+testing = ["pytest (>=6)", "pytest-checkdocs (>=2.4)", "pytest-flake8", "pytest-cov", "pytest-enabler (>=1.0.1)", "types-backports", "pytest-black (>=0.3.7)", "pytest-mypy"]
+
+[[package]]
 name = "coverage"
 version = "6.2"
 description = "Code coverage measurement for Python"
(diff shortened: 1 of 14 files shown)


mikcnt:
Yep, definitely. I'll give them a look tonight.

Do you also have any tip on how to structure tests for the logger adapter?

mikcnt:
I've updated the stallion notebook so that it uses the `ForecastingTensorBoardLogger` instead of the standard `TensorBoardLogger`.

mikcnt:
Any news on this?

B-Deforce:
There actually turns out to be a nice workaround for this. 
You can add `wandb.init(project='my-project', sync_tensorboard=True)` followed by your PyTorch code and it'll nicely log everything to Wandb without issues. 

For more info see [here](https://docs.wandb.ai/guides/integrations/tensorboard).

mikcnt:
Hey @B-Deforce! That's actually pretty nice, thanks for the heads up, I didn't know about that! So, you're suggesting to replace the `ForecastingWandbLogger` with the standard logger used in forecasting, and to add the wandb initialization with sync to tensorboard?

B-Deforce:
Yes @mikcnt, I was actually also struggling to get Wandb running with Pytorch Forecasting and used your implementation above (which worked great btw). But then I found this neat little feature of wandb! Yes, that's correct. Just using the standard logger (which is the `TensorBoardLogger`) with the wandb init + sync works great.

mikcnt:
That's terrific! Actually, my original plan was to include a large variety of loggers, not just Tensorboard or Wandb (e.g., MLFlow), but, since it appears like this won't be merged anytime soon, I guess I would go with your solution too if I were in your position! 😆

PR #817, opened 2022-01-03
Fix warning in timeseries collate function

mikcnt:
### Description

This PR fixes this warning given by PyTorch in the collate function of the timeseries data script:

```
../anaconda3/envs/env/lib/python3.8/site-packages/pytorch_forecasting/data/timeseries.py:1654:
UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow.
Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor.
(Triggered internally at  ../torch/csrc/utils/tensor_new.cpp:201.)
--> scale = torch.tensor([batch[0]["target_scale"][idx] for batch in batches], dtype=torch.float)
```

The list of numpy arrays is casted into a single numpy array before being casted to a torch tensor.

Notice that there is probably a small bug in the `log_gradient_flow` method of the `BaseModel`, for which some tests don't pass. More details on that can be found in the last note of PR #816. Replacing `p.grad` with `p.grad.cpu()` solves the issue and makes all tests pass, but a better solution is proposed in #816. 

### Checklist

- [ ] Added/modified test

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -1651,8 +1651,8 @@ class TimeSeriesDataSet(Dataset):
                 if isinstance(batches[0][0]["target_scale"][idx], torch.Tensor):  # stack tensor
                     scale = torch.stack([batch[0]["target_scale"][idx] for batch in batches])
                 else:
-                    scale = torch.tensor(
-                        np.array([batch[0]["target_scale"][idx] for batch in batches]), dtype=torch.float
+                    scale = torch.from_numpy(
+                        np.array([batch[0]["target_scale"][idx] for batch in batches], dtype=np.float32),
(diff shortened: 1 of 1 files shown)


jdb78:
Can we use a `torch.from_numpy` function? Should be more efficient but seems somehow to break things. See #783

mikcnt:
Yes, that's definitely possible.

As for the fact that `torch.from_numpy` was breaking the other PR, that's because there is a basic misunderstanding between torch dtypes and numpy dtypes: `torch.float` is an alias for `torch.float32`, while `np.float` is an alias for `np.float64`. That's why I changed `np.float32` inside the `torch.from_array` call, and that seems to work.

jdb78 (approved):

PR #842, opened 2022-01-24
Fix float errors

eugeneteoh:
### Description

This PR ...

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


jdb78:
Could you give an example? Would love to have a test to understand the problem better. Probably not ideal to just cast to float.

PR #846, opened 2022-01-27
fixed issue with calculate_prediction_actual_by_variable not working with int decoder_targets

DeastinY:
### Description

This PR converts x["decoder_target"] to double to ensure torch.mean() works.


jdb78:
If the target is actually a categorical, we should probably handle things a little bit differently and calculate a mean per category instead of binning. Would be keen to review a PR addressing this!

PR #858, opened 2022-02-04
Possible fix for adding new group ids when using from_dataset

eirikfjaerbu:
I am trying to use the TFT forecasting algorithm for the first time. I may have misunderstood the intended use of this library, but my idea is as follows: first make a training TimeSeriesDataSet using the class constructor, train the TFT on that, and then create a test dataset with the same transformations as the training set using 'from_dataset'. Since the test set contains independent data from the training set, most of the group ids in the test set are not present in the training set. 

When I try to run 'from_dataset', I get an error saying "Unknown category '----' encountered. Set `add_nan=True` to allow unknown categories" with one of my group ids where the bars go. When I looked at the code, it seemed like this was supposed to work, and I was able to run my code after doing a minor modification.

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -660,7 +660,7 @@ class TimeSeriesDataSet(Dataset):
             group_ids_to_encode = self.group_ids
         else:
             group_ids_to_encode = []
-        for name in dict.fromkeys(group_ids_to_encode + self.categoricals):
+        for name in dict.fromkeys(self.categoricals):
             if name in self.lagged_variables:
                 continue  # do not encode here but only in transform
             if name in self.variable_groups:  # fit groups
@@ -688,7 +688,11 @@ class TimeSeriesDataSet(Dataset):
             # targets and its lagged versions are handled separetely
             if name not in self.target_names and name not in self.lagged_targets:
                 data[name] = self.transform_values(
-                    name, data[name], inverse=False, ignore_na=name in self.lagged_variables
+                    name=name,
+                    values=data[name],
+                    inverse=False,
+                    group_id=name in self._group_ids_mapping,
+                    ignore_na=name in self.lagged_variables,
                 )
 
         # save special variables


eirikfjaerbu:
I made a small code example that fails for me:

import pandas as pd
from pytorch_forecasting import TimeSeriesDataSet

training_data = pd.DataFrame(
    {
        "gid": ["1", "1", "2", "2"],
        "time": [1, 2, 1, 2],
        "real": [1., 2., 3., 4.],
        "target": [0., 0., 0., 0.]
    }
)

training_set = TimeSeriesDataSet(
    training_data,
    time_idx="time",
    target="target",
    group_ids=["gid"],
    time_varying_known_reals=["real"],
    max_encoder_length=1,
    min_encoder_length=1,
)

test_data = pd.DataFrame(
    {
        "gid": ["3", "3", "4", "4"],
        "time": [1, 2, 1, 2],
        "real": [1., 2., 3., 4.],
        "target": [0., 0., 0., 0.]
    }
)

test_set = TimeSeriesDataSet.from_dataset(
    dataset=training_set,
    data=test_data,
)

jdb78:
Thanks for pointing this out. It is better to use the EncoderNormalizer in this case for the moment. One would need to adapt the GroupNormalizer on how to deal with new groups in the test set. Unfortunately, your workaround is silenting the error but you might get some undesired behaviour: New groups will be not normalized by their statistic but by some global statistic - probably very unexpected and should lead to far worse results on the test set.

eirikfjaerbu:
Thank you very much for the reply. I tried to set 'target_normalizer=EncoderNormalizer()' in my training dataset, and I also tried to run the code from [your PR](https://github.com/jdb78/pytorch-forecasting/pull/879). In both cases, the code ran, and as expected, your PR produced a long list of warning messages. 

Sadly, I can't post the code from my main project, but I did not get the expected results:
- Encoder Normalizer: Poor accuracy on training set, same poor accuracy on test set
- Your PR: Best accuracy on the training set, poor accuracy on test set
- My PR: Good accuracy on training set, and same good accuracy on the training set

I realize this may not be very helpful since I cannot post my actual data, but I wanted to mention it anyway. The results I am getting with my PR are also close and slightly better than my benchmark algorithm, so I am not very worried that there is undesired behaviour.

jdb78:
#949

eirikfjaerbu:
> #949

Thanks for the update! The more fine-grained minimum encoder length restriction enabled me to get predictions for items with no historical data. I think this solves my issues for now, so I will close the pull request

PR #892, opened 2022-03-16
Bug fix for warning: The metric 'val_loss' is not in the evaluation logs for pruning.

ruuttt:
With this change the following warning does not occur anymore:
```
/usr/local/lib/python3.7/dist-packages/optuna/integration/pytorch_lightning.py:52: UserWarning:
The metric 'val_loss' is not in the evaluation logs for pruning. Please make sure you set the correct metric name.
```


jdb78:
Is that the right thing to do, though? Do we really want to train on a sanity_check_loss?

ruuttt:
By adding num_sanity_val_steps=0 to the trainer_kwargs in optimize_hyperparameters the error also disspaears.

Issue #998, opened 2022-05-23
Dtype issue in `TorchNormalizer(method="identity")`


RonanFR:
- PyTorch-Forecasting version: 0.10.1
- PyTorch version: 1.10.2+cu111
- Python version: 3.8.10
- Operating System: Ubuntu

I recently experienced a `dtype` issue when using `TorchNormalizer` and managed to identify the source of the bug.

When `TorchNormalizer(method="identity", center=False)` is fitted on pandas DataFrame/Series, the attributes `self.center_`and `self.scale_` are set to `np.zeros(y_center.shape[:-1])` and `np.ones(y_scale.shape[:-1])` respectively (lines 467 and 468 of file `pytorch_forecasting.data.encoders.py`).

The default `dtype` for a numpy array is `np.float64` and so both `self.center_`and `self.scale_` have this dtype independently of the `dtype` of `y_center` and `y_scale`. On the other hand, the default dtype of torch (unless specified otherwise using `torch.set_default_dtype`) is `float32` (a.k.a. `Float`, as opposed to `Double` which is equivalent to `float64`).

This may cause problems at inference time (e.g., with DeepAR): the predicted values have type `float64` while the targets in the dataloader have type `float32` (due to scaling). More specifically, at some point during inference (method `predict`), the model calls the method `self.decode_autoregressive` that iteratively fills a torch tensor using method `self.decode_one` (line 292 of file `pytorch_forecasting.models.deepar.py`). The source tensor has type float32 while the other tensor has type float64.

My suggestion would be to modify the code of `TorchNormalizer`. One possibility would be to set the type of `self.center_`and `self.scale_` to be the same as `y_center`and `y_scale` respectively. The problem is that it may be a bit weird if the type of `y_scale` is `int32` for instance (we expect the scale to be a real value rather than an integer). This happens when using NegativeBinomialLoss for instance. Alternatively, we could set the type of `self.center_`and `self.scale_` to be equal to `str(torch.get_default_dtype())[6:]` (this is the current fix I made in my code but not sure it is the best idea).

PR #1025, opened 2022-06-09
bugfix: dtype inconsistency in TorchNormalizer with method='identity'

RonanFR:
### Description

This PR aims at fixing issue #998 by making changes in method `_set_parameters` of `TorchNormalizer`.
The following assertion was added in `tests/test_data/test_encoders.py`:
```
y = np.array([1, 2, 3], dtype=np.float32)
assert TorchNormalizer(method="identity").fit(y).get_parameters().dtype == torch.float32
```
Before the changes made, this assertion would fail (the dtype would be `torch.float64` due to numpy default dtype `numpy.float64`) which then would potentially cause dtype issues (e.g., in method `decode` of DeepAR and more specifically `decode_autoregressive`: the first lagged value was of dtype torch.float32 -true observed value- while the subsequent values -predictions- were of dtype torch.float64).

Not 100% sure the problem should not have been fixed on DeepAR side but note that a similar fix of dtype was already implemented in `TorchNormalizer`: 
```
# correct numpy scalar dtype promotion, e.g. fix type from `np.float32(0.0) + 1e-8` gives `np.float64(1e-8)`
if isinstance(self.scale_, np.ndarray):
    self.scale_ = self.scale_.astype(y_scale.dtype)
```

### Checklist

- [x] Related Github issues: #998
- [x] Added/modified tests: `tests/test_data/test_encoders.py`

### DIFF
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -464,8 +464,11 @@ class TorchNormalizer(InitialParameterRepresenterMixIn, BaseEstimator, Transform
                 self.center_ = torch.zeros(y_center.size()[:-1])
                 self.scale_ = torch.ones(y_scale.size()[:-1])
             elif isinstance(y_center, (np.ndarray, pd.Series, pd.DataFrame)):
-                self.center_ = np.zeros(y_center.shape[:-1])
-                self.scale_ = np.ones(y_scale.shape[:-1])
+                # numpy default type is numpy.float64 while torch default type is torch.float32 (if not changed)
+                # therefore, we first generate torch tensors (with torch default type) and then
+                # convert them to numpy arrays
+                self.center_ = torch.zeros(y_center.shape[:-1]).numpy()
+                self.scale_ = torch.ones(y_scale.shape[:-1]).numpy()
             else:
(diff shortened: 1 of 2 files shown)


jdb78:
Is there a better way than casting a tensor? Would rather like to create of the right dtype automatically

RonanFR:
We could force the `dtype`of the numpy array to be `numpy.float32` in any situation, but I do not know if this is really what we want. For instance, if the default `dtype ` of `torch ` is manually set by the user to `torch.float64` for some reasons (using `torch.set_default_dtype()`), then it may be preferable to  keep a `numpy.float64` array (numpy default). Also, it is not possible to change the default dtype of numpy (like for torch).

An alternative solution could be this:
`np.array(... , dtype=str(torch.get_default_dtype()).split(".")[1])`
i.e., convert the default torch dtype to string and use it as numpy dtype. This will make sure that `numpy.float32` is used whenever torch default dtype is `torch.float32` (default) and `numpy.float64` (default) is used whenever torch default dtype is `torch.float64`.

Let me know what you think.

jdb78:
Would prefer the alternative solution but given the small overhead, will merge anyways

PR #1129, opened 2022-09-08
Fix dtype check

JSpenced:
### Description

This PR fixes the dtype check in `is_numeric`. Currently, it will always evaluate to `pandas.core.series.Series` and never a categorical type as we are checking the `type` of the series that is passed in. This modifies to check if the dtype is categorical, which I think is the desired usage.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


jdb78:
Great find!

jdb78 (approved):

Issue #584, opened 2021-06-23
AssertionError: filters should not remove entries all entries - check encoder/decoder lengths and lags


ssooffiiaannee:
- PyTorch-Forecasting version: 0.9.0
- PyTorch version: 1.9.0
- Python version: 3.7.3
- Operating System: windows10

I'll not be able to share a notebook since I'm not allowed to share the data I'm working with.

```
max_prediction_length = 6
max_encoder_length = 24
training_cutoff = data["time_idx"].max() - max_prediction_length

training = TimeSeriesDataSet(
    data[lambda x: x.time_idx <= training_cutoff],
    time_idx = "time_idx",
    group_ids = ["year", "month", "dayOfWeek", "hour"],
    target = "count",
    min_encoder_length=1,  # keep encoder length long (as it is in the validation set)
    max_encoder_length=max_encoder_length,
    min_prediction_length = 1,
    max_prediction_length = max_prediction_length,
    time_varying_unknown_reals=["count"],
    add_relative_time_idx=True,
    add_target_scales=True,
    add_encoder_length=True,
    allow_missing_timesteps = True
)
I get the following error:
```
![image](https://user-images.githubusercontent.com/37119086/123179360-86ac9f00-d489-11eb-8e98-1e43d5bf1d49.png)

Here is what my data looks like:
![image](https://user-images.githubusercontent.com/37119086/123179067-ddfe3f80-d488-11eb-9f6f-5a6e775358dd.png)

However, when I set min_encoder_length=0, the previous code runs without error, but the exact same error appears when I run:
`
# create validation set (predict=True) which means to predict the last max_prediction_length points in time
# for each series
validation = TimeSeriesDataSet.from_dataset(training, data, predict=True, stop_randomization=True)

# create dataloaders for model
batch_size = 128  # set this between 32 to 128
train_dataloader = training.to_dataloader(train=True, batch_size=batch_size, num_workers=0)
val_dataloader = validation.to_dataloader(train=False, batch_size=batch_size * 10, num_workers=0)
`

ssooffiiaannee:
I managed to make it work by removing some columns in group_ids, Like in the following, But I still don't know why It didn't work before.
![image](https://user-images.githubusercontent.com/37119086/123225245-cf3c7a80-d4d2-11eb-950b-13e45e355d48.png)

[...]

PR #1162, opened 2022-10-26
Add some tests against simple mistakes in min_encoder_length and time_idx series

AndreyKrotkikh:
### Description

This PR adds obvious checks for input parameters. Keeping min_encoder_length with availability to set 0 will be reason of error in making attention plot which can be hardly connected without debugging. 

> ValueError: x and y must have same first dimension, but have shapes torch.Size([0]) and torch.Size([1])

### Checklist

- AssertionError: filters should not remove entries all entries - check encoder/decoder lengths and lags #584

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -333,8 +333,9 @@ class TimeSeriesDataSet(Dataset):
         if min_encoder_length is None:
             min_encoder_length = max_encoder_length
         self.min_encoder_length = min_encoder_length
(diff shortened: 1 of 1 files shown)


jdb78:
Strictly speaking neither do you need encoder length > 0 (equals zero just means a cold start), nor are different values for time required (imagine predicting one step ahead).

AndreyKrotkikh:
But how then to avoid this error message?

ValueError: x and y must have same first dimension, but have shapes torch.Size([0]) and torch.Size([1])

PR #1171, opened 2022-11-03
Added axis=0 to np.mean()

brendk:
### Description

This PR is to avoid the FutureWarning:
"FutureWarning: In a future version, DataFrame.mean(axis=None) will return a scalar mean over the entire DataFrame. To retain the old behavior, use 'frame.mean(axis=0)' or just 'frame.mean()'"
Added axis=0 argument to np.mean() calls where no axis was provided.


jdb78:
Think this works as intended. You are supposed to get a single scalar

PR #1176, opened 2022-11-08
Issue when optim lr find

sboal94:
### Description

Working with this model, when I launch the code to find the optimal learning rate. It gives an error Keyerror: 'radam_buffer', it is because it tries to find a value that is not in the state list, but in state['state'].

Piece of code show error.
```
# find optimal learning rate
res = trainer.tuner.lr_find(
    tft,
    train_dataloaders=train_dataloader,
    max_lr=10.0,
    min_lr=1e-6,
)

print(f"suggested learning rate: {res.suggestion()}")
fig = res.plot(show=True, suggest=True)
fig.show()
```

### Checklist

- [x] [Linked issues (if existing)](https://github.com/jdb78/pytorch-forecasting/issues/928#issuecomment-1256432343)
- [x] Amended changelog for large changes (and added myself there as contributor)
- [x] Added/modified tests


Christina-Young:
I'm getting this same issue in pytorch_forecasting v0.10.1.

woruoweikunlun:
I have the same error.

Titaniumtown:
Why are fixes like this not getting merged? Pretty large issues getting fixed.

PR #1194, opened 2022-12-02
fix. typo

lawrence-axb:
`decode_lenghts` -> `decode_lengths`


jdb78:
Fixing in another PR but good catch!

PR #1240, opened 2023-01-29
optim.py defaultdict, __setstate__

kvyaswanth:
added radam_buffer in default dict -- line 89

modified __setstate__ method to properly retrieve state dict -- lines (134-136)

code with changes reproduced here on stallion TFT model

### Description

This PR is to make changes to the "optim.py" file, the ranger optimizer, which is causing the issue [1233](https://github.com/jdb78/pytorch-forecasting/issues/1233)

### Checklist

- [x] Linked issues [1233](https://github.com/jdb78/pytorch-forecasting/issues/1233)
- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/pytorch_forecasting/optim.py
+++ b/pytorch_forecasting/optim.py
@@ -86,6 +86,7 @@ class Ranger(Optimizer):
             N_sma_threshhold=N_sma_threshhold,
             eps=eps,
             weight_decay=weight_decay,
+            radam_buffer = [[None, None, None] for ind in range(10)] # radam_buffer needs to be set to in defaults for it to updated in state dict
         )
         super().__init__(params, defaults)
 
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/optim.py
```diff
@@ -130,6 +131,9 @@ def __getstate__(self) -> dict:
 
     def __setstate__(self, state: dict) -> None:
         super().__setstate__(state)
+        # print(state['state'])
```
kvyaswanth, line 134: >> print(state['state'])
[output]: defaultdict(<class 'dict'>, {})

jdb78:
Good catch but now using the `pytorch_optimizer` package to avoid maintaining Ranger

Issue #1207, opened 2022-12-20
Pytorch-Forecasting imports deprecated property from transient dependency on numpy


JeroenPeterBos:
- PyTorch-Forecasting version: 0.10.3
- PyTorch version: 1.12.1
- Python version: 3.8
- Operating System: Ubuntu 20.04

### Expected behavior

I created a simple `TimeSeriesDataset` without specifying an explicit `target_normalizer`. I expected it to simply create a default normalizer deduced from the other arguments as explained in the documentation.

### Actual behavior

An exception was raised of the type `AttributeError` by the `numpy` package. The cause is that aliases like `numpy.float` and `numpy.int` have been deprecated as of numpy `1.20` which has been out for almost two years. The deprecation is explained [here](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations). This dependency on `numpy<=1.19` is not specified by `pytorch-forecasting` as described in #1130 .

### Code to reproduce the problem

Create an environment with the above versions and install numpy >= 1.20. Then run the first few cell's of the tutorial in the documentation for TFT: [https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html)

borab96:
numpy<1.24 will only raise warnings. Probably preferable to <=1.19

PR #1257, opened 2023-02-12
Fix np.float and np.float32 to float.

### LINKED ISSUE #1236: AttributeError: module 'numpy' has no attribute 'float'. Did you mean: 'cfloat'?
- PyTorch-Forecasting version: 0.10.3
- PyTorch version: 1.13.1
- Python version: 3.10.8
- Operating System: mac os
- Numpy version: 1.24.1

### Expected behavior

I executed the [stallion example](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html) and expected to run it successfully.

### Actual behavior

However, it failed and raised an error `AttributeError: module 'numpy' has no attribute 'float'. Did you mean: 'cfloat'?`. This is because 'float' was deprecated in the latest version of numpy.

BrunoBelucci:
### Description

This PR fixes #1236 and #1207.

Following the numpy error message, `np.float` was a deprecated alias for the builtin `float`.  The aliases was originally deprecated in NumPy 1.20.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -470,7 +470,7 @@ class TorchNormalizer(InitialParameterRepresenterMixIn, BaseEstimator, Transform
         if isinstance(y_center, torch.Tensor):
             eps = torch.finfo(y_center.dtype).eps
         else:
-            eps = np.finfo(float).eps
+            eps = np.finfo(np.float32).eps
         if self.method == "identity":
             if isinstance(y_center, torch.Tensor):
                 self.center_ = torch.zeros(y_center.size()[:-1])
@@ -785,7 +785,7 @@ class GroupNormalizer(TorchNormalizer):
             self
         """
         y = self.preprocess(y)
-        eps = np.finfo(float).eps
+        eps = np.finfo(np.float32).eps
         if len(self.groups) == 0:
             assert not self.scale_by_group, "No groups are defined, i.e. `scale_by_group=[]`"
             if self.method == "standard":
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -470,7 +470,7 @@ class TorchNormalizer(InitialParameterRepresenterMixIn, BaseEstimator, Transform
         if isinstance(y_center, torch.Tensor):
             eps = torch.finfo(y_center.dtype).eps
         else:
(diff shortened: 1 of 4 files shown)


graille:
Not sure np.float32 is deprecated. I am pretty sure float is np.float64. So you should keep np.float32 expressions since GPU usually works with 32 bits floatting point numbers.

pytorch_forecasting/data/encoders.py
```diff
@@ -470,7 +470,7 @@ def _set_parameters(
         if isinstance(y_center, torch.Tensor):
             eps = torch.finfo(y_center.dtype).eps
         else:
-            eps = np.finfo(np.float).eps
```
BrunoBelucci, line 473: 32 or 64 bits? Why not `y_center.dtype?`

pytorch_forecasting/data/encoders.py
```diff
@@ -785,7 +785,7 @@ def fit(self, y: pd.Series, X: pd.DataFrame):
             self
         """
         y = self.preprocess(y)
-        eps = np.finfo(np.float).eps
```
BrunoBelucci, line 788: 32 or 64 bits? We cannot use `y.dtype` because it is possible to have different type for y and X.

BrunoBelucci:
> np.float is indeed depreacated, but I'm not sure np.float32 is also deprecated. I am pretty sure float is np.float64. So you should keep np.float32 expressions since GPU usually works with 32 bits floatting point numbers.

You are absolutely right, I have assumed the inverse, that float was np.float32. Therefore, I should not have changed np.float32 to float, but it is not clear to me when we are doing `eps = np.finfo(float).eps` if we want 32 or 64 bits.

BrunoBelucci:
After some thought, I decided to change np.float to np.float32 following the assumption that we usually work with GPUs, so eps from np.float64 can be 0 if we use np.float32, but eps from np.float32 will never be 0.

ben1787 (approved):

EthanReid:
Is this update still in progress?

fnhirwa (commented):
Closing this PR as this was added in this commit
https://github.com/sktime/pytorch-forecasting/pull/1280/files#diff-c27f0d8f4e94846118c3f33ba63b95892f39f9a552dfa62b33fe36089fdab504

PR #1289, opened 2023-04-13
'encoder_features' referenced before assignment

orangehe:
the static covariate should be extracted first before referenced in line 256.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


jdb78:
Are you sure about this? Because the features are static, they don't need extra extraction. You can just take the first timestep of the encoder features.

Issue #1314, opened 2023-05-26
Multiple Target Prediction Plotting Bug


terbed:
- PyTorch-Forecasting version: 1.0.0
- PyTorch version: 2.0
- Python version: 3.9
- Operating System: Ubuntu

# Description:
When calling the plot_prediction() function in PytorchForecasting with multiple targets, the function reuses the same axes for each target. This behavior results in overlapped plots for different targets, rather than separate plots for each target.

# To reproduce:
- Train a PytorchForecasting model with multiple targets.
- Call the plot_prediction() function with an output dictionary that includes multiple targets.

# The faulty code part
The bug is in the base model plot prediction function:
[def plot_prediction()](https://github.com/jdb78/pytorch-forecasting/blob/7c775c1a2fe3effe62eac0a476db98ca5b24ca4c/pytorch_forecasting/models/base_model.py#L957)

```
def plot_prediction(
    self,
    x: Dict[str, torch.Tensor],
    out: Dict[str, torch.Tensor],
    idx: int = 0,
    add_loss_to_title: Union[Metric, torch.Tensor, bool] = False,
    show_future_observed: bool = True,
    ax=None,
    quantiles_kwargs: Dict[str, Any] = {},
    prediction_kwargs: Dict[str, Any] = {},
) -> plt.Figure:

    #...

    # for each target, plot
    figs = []
    for y_raw, y_hat, y_quantile, encoder_target, decoder_target in zip(
        y_raws, y_hats, y_quantiles, encoder_targets, decoder_targets
    ):
        # ...

        # create figure
        if ax is None:
            fig, ax = plt.subplots()
        else:
            fig = ax.get_figure()
        
        # ...

        figs.append(fig)
    
    return figs
```

## Expected behavior:
Each target should be plotted on a separate figure.

## Actual behavior:
All targets are plotted on the same figure, resulting in overlapped plots.

# Solution:
In the above snippet, the variable ax should be updated within the loop over targets but instead after the first target, the same ax is reused (as ax is no longer None). A possible solution and fix for the problem:

```
    def plot_prediction(
        self,
        x: Dict[str, torch.Tensor],
        out: Dict[str, torch.Tensor],
        idx: int = 0,
        add_loss_to_title: Union[Metric, torch.Tensor, bool] = False,
        show_future_observed: bool = True,
        ax=None,
        quantiles_kwargs: Dict[str, Any] = {},
        prediction_kwargs: Dict[str, Any] = {},
[...]

Issue #1247, opened 2023-02-06
Wrong documentation Groupnormalizer


terbed:
https://pytorch-forecasting.readthedocs.io/en/stable/_modules/pytorch_forecasting/data/encoders.html#GroupNormalizer

>         """
>         Group normalizer to normalize a given entry by groups. Can be used as target normalizer.
> 
>         Args:
>             method (str, optional): method to rescale series. Either "standard" (standard scaling) or "robust"
>                 (scale using quantiles 0.25-0.75). Defaults to "standard".
>             method_kwargs (Dict[str, Any], optional): Dictionary of method specific arguments as listed below
>                 * "robust" method: "upper", "lower", "center" quantiles defaulting to 0.75, 0.25 and 0.5
>             groups (List[str], optional): Group names to normalize by. Defaults to [].
>             center (bool, optional): If to center the output to zero. Defaults to True.
>             scale_by_group (bool, optional): If to scale the output by group, i.e. norm is calculated as
>                 ``(group1_norm * group2_norm * ...) ^ (1 / n_groups)``. Defaults to False.
>             transformation (Union[str, Tuple[Callable, Callable]] optional): Transform values before
>                 applying normalizer. Available options are
> 
>                 * None (default): No transformation of values
>                 * log: Estimate in log-space leading to a multiplicative model
**>                 * logp1: Estimate in log-space but add 1 to values before transforming for stability
>                     (e.g. if many small values <<1 are present).**
>                     Note, that inverse transform is still only `torch.exp()` and not `torch.expm1()`.
>                 * logit: Apply logit transformation on values that are between 0 and 1
>                 * count: Apply softplus to output (inverse transformation) and x + 1 to input
>                     (transformation)
>                 * softplus: Apply softplus to output (inverse transformation) and inverse softplus to input
>                     (transformation)
>                 * relu: Apply max(0, x) to output
>                 * Dict[str, Callable] of PyTorch functions that transforms and inversely transforms values.
>                   ``forward`` and ``reverse`` entries are required. ``inverse`` transformation is optional and
>                   should be defined if ``reverse`` is not the inverse of the forward transformation. ``inverse_torch``
>                   can be defined to provide a torch distribution transform for inverse transformations.
> 
>         """

not logp1, it is log1p!

PR #1317, opened 2023-05-27
Multiple Target Prediction Plotting Bug

terbed:
### Description
When calling the plot_prediction() function in PytorchForecasting with multiple targets, the function reuses the same axes for each target. This behavior results in overlapped plots for different targets, rather than separate plots for each target.
This pull request fixes this issue.

### Issue
#1314 

### The faulty code part
```
def plot_prediction(
    self,
    x: Dict[str, torch.Tensor],
    out: Dict[str, torch.Tensor],
    idx: int = 0,
    add_loss_to_title: Union[Metric, torch.Tensor, bool] = False,
    show_future_observed: bool = True,
    ax=None,
    quantiles_kwargs: Dict[str, Any] = {},
    prediction_kwargs: Dict[str, Any] = {},
) -> plt.Figure:

    #...

    # for each target, plot
    figs = []
    for y_raw, y_hat, y_quantile, encoder_target, decoder_target in zip(
        y_raws, y_hats, y_quantiles, encoder_targets, decoder_targets
    ):
        # ...

        # create figure
        if ax is None:
            fig, ax = plt.subplots()
        else:
            fig = ax.get_figure()
        
        # ...

        figs.append(fig)
    
    return figs
```

_Expected behavior:_
Each target should be plotted on a separate figure.

_Actual behavior:_
All targets are plotted on the same figure, resulting in overlapped plots.

### Solution
In the above snippet, the variable ax should be updated within the loop over targets but instead after the first target, the same ax is reused (as ax is no longer None). The proposed issue fix is:

```
    def plot_prediction(
        self,
        x: Dict[str, torch.Tensor],
        out: Dict[str, torch.Tensor],
        idx: int = 0,
        add_loss_to_title: Union[Metric, torch.Tensor, bool] = False,
        show_future_observed: bool = True,
        ax=None,
        quantiles_kwargs: Dict[str, Any] = {},
        prediction_kwargs: Dict[str, Any] = {},
    ) -> plt.Figure:

        # ...
        # for each target, plot
        figs = []
        ax_provided = ax is not None
        for y_raw, y_hat, y_quantile, encoder_target, decoder_target in zip(
            y_raws, y_hats, y_quantiles, encoder_targets, decoder_targets
        ):

            # ...
            # create figure
            if (ax is None) or (not ax_provided):
                fig, ax = plt.subplots()
            else:
                fig = ax.get_figure()
```

### Bonus
Corrected mistakes in documentation. The encoder's `log1p` transformation is incorrectly called `logp1` in the documentation.
#1247

### DIFF
--- a/pytorch_forecasting/models/base_model.py
+++ b/pytorch_forecasting/models/base_model.py
@@ -993,6 +993,7 @@ class BaseModel(InitialParameterRepresenterMixIn, LightningModule, TupleOutputMi
 
         # for each target, plot
         figs = []
+        ax_provided = ax is not None
         for y_raw, y_hat, y_quantile, encoder_target, decoder_target in zip(
             y_raws, y_hats, y_quantiles, encoder_targets, decoder_targets
         ):
@@ -1012,7 +1013,7 @@ class BaseModel(InitialParameterRepresenterMixIn, LightningModule, TupleOutputMi
             # move to cpu
             y = y.detach().cpu()
             # create figure
-            if ax is None:
+            if (ax is None) or (not ax_provided):
                 fig, ax = plt.subplots()
             else:
                 fig = ax.get_figure()
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -414,7 +414,7 @@ class TorchNormalizer(InitialParameterRepresenterMixIn, BaseEstimator, Transform
 
                 * None (default): No transformation of values
                 * log: Estimate in log-space leading to a multiplicative model
-                * logp1: Estimate in log-space but add 1 to values before transforming for stability
+                * log1p: Estimate in log-space but add 1 to values before transforming for stability
                   (e.g. if many small values <<1 are present).
                   Note, that inverse transform is still only `torch.exp()` and not `torch.expm1()`.
                 * logit: Apply logit transformation on values that are between 0 and 1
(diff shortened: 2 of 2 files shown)


terbed:
Is this repo not maintained?

pytorch_forecasting/models/base_model.py
```diff
@@ -1213,7 +1217,7 @@ def configure_optimizers(self):
                     min_lr=self.hparams.reduce_on_plateau_min_lr,
                 ),
                 "monitor": "val_loss",  # Default: val_loss
-                "interval": "epoch",
+                "interval": "step",
```
jdb78, line 1220: don't think this makes sense. reduce each step is very aggressive

pytorch_forecasting/models/base_model.py
```diff
@@ -1048,13 +1049,16 @@ def plot_prediction(
...
+                    try:
+                        ax.errorbar(
+                            x_pred,
+                            y[[-n_pred]],
+                            yerr=quantiles - y[-n_pred],
+                            c=pred_color,
+                            capsize=1.0,
+                        )
+                    except ValueError:
+                        print(f"Warning: could not plot error bars. Quantiles: {quantiles}, y: {y}, yerr: {quantiles - y[-n_pred]}")
```
jdb78, line 1061: this would have to be a `logger.warning()` instead of `print()`

jdb78: why does the error actually happen. Seems like this solves a different problem

pytorch_forecasting/models/base_model.py
```diff
@@ -1012,7 +1013,7 @@ def plot_prediction(
             # move to cpu
             y = y.detach().cpu()
             # create figure
-            if ax is None:
+            if (ax is None) or (not ax_provided):
```
jdb78, line 1016: this is confusing. I would say: `ax_not_provided = ax is None` before the loop and then `if ax_not_provided` after the loop

PR #1340, opened 2023-06-30
Add None as valid type for target_normalizer

tmct:
`None` is already allowed in the docs and code - let's put it in the Python typing please


jdb78 (approved):

tmct:
Thank you

PR #1357, opened 2023-08-04
Bugfix: handle case where 'losses' is a list of tensors

TeodorChiaburu:
### Description

This PR fixes an issue I came across when applying NHiTS and TFT on my use case, namely that `losses` can be tensors within a list. In that case, the function `mask_losses` will fail. It is just a one-liner in **base_metrics.py**.

### DIFF
--- a/pytorch_forecasting/metrics/base_metrics.py
+++ b/pytorch_forecasting/metrics/base_metrics.py
@@ -822,6 +822,9 @@ class MultiHorizonMetric(Metric):
         """
         if reduction is None:
             reduction = self.reduction
+
+        if isinstance(losses, list): losses = losses[0]
+
         if losses.ndim > 0:
             # mask loss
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -822,6 +822,9 @@ def mask_losses(self, losses: torch.Tensor, lengths: torch.Tensor, reduction: st
         """
         if reduction is None:
             reduction = self.reduction
+
+        if isinstance(losses, list): losses = losses[0]
```
jdb78, line 826: Should that not be something like
`losses = [self.mask_losses(loss, lengths, reduction=reduction) for loss in losses]`?

jdb78 (commented):
Think this might need some rework. selecting the first loss might make the function work but if a list goes in, a list should go out

jdb78:
BTW: Would be great to have a test for this!

Issue #1071, opened 2022-07-25
Bug Report: Error encountered when the covariates' size is different  in N-Hits


zhangguangxun:
- PyTorch-Forecasting version:0.10.1
- PyTorch version:1.12.0
- Python version:3.8
- Operating System:Linux version 4.19.91-007.ali4000.alios7.x86_64

### Abstract

As #1065 issus metioned, N-HiTS encounters a RuntimeError when using 'time_varying_unknown_reals' and 'time_varying_known_reals' covariates simultaneously. The reason of that error is becasue the dimension of the full-connected network is mismatched during the network define process and tensor forward process.

### Error Location In Source Code
In `pytorch_forecasting.models.nhits.sub_modules.py` From line 124 to line 128, The Linear layers' dimension is defined by
```
self.hidden_size = [
            self.context_length_pooled * self.output_size
            + (self.context_length + self.prediction_length) * self.covariate_size
            + self.static_hidden_size
        ] + hidden_size
```
The `covariate_size ` is input and is defined in `pytorch_forecasting.models.nhits.__init__.py` From line 200 to line 208
```
def covariate_size(self) -> int:
        """Covariate size.

        Returns:
            int: size of time-dependent covariates
        """
        return len(set(self.hparams.time_varying_reals_decoder) - set(self.target_names)) + sum(
            self.embeddings.output_size[name] for name in self.hparams.time_varying_categoricals_encoder
        )
```
From the defination of variable `self.covariate_size` we can figure out the defination is only consider the size of categoricals variables in encoder and the size of reals variables in decoder.

Howerer during the forward process, the dimension of input tensor is defined in  `pytorch_forecasting.models.nhits.sub_modules.py` From line 163 to line 181
```
        encoder_y = encoder_y.transpose(1, 2)
        # Pooling layer to downsample input
        encoder_y = self.pooling_layer(encoder_y)
        encoder_y = encoder_y.transpose(1, 2).reshape(batch_size, -1)

        if self.covariate_size > 0:
            encoder_y = torch.cat(
                (
                    encoder_y,
                    encoder_x_t.reshape(batch_size, -1),
                    decoder_x_t.reshape(batch_size, -1),
                ),
                1,
            )

        # Static exogenous
        if (self.static_size > 0) and (self.static_hidden_size > 0):
            x_s = self.static_encoder(x_s)
            encoder_y = torch.cat((encoder_y, x_s), 1)
```
[...]

PR #1359, opened 2023-08-04
Bug Fix for #1071: NHiTs encoder/decoder covariate size difference bug

bendavidsteel:
### Description

Fix for https://github.com/jdb78/pytorch-forecasting/issues/1071 and https://github.com/jdb78/pytorch-forecasting/issues/1065


jdb78 (commented):
Can we add a test. Looks like it makes a lot of sense but would like to have a test for this as it is an actual change in the model

bendavidsteel:
Sure! Added a new test for NHiTs, confirmed it fails on master and passes on my branch

jdb78 (dismissed):
Great PR! thanks a lot

bendavidsteel:
Thanks! Just fixed the linter problem, looks like you'll need to approve again to get the workflows to run

manitadayon:
@ jdb78, it seems, the merge did not take effect or there has been an update to modify this behavior since it still throws an error when covariates for encoder and decoder are different in NHiTS. 
Initially, using the @bendavidsteel  code was working, now even his modified version is throwing the same error (I forked his branch and tested his code is working)
Wondering if anything/update is changed to the core model? or if someone can test the model behavior?

PR #1432, opened 2023-11-06
Nhits weight argument fix in TimeSeriesDataSet

manitadayon:
Adding a **weight** argument to TimeSeriesDataSet will cause the trainer to throw an error for NhiTS because of a dimension mismatch in dimension 1. 
This is because of unsqueeze operation and it is not solely because of missing values (the error will happen even when there is no missing values).
To fix this we can remove the unsqueeze(-1) and multiply the losses and weight directly.
(Ignore the first few commits as they are for NHitS encode-decoder length which is already merged to main)
This is the commit: ([**Remove Unsqueeze operation to solve the mismatch operation](https://github.com/jdb78/pytorch-forecasting/commit/f972a8c607dccbb39ab558fa8c1d0d376ca03cba**)

I confirm this by running the same NhiTS model with weight operation set:

Issues:
https://github.com/jdb78/pytorch-forecasting/issues/1431
https://github.com/jdb78/pytorch-forecasting/issues/1040

### DIFF
--- a/tests/test_models/conftest.py
+++ b/tests/test_models/conftest.py
@@ -130,6 +130,36 @@ def make_dataloaders(data_with_covariates, **kwargs):
 def multiple_dataloaders_with_covariates(data_with_covariates, request):
     return make_dataloaders(data_with_covariates, **request.param)
 
+@pytest.fixture(scope="session")
+def dataloaders_with_different_encoder_decoder_length(data_with_covariates):
+    return make_dataloaders(
+        data_with_covariates.copy(),
+        target="target",
+        time_varying_known_categoricals=["special_days", "month"],
+        variable_groups=dict(
+            special_days=[
+                "easter_day",
+                "good_friday",
+                "new_year",
+                "christmas",
+                "labor_day",
+                "independence_day",
+                "revolution_day_memorial",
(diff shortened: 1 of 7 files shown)


jobs-git:
Any update?

fkiraly (changes requested):
We should add a test that ensures the bug is fixed.

The changes to `conftest` also seem to override an already existing fixture - is that intentional?

FYI @jobs-git, this PR is 2 years old, feel free to finalize it (branch off the PR)

tests/test_models/conftest.py
```diff
@@ -199,6 +199,50 @@ def dataloaders_with_different_encoder_decoder_length(data_with_covariates):
     )
 
 
+@pytest.fixture(scope="session")
```
fkiraly: this is just a copy-paste duplication so I will remove it.

fkiraly:
I am going through the issues and realize no one has posted a full example that we could use in testing.

Could one of you post a full MRE https://stackoverflow.com/help/minimal-reproducible-example

@manitadayon, @luigidarksimeone, @fnavruzov, @ronanfr, @francescofondaco, @qijiashao, @terbed

PR #1449, opened 2023-11-22
init

svnv-svsv-jm:
### Description

This PR fixes my own issues:
* https://github.com/jdb78/pytorch-forecasting/issues/1440
* https://github.com/jdb78/pytorch-forecasting/issues/1448

It allows to tune "any" model, generalizing the `optimize_hypetparameters` function from `TemporalFusionTransformer` model, and it also add a "better" `LSTMModel` than the one shown in the documentation, which I tried to use on a multi-target dataset without success. My version does work on multi-target datasets, thanks to an extension to `AutoRegressiveBaseModel`.

My version of `AutoRegressiveBaseModel` inherits from the original to override a couple of methods. I should've probably just edited the original, but I'd like to hear your feedback before doing that. I'm also sure there was a simpler way for `AutoRegressiveBaseModel` to work with multi-target data. When I tried to just run the Documentation's example of `LSTMModel` but on a multi-target dummy dataset, it just did not work out of the box.

### Checklist

- Linked issues: [https://github.com/jdb78/pytorch-forecasting/issues/1440](1440) , [https://github.com/jdb78/pytorch-forecasting/issues/1448](1448)
- Amended changelog for large changes (and added myself there as contributor)
- Added/modified tests: [tests/test_models/test_tuning.py](tests/test_models/test_tuning.py) 
- [Yes] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.

Make sure to have fun coding!


svnv-svsv-jm:
What's blocking this?

svnv-svsv-jm:
I don't understand why the docs test fails (I haven't changed anything there) and why the Black test fails if I ran `black` and on my machine `black --check .` returns:

```bash
>> black --check .
All done! ✨ 🍰 ✨
54 files would be left unchanged.
```

PR #1499, opened 2024-01-24
Bugfix: if infite loss is detected, error will be thrown during the backpropogation

germanKoch:
infinite loss bug fixed

### Description

The bug is described here: https://lightning.ai/forums/t/why-does-training-fails-with-require-grad-and-does-not-have-a-grad-fn/2782

If loss is not finite, the following error will be thrown:
```
require grad and does not have a grad_fn
```

This bug was reproduced by me and proposed fix is checked bu me as well.


jdb78:
That's intentional and tells you that something is wrong - probably learning rate is too high and therefore running into instability

Issue #1325, opened 2023-06-05
Training TFT with mixed precision


Marcrb2:
- PyTorch-Forecasting version: 1.0.0
- PyTorch version: 1.0.1
- Python version: 3.8
- Operating System: Ubuntu

### Expected behavior

I was expecting some speedup training on the TemporalFusionTransformer with mixed precision.

First of all, to be able to train with mixed precision, the source code needs to be modified (as stated in #285):

line 391 in the temporal_fusion_...\sub_modules.py file:
From: `attn = attn.masked_fill(mask, -1e9)`
To: `attn = attn.masked_fill(mask, -float("inf"))`

### Actual behavior

Doing so, I managed to train using FP16 Mixed Precision, and I see the following message confirming so in the logs:
"Using 16bit Automatic Mixed Precision (AMP)".

However, I've noticed that the training speed has unexpectedly slowed down significantly:

With FP16 Mixed Precision (`pl.Trainer precision='16-mixed'`), it takes approximately **1.46 seconds per iteration** (1.46s/it).
With FP32 (default), it achieves approximately **2.15 iterations per second** (2.15it/s).

I have tested and seen this behavior using both a T4 GPU and an A100

Is there anything else I need to configure to use mixed precision effectively?

Thanks in advance.

Issue #285, opened 2021-01-23
Support for 16bit precision training


BlockWaving:
Thanks for the fantastic implementation... this is really a awesome job.
When testing with "precision=16" set on Trainer, however, an error triggered because of following code:

temporal_fusion_...\sub_modules.py
Line 393:  attn = attn.maskked_fill(mask, -1e9)
RuntimeError:  value cannot be converted to type at::Half without overflow -1e9

Please advise to make it support for 16bit training, what should I do here? changing all -1e9 to -1e4 will make it?

jdb78:
Sounds like a great idea. Do you want to replace the number with -3e4 and write a small test to confirm it works? Would be an amazing PR!

BlockWaving:
Test it, changing to -3e4 disappeared the line 393 error, but later in optim.py:
Line 131:  _= closure()
TypeError: 'NoneType' object is not callable

Looks more places need to be touched.

Tracing to training_loop.py(L719-730), closure is created here:

                       def train_step_and_backward_closure():
                            result = self.training_step_and_backward(
                                split_batch,
                                batch_idx,
                                opt_idx,
                                optimizer,
                                self.trainer.hiddens
                            )
                            return None if result is None else result.loss

                        # optimizer step
                        self.optimizer_step(optimizer, opt_idx, batch_idx, train_step_and_backward_closure)
Somehow 'result' was returned as None.
Or is this also a lightning fix?

jdb78:
Hm. Can you try a different optimizer? Maybe ranger is the issue. Could you give `"adam"` a try?

mikcnt:
Hey @jdb78! I'm currently facing the same issue when trying to train the model with mixed precision (`precision = 16`). The first issue is easily solved by replacing -1e9 with either a smaller number like -3e4 like you advised, or with `- math.inf`, which seems like a better option in my opinion.

As for the second one, I've tried using `"adam"` as optimizer, and it seems to work, so ranger is the issue. Any idea on how to proceed? I can totally work on that!

sayanb-playground:
Hi @jdb78, this issue still persists while trying to train TFT with lightning's native mixed precision mode on the following versions.

```
pytorch-forecasting==0.9.0
pytorch-lightning==1.6.5
torch==1.11.0
```

Any chance of getting it fixed in the upcoming version? Let me know if I can be of any help here. I'd be curious to know if mixed precision training will bring down our TFT runtime.

wmoreira-abi:
Up!

PR #1518, opened 2024-02-19
Allow mixed precision TFT training

Marcrb2:
### Description

This PR modifies the attention mask in the TFT model from 1e-9 to float("inf") to allow Pytorch mixed precision training.

Related issues: 
#1325
#285

### DIFF
--- a/pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py
+++ b/pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py
@@ -388,7 +388,7 @@ class ScaledDotProductAttention(nn.Module):
             attn = attn / dimension
 
         if mask is not None:
-            attn = attn.masked_fill(mask, -1e9)
+            attn = attn.masked_fill(mask, -float("inf"))
         attn = self.softmax(attn)
 
         if self.dropout is not None:


MrGG14 (approved):

fkiraly (changes requested):
Thanks!

From the API perspective, we need to address the problem that this changes default behaviour, and thus may break or change user code downstream.

What we need to do is to expose the constant in `masked_fill` as a parameter in `ScaledDotProductAttention`, up until `TemporalFusionTransformer`, that means also in `InterpretableMultiHeadAttention`.

The default must be left at 1e9 for now (to avoid breaking/changing other peoples' code), but this will allow you to set `float("inf")` by using the new parameter.

The docstrings of the modules and components should also get the description of the new parameter. Where classes have no docstring, it would be great if you could add one, but that is not blocking.

fkiraly:
I would also appreciate feedback on which value is better as a default - even if we cannot change it right now, we could after a couple releases, giving the users a forewarning that the default will change.

tr01-crypto:
is there an update to this issue? i would really need this feature..

hima-erla:
Same, I really need this feature too. Mixed precision is an important aspect to be able to tune train heavy models like TFT. Can someone action this?

fkiraly (approved):
I think @Marcrb2 has disengaged, I've wrapped this up now.

Instead of changing the default, the `mask_bias` can now be manually set to -inf, which should be sufficiente for your use cases, @tr01-crypto, @hima-erla?

PR #1599, opened 2024-08-23
[MNT] Fix dependency issues and CI runners: `numpy2`, `optuna`, `requests`, MacOS MPS

XinyuWuu:
Fixes #1594, fixes #1595, fixes #1596

Added or moved some dependencies to core dependency set.

Fixed some `numpy2` and `optuna-integrations` problems.

`requests` replaced by `urllib.request.urlretrieve`.

PR #1624, opened 2024-08-26
[MNT] Relax `numpy` bound to `numpy<3.0.0`

XinyuWuu:
Relaxes `numpy` bound to `numpy<3.0.0`.

(updated by @fkiraly - removed the `optuna` bound which has been dealt with in https://github.com/jdb78/pytorch-forecasting/pull/1629)

---

It should work with newer `numpy` and `optuna` version after #1599.


XinyuWuu:
numpy2 support of torch: https://github.com/pytorch/pytorch/issues/107302#issuecomment-2048405223

macOS x86 builds / test deprecation: https://github.com/pytorch/pytorch/issues/114602

It seems impossible to have numpy2 and torch working in macOS x86.

fkiraly:
Why is this not a problem in the `sktime` runners, where we have `torch` models and also an OS matrix?

XinyuWuu:
> Why is this not a problem in the `sktime` runners, where we have `torch` models and also an OS matrix?

Actully we are having torch 2.2.2 and numpy 1.24 for macos in sktime.

fkiraly:
do you know what causes that? because for sure we have numpy 2 on most jobs (right?)

XinyuWuu:
> do you know what causes that? because for sure we have numpy 2 on most jobs (right?)

pytorch has deprecated macOS x86 after version 2.2.2, but torch 2.2.2 do not suppurt numpy2.

fkiraly:
I see, so we are forced torch <2.2 because we enforce numpy 2, am I understanding correctly?

XinyuWuu:
> I see, so we are forced torch <2.2 because we enforce numpy 2, am I understanding correctly?

We are forced numpy 1 because we are forced torch<=2.2.2. We can not install torch>2.2.2 in macOS x86 because pytorch has deprecated it after 2.2.2. We can not support numpy 2 in macOS x86 because pytorch supports numpy 2 only with torch>=2.3.0 which can not be installed in macOS x86.

fkiraly:
I think the issue is resolved by using `macos-latest`, which is 64 and not 86

PR #1630, opened 2024-08-29
[MNT] Update pre commit hooks and run all

yarnabrina:
1. updates all the `pre-commit` hooks to their latest versions using `pre-commit autoupdate`
2. runs all hooks on all files using `pre-commit run --all-files`


fkiraly (approved):
Seems like it did not change too many files?

PR #1637, opened 2024-08-30
[MNT] isolate `cpflows` to its soft dependency set

fkiraly:
This Pr isolates `cpflows` to its soft dependency set again - I think it was already a soft dependency before we started to rework `pyproject`, somehow it ended up in `dev`?

This PR restores its original status as a soft dependency, as part of `mqf2`.


benHeid:
Oh this is a mistake. This shouldn't be in dev. It is required for some distributions I think...

PR #1633, opened 2024-08-30
[MNT] switch CI macos runner to `macos-latest`

fkiraly:
This switches the macos runner used in the CI to `macos-latest`, from `macos-13`.

PR #1646, opened 2024-09-03
[MNT]Remove `.env` file (we shouldn't be pushing the `.env` file

fnhirwa:
### Description

This PR removes the `.env` file present in the repository as it is not a good practice to push changes to `.env` file.

Will first need to inspect the issue that this commit https://github.com/jdb78/pytorch-forecasting/commit/038b0c30c45866b83fd568a2069aeff42187f920 was resolving as it is when the file was added.

Once I am sure that this PR doesn't break anything we can go ahead with the review.


fkiraly:
The "MPS" line might be linked to this?
https://github.com/jdb78/pytorch-forecasting/issues/1596

It does not look like anything is failing, anyway.

fkiraly:
See here for the actual failure: https://github.com/jdb78/pytorch-forecasting/pull/1633

fkiraly:
might I suggest to create a branch off this PR, and merge it with #1633, just to see what happens?

fnhirwa:
This is device handling related, I am going to open an issue to fix this generally and start working on a Pull Request to fix this.

.github/workflows/test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
fkiraly: I would suggest, let's not reconfigure the mac VM in this PR, but only separately - now the CI also fails.

fnhirwa: Yeah we should isolate that I am experimenting this in https://github.com/jdb78/pytorch-forecasting/pull/1648
Still haven't found the working solution.

fkiraly: could you split and keep only the env removal in this PR? That looks fine by itself.

fkiraly:
ready for merge?

fnhirwa:
Yes it is ready to merge.

fkiraly (approved):

PR #1651, opened 2024-09-04
[MNT] remove `poetry.lock`

fkiraly:
Deletes `poetry.lock` to remove the burden of maintaining it - the developer workflow of installing `pytorch-forecasting[dev]` is more general and less complex.


jdb78:
Think the only thing to consider is how in the future we want to detect new releases that are breaking the current tests - having a lock file really helps because dependabot creates PRs automatically, updating the lock file and it is immediately clear which updates to dependencies break the tests.

fkiraly:
That's an interesting point! I was wondering where in the workflow the `poetry.lock` would be useful.

Just to clarify:

* does `poetry.lock` monitor *each direct and indirect dependency*?
* does it still lock dependencies after the CI is no longer using poetry?

> updating the lock file and it is immediately clear which updates to dependencies break the tests.

In my other maintenance workflows, I just run diffs of `pip list`-s on the first failing CI run, and the last passing CI run. Since testing happens regularly, at least once a week by CRON, only a very small number of versions typically change, and that combined with the error message normally allows to pinpoint the culprit.

For the `poetry` workflow, I'm just concerned about PR spam...

PR #1655, opened 2024-09-07
Fix/ar-tutorial

jdb78:
This PR fixes a device issue for the 'ar' tutorial where tensors were on different devices if a "mps" device is present leading to an error.


fkiraly (approved):
I see - any suggestions on how we test this?

I suppose we know that MPS devices are *not* present on `macos-latest` by virtue of this?
https://github.com/jdb78/pytorch-forecasting/issues/1596

So we ought to run the notebooks on `macos-latest` as well?

PR #1667, opened 2024-09-10
[BUG] Bugfix when no exogenous variable is passed to TFT

XinyuWuu:
This PR fix the bug when no exogenous variable is passed to TFT.

Previously, if no exogenous variable is passed, TFT will fail with `log_interval != -1` and do nothing when `log_interval == -1`.

Details can be found: https://github.com/sktime/sktime/pull/6228


fnhirwa (dismissed):
Looks good to me and the tests are passing only the common linting errors unrelated to this PR.

Merging!

PR #1688, opened 2024-09-25
[MNT] `show_versions` utility

fkiraly:
Adds a `show_versions` utility for users to easily share version information in bug reports or questions.

Adapted from `sktime`, which in turn is an evolution of the `sklearn` utility of the same name.


fkiraly:
Anyone understands what is going on with the docs failure?

fkiraly:
Fixed, it was not a docs failure, but faulty location and imports in the tests.

In ptf, the tests need to be in root/tests, and the tests put in the utils folder were not executed.

PR #1692, opened 2024-10-05
[MNT] Move linting to ruff

### LINKED ISSUE #1689: [MNT] move linting to `ruff`, resolve strange linting behaviour
We should move the linting to `ruff` - good first issue for a maintenance affine contributor.

There are also some strange behaviours in the precommits which I cannot pinpoint, e.g., the linter insisting on long single lines instead of line breaks where they would be appropriate. Though I would hope this is resolved as a side effect of moving to `ruff`.

airookie17:
### Description

This PR replaces the existing linters `flake8` and `isort` with `ruff`. 

Fixes #1689

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


pyproject.toml
```diff
@@ -1,3 +1,34 @@
+[tool.ruff]
+line-length = 120
```
fkiraly, line 2: could you change this to 88?

fkiraly (changes requested):
Looks good! I only have a comment about the max line length.

fkiraly (approved):
Actually, I am happy with this since it is an 1:1 translation.

We should change line length in another PR.

airookie17:
Thanks for the review! I will change the line length in a different PR as suggested.

PR #1697, opened 2024-10-15
[MNT]`black` doesn't have `extras` dependency

fnhirwa:
This PR adds the correct dependency for `black` jupyter instead of `extras` as it doesn't have that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ dev = [
   # jupyter notebook
   "ipykernel",
   "nbconvert",
-  "black[extras]",
+  "black[jupyter]",
   # documentatation
   "sphinx",
   "pydata-sphinx-theme",


fkiraly (approved):
Makes sense.

Btw, do we need `black` still, now that we are using `ruff`?

fnhirwa:
> Makes sense.
> 
> Btw, do we need `black` still, now that we are using `ruff`?

If we are using ruff we can remove black or depending on @yarnabrina suggestion on linters we might need to add extra linters given that ruff doesn't have some rules implemented.

yarnabrina:
I was quite a strong supporter of ruff, but given the limitation of multi file analysis, I'm now quite concerned in general. But it's for the linting component so far, I do not have any opinion for the formatting component.

Issue #1668, opened 2024-09-10
[MNT] remove mutable objects from defaults


fkiraly:
We should ensure that no mutable objects are argument defaults, e.g., lists, dicts.

All mutable defaults should be replaced by appropriate defaults, e.g., strings if applicable, or `None`, which internally is then replaced by a newly initialized mutable defaults.

Care needs to be taken in cases where a `self` write happens, e.g., dataclass-like structures should not overwrite the `self` attr with a mutable default either, instead write the replacement default to `self`.

eugenio-mercuriali:
Hi @fkiraly!

I would be interested in picking up this issue - should these changes be applied to all subdirectories within the same PR?

Thanks!

fkiraly:
@eugenio-mercuriali, great, all yours!

Re changes, up to you - you can handle some, or all of the instances!

eugenio-mercuriali:
@fkiraly This issue was resolved by the above PR (https://github.com/sktime/pytorch-forecasting/pull/1699). Could you please assign this issue to me + shall we close it?

fkiraly:
ah yes - this issue was not automatically closed.

Thanks for the reminder, @eugenio-mercuriali.

Pro tip: if you use closing keywords such as "fixes #abc", the issue is linked to the PR and closed automatically when merged. Otherwise, you can manually link the issue using the GitHub GUI on the right, which has the same effect upon merge.

PR #1699, opened 2024-10-20
[MNT] Remove mutable objects from defaults

eugenio-mercuriali:
### Description

This PR is a fix for #1668. Removes mutable default arguments and replaces them with internal newly initialized mutable defaults.

### Checklist

- [x] Added/modified tests


fkiraly (changes requested):
Nice!

I think this is great and can be in a single PR.

Some small issues:

* the tests are failing due to type annotations, have a look at the logs
* not a strict requirement, but it might be safer in case of `__init__` to write the resolved argument to a parameter `self._parametername`, to ensure `self.parametername` is always identical with the `__init__` input. This is a "dataclass-like" interface which may be useful later in interfacing sklearn and similar libraries.

eugenio-mercuriali:
> Nice!
> 
> I think this is great and can be in a single PR.
> 
> Some small issues:
> 
> * the tests are failing due to type annotations, have a look at the logs
> * not a strict requirement, but it might be safer in case of `__init__` to write the resolved argument to a parameter `self._parametername`, to ensure `self.parametername` is always identical with the `__init__` input. This is a "dataclass-like" interface which may be useful later in interfacing sklearn and similar libraries.

I see, thanks for the feedback!
Added some changes - how does this look?

fkiraly (changes requested):
Looks great!

Though, for the `__init__`-s, you would *still* do `self.paramname = paramname`, just not change it, e.g.,

```python
self.groups = groups
self._groups = do_sth_with_groups(groups)
```

so the class behaves like a `dataclass` with respect to its init parameters.

There were also a lot of linting failures related to the introduction of stricter requirements in `ruff`, I fixed most of them. The remaining ones seem genuine code issues, could you have ea look?

eugenio-mercuriali:
> Looks great!
> 
> Though, for the `__init__`-s, you would _still_ do `self.paramname = paramname`, just not change it, e.g.,
> 
> ```python
> self.groups = groups
> self._groups = do_sth_with_groups(groups)
> ```
> 
> so the class behaves like a `dataclass` with respect to its init parameters.
> 
> There were also a lot of linting failures related to the introduction of stricter requirements in `ruff`, I fixed most of them. The remaining ones seem genuine code issues, could you have ea look?

Just for me to understand, let's take the `__init__` of _encoders.py_ as an example:
if instead of:

`self._groups = list(groups) if groups is not None else []`

we do:

```
self.groups = groups
self._groups = list(groups) if groups is not None else []
```

then we would have to refactor usages of `self.groups` to `self._groups` in order to use the resolved argument within the class (+ no need to create a _groups_ property) - is this what you suggest?

fkiraly:
> is this what you suggest?

Yes, exactly what I mean.
I confirm your understanding agrees with mine, @eugenio-mercuriali.

The high-level rationale is that the class then behaves, at all times of usage, like a `dataclass` with respect to its init parameters.

eugenio-mercuriali:
> > is this what you suggest?
> 
> Yes, exactly what I mean. I confirm your understanding agrees with mine, @eugenio-mercuriali.
> 
> The high-level rationale is that the class then behaves, at all times of usage, like a `dataclass` with respect to its init parameters.

@fkiraly added changes based on the above. Could you take a look?

eugenio-mercuriali:
@fkiraly are there any adjustments needed or good to merge?

PR #1704, opened 2024-10-26
[MNT] Delete poetry.lock

benHeid:
Delete poetry lock file or alternatively implement a regular update strategy.

### Description

Poetry lock files are not required for package developers and are ignored during installation in dependency resolution if installed via pypi if I understand the following correctly: https://github.com/python-poetry/poetry/blob/d428e79fd22ea6852151ee300ccd980386dd4814/docs/basic-usage.md?plain=1#L314

Thus, there is no harm in having the poetry.lock file but if we have it, we should update it on a regular basis. Perhaps with a GithubAction? Or something similar. 

Up for discussion what is preferable.


fkiraly (approved):
Agreed

Issue #1708, opened 2024-11-09
[MNT] readthedocs build preview in CI


fkiraly:
We should replace the "build" CI step with a full documentation preview from readthedocs, which allows to view the built documentation in the browser.

This should be as in `sktime` .

Suggested by @yarnabrina. I have forgotten how to set this up though, you do this once per repo, and maybe it was not me who set it up...

@yarnabrina, @julian-fong, do you know how to enable this?

yarnabrina:
It should be possible by the admin, so you (and may be CC? not sure). The steps are documented [here](https://docs.readthedocs.io/en/stable/guides/pull-requests.html).

![image](https://github.com/user-attachments/assets/8f21dd59-e9fd-4042-98c2-83d5c55615a1)

fkiraly:
ok, I have tried this.

Re current state, FYI: only I and @jdb78 are admins for the readthedocs. We need to figure out the governance structure for individual projects under the umbrella and how asset ownership relates to the incorporations. We are setting up a working group for this, now that the German umbrella is operational.

fkiraly:
Works, tested in #1712.

PR #1712, opened 2024-11-10
[MNT] remove docs build in ci for all pr

yarnabrina:
As per latest status of #1708, it is configured that Read the Docs build will run in al pull requests. So, it should be unnecessary to do it once more as part of Github CI.


fkiraly (approved):
👍

Issue #1722, opened 2024-11-26
[DOC] Incorrect name on tutorial "Demand forecasting with the Temporal Fusion Transformer"


mbisan:
I've been trying to follow the tutorial on [Demand forecasting with the Temporal Fusion Transformer](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html), but found this error in the documentation:

On the step [Find optimal learning rate](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html#Find-optimal-learning-rate), the `optimizer` parameter passed to `from_dataset` is `"Ranger"`, with capital R.

Apparently, the [base model](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base_model.py#442) class only admits the following (version 1.2.0):

```
optimizer (str): Optimizer, "ranger", "sgd", "adam", "adamw" or class name of optimizer in ``torch.optim``
                or ``pytorch_optimizer``.
                Alternatively, a class or function can be passed which takes parameters as first argument and
                a `lr` argument (optionally also `weight_decay`). Defaults to
                `"ranger" <https://pytorch-optimizers.readthedocs.io/en/latest/optimizer_api.html#ranger21>`_,
                if pytorch_optimizer is installed, otherwise "adam".
```

So the value passed must be changed to "ranger".

The current code causes a `ValueError: Optimizer of self.hparams.optimizer=Ranger unknown``.

The same error appears on the second time `from_dataset` is used, on the [Train model](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html#Train-model) step.

fnhirwa:
Nice catch @mbisan 😊
I am going to open a PR for a fix

PR #1724, opened 2024-12-09
[DOC] Correct argument for optimizer `ranger` in `Temporal Fusion Transformer` tutorial

fnhirwa:
fixes error in the tutorial  which was causing examples to fail.
fixes https://github.com/sktime/pytorch-forecasting/issues/1722


fkiraly (changes requested):
the notebook CI failed, looks like a timeout? I will restart

fnhirwa:
The issue was the training of the `TFT` model which is taking longer than `nbsphinx` timeout time which is `600ms` I run the example locally and disabled notebooks run on CI in `conf.py` You can check the notebook output from ReviewNB before merging.

docs/source/conf.py
```diff
@@ -180,3 +180,8 @@ def setup(app: Sphinx):
 suppress_warnings = [
     "autosummary.import_cycle",
 ]
+
+# -----------nbsphinx extension ----------
+nbsphinx_execute = "never"  # always
```
yarnabrina, line 190: If you never run the notebooks, do the below the configurations have any effect?

docs/source/tutorials/stallion.ipynb
yarnabrina, line 1: I'm unable to see this diff in Github (it saus unable to render rich diff), and ReviewNB also seems to lack the notebook.

docs/source/conf.py
fnhirwa: I found that there is a workflow run for testing the notebook and it was one causing the timeout error, this is for building the docs.

docs/source/tutorials/stallion.ipynb
fnhirwa: I'm not quite sure about that, I logged in with my github and was able to see diff on ReviewNB
image

docs/source/conf.py
yarnabrina: If I am not wrong, `nbsphinx_execute = "never"` will ensure that RTD will not execute the example notebooks while building the docs (ref. https://nbsphinx.readthedocs.io/en/0.9.5/never-execute.html). If that is the case, whether that execution need to ignore errors or not (`nbsphinx_allow_errors = False`) or how long execution takes (`nbsphinx_timeout = 600`) should not matter.

Does that make sense? Let me know if I am missing something, or misunedrstanding the purpose of these configurations.

fnhirwa: You're right, these extra configurations are meaningless.
should we keep the `nbsphinx_execute = "never"`? as we run notebooks in this workflow to test if they don't fail.https://github.com/sktime/pytorch-forecasting/blob/e08f4e6d23737c36949d704215ccbbbaa84fc1b6/.github/workflows/test.yml#L17

docs/source/tutorials/stallion.ipynb
yarnabrina: Not sure why it's not working for me, but not a big deal. I checked it from branch it directly.

I am not what change of yours is making the predictions worse. In `main` branch, the predicted orange line and shaded area seemed to contain the blue actuals pretty well, but in the modified one the gaps seem to have increased notably. Do you agree with observation? Any guess what's causing it (seed remained same at 42 in both cases)?

docs/source/conf.py
yarnabrina: I believe `never` is fine, these executions increase RTD time significantly. Let's skip in `sktime` too for that matter, if @fkiraly agree as well (I have seen too many timeouts recently).

docs/source/tutorials/stallion.ipynb
fnhirwa: I'm not sure about the cause as I can see even the learning_rate got changed from `0.041` to `0.009`.  I think I'll need to check the logs and see where it is going wrong

fnhirwa: I rerun the notebook now the losses seems to have been reduced

fkiraly (approved):
Great!

Question, non-blocking: can we perhaps change the examples that they are still meaningful, but have lower runtime?

fnhirwa:
> Great!
> 
> Question, non-blocking: can we perhaps change the examples that they are still meaningful, but have lower runtime?

Yeah, this sounds great we can do it.

fnhirwa:
I think the best approach is to have a broader task for reworking the examples as the ones available now seem to be written a long time ago. With this we can ensure that we have low runtime working examples.

Issue #1721, opened 2024-11-23
[BUG]  train TFT model on MAC M1 mps device.  element 0 of tensors does not require grad and does not have a grad_fn


zy636:
**Describe the bug**

While following the tutorial at [Demand forecasting with the Temporal Fusion Transformer](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html), I encountered the following issue: the error occurs only when calling pl.Trainer and changing the accelerator from 'cpu' to 'mps':

```
File ~/miniconda3/envs/tft/lib/python3.12/site-packages/torch/autograd/graph.py:825, in _engine_run_backward(t_outputs, *args, **kwargs)
    823     unregister_hooks = _register_logging_hooks_on_whole_graph(t_outputs)
    824 try:
--> 825     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
    826         t_outputs, *args, **kwargs
    827     )  # Calls into the C++ engine to run the backward pass
    828 finally:
    829     if attach_logging_hooks:

RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

**To Reproduce**

```python
trainer2 = pl.Trainer(
    max_epochs=50,
    accelerator="mps",  # change from "cpu" to "mps"
    enable_model_summary=True,
    gradient_clip_val=0.1,
    limit_train_batches=50,  # coment in for training, running valiation every 30 batches
    # fast_dev_run=True,  # comment in to check that networkor dataset has no serious bugs
    callbacks=[lr_logger, early_stop_callback],
    logger=logger,
)

tft2 = TemporalFusionTransformer.from_dataset(
    training,
    learning_rate=0.03,
    hidden_size=16,
    attention_head_size=2,
    dropout=0.1,
    hidden_continuous_size=8,
    loss=QuantileLoss(),
    log_interval=10,  # uncomment for learning rate finder and otherwise, e.g. to 10 for logging every 10 batches
    optimizer="ranger",
    reduce_on_plateau_patience=4,
)
print(f"Number of parameters in network: {tft2.size() / 1e3:.1f}k")

trainer2.fit(
    tft2,
    train_dataloaders=train_dataloader,
    val_dataloaders=val_dataloader,
)
```

**Expected behavior**

**Additional context**

```
trainer2 = pl.Trainer(
   ...:     max_epochs=50,
   ...:     accelerator="mps",
   ...:     enable_model_summary=True,
[...]

PR #1725, opened 2024-12-09
[BUG] Fix issue when training TFT model on mac M1 mps device. element 0 of tensors does not require grad and does not have a grad_fn

### LINKED ISSUE #1756: [BUG] TFT returns NaN values in raw outputs
### **Describe the bug**

Hi, I am trying to use the library for a personal project.
However, I noticed after training that the outputs were all NaN values.

To make sure the issue wasn't from the data or implementation, I copy pasted the code found in the following page: 
https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html

However, I see that the outputs are also all NaN instead of real values like in the tutorial.

Namely, 
```
raw_predictions = best_tft.predict(val_dataloader, mode="raw", return_x=True)
print(raw_predictions.output)
```
Returns tensors of NaNs:
```
Output(prediction=tensor([[[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        [[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        [[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan]],

        ...,

        [[nan, nan, nan,  ..., nan, nan, nan],
         [nan, nan, nan,  ..., nan, nan, nan],
...
```

I noticed that this doesn't happen when I set the architecture hyperparameters to 1:
```
hidden_size=1, 
attention_head_size=1,
hidden_continuous_size=1,
```
Results aren't NaN values, even though they are pretty bad in terms of predictive performance.

So I suspect this isn't an error linked to the library, but either versions of modules or something due to the fact that I run the code on a mac.

Anyone got an idea of what is happening? Thanks!

### **To Reproduce**

Code can be found at:
https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/stallion.html

Before importing `torch`, add:
```
import os
os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
```
Otherwise it returns an error because it isn't implemented yet on MPS.

### **Versions**

<details>

```
System:
    python: 3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)]
executable: [/usr/local/bin/python3](https://file+.vscode-resource.vscode-cdn.net/usr/local/bin/python3)
   machine: macOS-15.2-arm64-arm-64bit

Python dependencies:
          pip: 24.3.1
       sktime: 0.35.0
      sklearn: 1.5.2
       skbase: 0.12.0
        numpy: 2.0.2
        scipy: 1.15.1
       pandas: 2.2.3
   matplotlib: 3.9.2
       joblib: 1.4.2
        numba: None
  statsmodels: 0.14.4
     pmdarima: None
statsforecast: None
      tsfresh: None
      tslearn: None
        torch: 2.5.0
   tensorflow: 2.18.0
```
Also, I am using a Macbook Pro M2.

</details>

fnhirwa:
This PR partially resolves an issue with training when we enable MPS fallback where tensors are being detached from the computation graph, this fix works for all optimizers supported in https://github.com/sktime/pytorch-forecasting/blob/d57b0bb84b07c54a9bdbb2f1b3bd025426109efc/pytorch_forecasting/models/base_model.py#L442 except for `ranger`

part of https://github.com/sktime/pytorch-forecasting/issues/1721 
fixes #1756


fnhirwa:
pinging @zy636 

This seems to be a problem with `Ranger21` optimizer on the `MPS` accelerator the proposed solution works for other optimizers. 

When PyTorch detects an unsupported operation on MPS, it silently switches to CPU for that operation. However:

1. **Tensor detachment:** The switch between devices breaks the link in the computation graph.
2. **Non-differentiable parameters:** Parameters involved in unsupported operations may no longer be tracked by PyTorch's autograd.
3. **Gradient issues:** Ranger21 depends on having proper gradients for updates. If any parameter is non-differentiable or detached, it results in failures like ZeroParameterSizeError. 

For now, I would advise using CPU for training, and if you want to use other optimizers once this PR is merged they'll work, I tested with `adam`, `adamw` and `sgd`.

Thanks

fkiraly (changes requested):
Commenting from a general best practice point - when fixing a bug and all tests passed before the fix, we should add a test tha would have failed before the fix, but passes afterwards ("regression" testing).

fnhirwa:
> Commenting from a general best practice point - when fixing a bug and all tests passed before the fix, we should add a test tha would have failed before the fix, but passes afterwards ("regression" testing).

this was failing on macOS when using `mps` as an accelerator to run the stallion example notebook, a unit test can't be provided for this case because we aren't using MPS on CI as discussed in https://github.com/sktime/pytorch-forecasting/pull/1648, however I'll comment with working example on this PR.

fnhirwa:
The following example works well on MPS accelerator:
```python
import os
os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"

import copy
from pathlib import Path
import warnings

from lightning.pytorch.callbacks import EarlyStopping, LearningRateMonitor
from lightning.pytorch.loggers import TensorBoardLogger
import numpy as np
import pandas as pd
import torch

from pytorch_forecasting import Baseline, TemporalFusionTransformer, TimeSeriesDataSet
from pytorch_forecasting.data import GroupNormalizer
from pytorch_forecasting.metrics import MAE, SMAPE, PoissonLoss, QuantileLoss
from pytorch_forecasting.models.temporal_fusion_transformer.tuning import optimize_hyperparameters
from lightning.pytorch.tuner import Tuner


warnings.filterwarnings("ignore")

import lightning.pytorch as pl

pl.seed_everything(42)

from pytorch_forecasting.data.examples import get_stallion_data

data = get_stallion_data()

# add time index
data["time_idx"] = data["date"].dt.year * 12 + data["date"].dt.month
data["time_idx"] -= data["time_idx"].min()

# add additional features
data["month"] = data.date.dt.month.astype(str).astype("category")  # categories have be strings
data["log_volume"] = np.log(data.volume + 1e-8)
data["avg_volume_by_sku"] = data.groupby(["time_idx", "sku"], observed=True).volume.transform("mean")
data["avg_volume_by_agency"] = data.groupby(["time_idx", "agency"], observed=True).volume.transform("mean")

# we want to encode special days as one variable and thus need to first reverse one-hot encoding
special_days = [
    "easter_day",
    "good_friday",
    "new_year",
    "christmas",
    "labor_day",
    "independence_day",
    "revolution_day_memorial",
    "regional_games",
    "fifa_u_17_world_cup",
    "football_gold_cup",
    "beer_capital",
    "music_fest",
]
data[special_days] = data[special_days].apply(lambda x: x.map({0: "-", 1: x.name})).astype("category")

max_prediction_length = 6
max_encoder_length = 24
training_cutoff = data["time_idx"].max() - max_prediction_length

training = TimeSeriesDataSet(
    data[lambda x: x.time_idx <= training_cutoff],
    time_idx="time_idx",
    target="volume",
    group_ids=["agency", "sku"],
    min_encoder_length=max_encoder_length // 2,  # keep encoder length long (as it is in the validation set)
    max_encoder_length=max_encoder_length,
    min_prediction_length=1,
    max_prediction_length=max_prediction_length,
    static_categoricals=["agency", "sku"],
    static_reals=["avg_population_2017", "avg_yearly_household_income_2017"],
    time_varying_known_categoricals=["special_days", "month"],
    variable_groups={"special_days": special_days},  # group of categorical variables can be treated as one variable
    time_varying_known_reals=["time_idx", "price_regular", "discount_in_percent"],
    time_varying_unknown_categoricals=[],
    time_varying_unknown_reals=[
        "volume",
        "log_volume",
        "industry_volume",
        "soda_volume",
        "avg_max_temp",
        "avg_volume_by_agency",
        "avg_volume_by_sku",
    ],
    target_normalizer=GroupNormalizer(
        groups=["agency", "sku"], transformation="softplus"
    ),  # use softplus and normalize by group
    add_relative_time_idx=True,
    add_target_scales=True,
    add_encoder_length=True,
)

# create validation set (predict=True) which means to predict the last max_prediction_length points in time
# for each series
validation = TimeSeriesDataSet.from_dataset(training, data, predict=True, stop_randomization=True)

# create dataloaders for model
batch_size = 128  # set this between 32 to 128
train_dataloader = training.to_dataloader(train=True, batch_size=batch_size, num_workers=0)
val_dataloader = validation.to_dataloader(train=False, batch_size=batch_size * 10, num_workers=0)



# configure network and trainer
early_stop_callback = EarlyStopping(monitor="val_loss", min_delta=1e-4, patience=10, verbose=False, mode="min")
lr_logger = LearningRateMonitor()  # log the learning rate
logger = TensorBoardLogger("lightning_logs")  # logging results to a tensorboard


trainer = pl.Trainer(
    max_epochs=1,
    accelerator="mps",
    enable_model_summary=True,
    gradient_clip_val=0.1,
    limit_train_batches=50,  # coment in for training, running valiation every 30 batches
    # fast_dev_run=True,  # comment in to check that networkor dataset has no serious bugs
    callbacks=[lr_logger, early_stop_callback],
    logger=logger,
)

tft = TemporalFusionTransformer.from_dataset(
    training,
    learning_rate=0.03,
    hidden_size=16,
    attention_head_size=2,
    dropout=0.1,
    hidden_continuous_size=8,
    loss=QuantileLoss(),
    log_interval=10,  # uncomment for learning rate finder and otherwise, e.g. to 10 for logging every 10 batches
    optimizer="sgd",
    reduce_on_plateau_patience=4,
)


trainer.fit(
    tft,
    train_dataloaders=train_dataloader,
)

```

moshesimon:
when i try your fix on my code i get:
UserWarning: Loss is not finite. Resetting it to 1e9 warnings.warn("Loss is not finite. Resetting it to 1e9")

and train_loss_step=1.04e+4

but when i just change accelerator to cpu instead of mps i get no warning and train_loss_step=2.220

moshesimon:
so i don't think it solves the problem

fnhirwa:
> when i try your fix on the above code i get: train_loss_epoch=1.42e+6 and pytorch_forecasting/metrics/base_metrics.py:817: UserWarning: Loss is not finite. Resetting it to 1e9 warnings.warn("Loss is not finite. Resetting it to 1e9")
> 
> but when i just change accelerator to cpu instead of mps i get: train_loss_epoch=301.0
> 
> so i Don't think the problem is solved

Oh, I see. This is probably the effect of the mps accelerator. I will continue to debug to see where disparities are originating from. Somehow, the computation graph is being broken.

moshesimon:
thanks so much for working on this! 
I really need it fixed xD

moshesimon:
any luck?

fnhirwa:
> any luck?

sorry for the late reply I'm still digging into this.

But for now, the temporary solution would be disabling `mps` during training so that any code that uses `accelerator="auto"` don't pick `mps` when available. This can be achieved by definind a monkeypatch to the `torch._C._mps_is_available` functions which makes `torch.backends.mps.is_available()` to return False hence it is disabled.

``` python
import torch

def disable_torch_mps_is_available():
    return False

torch._C._mps_is_available = disable_torch_mps_is_available

def main():
    print(torch.backends.mps.is_available())

main()
```

fkiraly:
Questions:

* is the problem possibly related to one of these? https://github.com/sktime/pytorch-forecasting/pull/1648, https://github.com/sktime/pytorch-forecasting/pull/1599, FYI @XinyuWuu
* why do the tests not pick this up?

moshesimon:
I found the issue I think.

there seams to be an issue with line 48 in pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py:

`upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)`

when x is on mps it returns nan
when x is on cpu it works.

my temp fix is:

`def interpolate(self, x):
        if x.device.type == 'mps':
            x = x.to('cpu')
            upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)
            upsampled = upsampled.to('mps')
        else:
            upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)
        if self.trainable:
            upsampled = upsampled * self.gate(self.mask.unsqueeze(0)) * 2.0
        return upsampled`

fnhirwa:
> I found the issue I think.
> 
> there seams to be an issue with line 48 in pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py:
> 
> `upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)`
> 
> when x is on mps it returns nan when x is on cpu it works.
> 
> my temp fix is:
> 
> ```
> def interpolate(self, x):
>         if x.device.type == 'mps':
>             x = x.to('cpu')
>             upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)
>             upsampled = upsampled.to('mps')
>         else:
>             upsampled = F.interpolate(x.unsqueeze(1), self.output_size, mode="linear", align_corners=True).squeeze(1)
>         if self.trainable:
>             upsampled = upsampled * self.gate(self.mask.unsqueeze(0)) * 2.0
>         return upsampled
> ```

Nice Catch I tried it locally and started working.

fnhirwa:
> Questions:
> 
> * is the problem possibly related to one of these? [[MNT] handle `mps backend` for lower versions of pytorch and fix `mps` failure on `macOS-latest` runner #1648](https://github.com/sktime/pytorch-forecasting/pull/1648), [[MNT] Fix dependency issues and CI runners: `numpy2`, `optuna`, `requests`, MacOS MPS #1599](https://github.com/sktime/pytorch-forecasting/pull/1599), FYI @XinyuWuu
> * why do the tests not pick this up?

Our tests run entirely on CPU; due to the limited resources on MacOs runners.  This is  only an issue for users who want to use `mps` accelerator.

moshesimon:
I am worried that spewing devices will slow it down. is there a better fix?

pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py
moshesimon, line 1: just set align_corners=False instead of moving to cpu

maybe add a condition that device=mps to set align_corners=False

pytorch_forecasting/models/base_model.py
moshesimon, line 1: no need for this after align_corners=False fix

moshesimon (changes requested):

pytorch_forecasting/models/base_model.py
fnhirwa: will test this, then merge the PR.

benHeid: I suppose that this parameter is set intentionally to True (default is False). So we should be careful about it, since this might changes result. So changing this should be well explained.

fnhirwa: I agree with you, as it is indicated in the docs when the `align_corners` is set to `False.` It will affect the function output as this mode affects the `linear` mode by preserving the values at the corner pixels.

fnhirwa: Keeping it this way.

fnhirwa:
@fkiraly would you take a look at the changes I will then continue to fix this in the examples notebook.

fkiraly (approved):
Again mps magic...

moshesimon:
i think moving to cpu then back to mps is a sub-optimal solution to this bug...

pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py
moshesimon, line 1: inefficient solution switching devices like that. Setting align_corners=False does the same thing

pytorch_forecasting/models/base_model.py
moshesimon, line 1: the fix to pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py makes this change redundant

pytorch_forecasting/models/temporal_fusion_transformer/sub_modules.py
fnhirwa: I think switching devices should be the option for now given that `align_corners` https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html
determines how input pixels map to output pixels during resizing.

The implementations should be different on different backends which I think we don't have control over as framework users.

@benHeid opened an issue on PytTorch repo to fix this certain issue for mps backend 
https://github.com/pytorch/pytorch/issues/144245

benHeid:
@fnhirwa can you check if we can revert this with the newest fix in PyTorch https://github.com/pytorch/pytorch/issues/144245

Issue #1718, opened 2024-11-20
Improving docs, examples, and tutorials of `pytorch-forecasting`


julian-fong:
This list is a quick bucket list of items that can be potentially improved from what I've seen in the docs. 

As requested by @yarnabrina, we should create simple examples, reducing complexity of the library as much as possible, to encourage users who are not too familiar with `pytorch-forecasting` or time series machine learning in general to try out the library. I think for example, the 'Example' inside https://pytorch-forecasting.readthedocs.io/en/stable/getting-started.html# is way too complicated, and will deter users from using the library because it is too complicated.

Thus for now, I'm proposing we 
- refactor the main example inside the getting-started page 
- improve upon the tutorials [page](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials.html) that is currently setup in the library.

I think these two items can improved in the following way:

- Restructure the tutorials page into sections, with increasing 'complexity' of tutorials.

1. This first section will be the 'introduction' section, which will contain one extremely basic introductory tutorial with minimal code, featuring all the different modules (`TimeSeriesDataset`, different `Model`s, `Trainers`, etc), but with little to no explanations. This basic tutorial is designed to essentially be a minimal walkthrough on how to use the library. The preceding tutorials will explain more in depth about each module. For example, we will have one tutorial explaining `TimeSeriesDataset` and so on.
2. The second section will be "Modelling tutorials", where most of the autoregressive models or forecasting model tutorials will be located here. For example, [this](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/deepar.html) and [this](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/nhits.html).
3. The last section will be used for any other miscellaneous and additional tutorials, for example, [this](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials/building.html)

- For the tutorial in the getting-started page, we can re-use the very first intro notebook that we create to provide an extremely simple tutorial to explain to the user how it works.

I am also proposing that we split the directory that is used to load the [data](https://github.com/sktime/pytorch-forecasting/tree/main/examples/data) into a new directory inside the root folder named `datasets`. This folder will be the primarily method to load datasets for tutorials or for modelling purposes. To reduce duplication, we can just import loading methods from the sktime library.

[...]

Issue #1736, opened 2024-12-20
[API] Redesign towards pytorch-forecasting 2.0


fkiraly:
Discussion thread for API re-design for `pytorch.forecasting` next 1.X and towards 2.0. Comments appreciated from everyone!

Summary of discussion on Dec 20, 2024 and prior, about re-design of `pytorch-forecasting`. 

FYI @agobbifbk, @thawn, @sktime/core-developers.

High-level directions:
* the DSIPTS project (https://github.com/DSIP-FBK/DSIPTS) be used as an experimental branch for `pytorch-forecasting 2.0`. We will need to homogenize interfaces, consolidate design ideas, and ensure downwards compatibility.
* further inspiration can be taken from the `thuml` project, also see https://github.com/sktime/sktime/issues/7243.
* @agobbifbk and a team at FBK with substantial time on this will drive this together with @fkiraly and developers at `sktime`.

High-level features for 2.0 with MoSCoW analysis:
* M: unified model API which is easily extensible and composable, similar to `sktime` and DSIPTS, but as closely to the `pytorch` level as possible. The API need not cover forecasters in general, only `torch` based forecasters.
    * M: unified monitoring and logging API
    * M: extension templates need to be created
    * S: `skbase` can be used to curate the forecasters as records, with tags, etc
    * S: model persistence
    * C: third party extension patterns, so new models can "live" in other repositories or packages, for instance `thuml`
* M: reworked and unified data input API
    * M: support static variables and categoricals
    * S: support for multiple data input locations and formats - pandas, polars, distributed solutions etc
* M: MLops and benchmarking features as in DSIPTS
* S: support for pre-training, model hubs, foundation models, but this could be post-2.0

fkiraly:
Having reviewed multiple code bases - `pytorch-forecasting`, DSIPTS, `neuralforecast`, `thuml`, I have come to understand that the `DataLoader` and `DataSet` conventions are key, in particular the input convention for `forward`. Interestingly, all the above-mentioned packages have different conventions here, and none seems satisfactory. What is probably most promising is a "merge" of `pytorch-forecasting` and DSIPTS.

The model layer will mostly follow the data layer, given that `torch` has an idiosyncratic `forward` interface.

My suggestions for high-level requirements on data loaders:

* easy to use in pure `torch`, detour via `pandas` can be avoided (this is currently possible but not easy)
* support for future-known and unknown, endo- and exogenous, group ID and static variables
[...]

PR #1745, opened 2025-01-01
[ENH] Improving docs, examples, and tutorials of pytorch-forecasting

julian-fong:
relates to #1718 

Implements the proposed changes in above issue.

### DIFF
--- a/docs/source/tutorials.rst
+++ b/docs/source/tutorials.rst
@@ -12,7 +12,7 @@ Chapter 1 - Introduction
 .. _chapter1:
 
 .. list-table::
-   :widths: 25 50 15
+   :widths: 25 50 10
    :header-rows: 1
 
    * - Notebook
@@ -30,6 +30,9 @@ Chapter 2 - Time Series Models
 
 .. _chapter2:
 
+.. list-table::
+   :widths: 25 50 15
+   :header-rows: 1
    * - Notebook
      - Description
      - Link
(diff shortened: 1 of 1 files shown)


fkiraly (commented):
question, since this is not a draft - is this complete or still in progress? The changes currently are only a new table of contents, which is nice but not the entire scope of the issue.

julian-fong:
It's still in progress at the moment! Do drafts also trigger the read the docs build? I'm primarily keeping it in this state so that I can see how it looks like after changes

fkiraly:
yes, CI is triggered both by "ready" and "draft" PR. The "draft" state is mainly cosmetic and for sign posting.

julian-fong:
> yes, CI is triggered both by "ready" and "draft" PR. The "draft" state is mainly cosmetic and for signposting that it is not ready.

Thanks! If you have anymore suggestions, feel free to post it here in or in the issue thread :). I'm not sure if we are planning to refactor the `TimeSeriesDataSet` class, but I feel like storing some pretrained dataset objects somewhere that is easy to load would make the tutorials much more simple

fkiraly:
> I'm not sure if we are planning to refactor the `TimeSeriesDataSet` class, 

Yes, we probably will, see here: https://github.com/sktime/pytorch-forecasting/issues/1736
Comments appreciated, especially on usage!

What did you have in mind for the examples?

Issue #1685, opened 2024-09-23
[ENH] decouple data input form `pandas`, allow `polars`, `dask`, and/or `spark`


fkiraly:
A key limitation of current architecture seems to be the reliance on `pandas` of the input, which limites useability in large data cases.

While `torch` with appropriate backends should be able to handle large data, `pandas` as a container choice, in particular the current instantiation which seems to rely on in-memory, will prove to be the bottleneck.

We should therefore consider and implement support for data backends that scale better, such as `polars`, `dask`, or `spark`, and see how easy it is to get the `pandas` `pyarrow` integration to work.

Architecturally, I think we should:

* build a more abstract data loader layer
* make `pandas` one of multiple potential data soft dependencies
* try to prioritize the solution that would provide us with the quickest "impact for time invested"

Spinachboul:
@fkiraly 
Can I be assigned this issue?
I have a little idea about using `rust` under the hood for integrating polars as one of the dataset.
Also this might be a bit complex given we are handling multiple backend formats
So this might take some time to maintain consistency across all frameworks

fkiraly:
@Spinachboul, this is a very complex issue because it relies on at least a redesign of the API.

You can work on this, but first we need an API written down that would allow multiple backends.

Feel free to suggest something, or participate in the discussions in https://github.com/sktime/pytorch-forecasting/issues/1736 until we have consolidated this!

We will also use some of the meet-ups for the pytorch-forecasting rework.

Spinachboul:
@fkiraly  Ohk then! I am not quite familiar with the abbreviation DSIPTS, could you please give some reference for it?
Also could you direct me towards what code files to look for amendments in the backend for better understanding as I am unfamiliar with the codebase!!
Thanks!

fkiraly:
> I am not quite familiar with the abbreviation DSIPTS, could you please give some reference for it?

There is a link directly at the top of the mentioned issue #1736

> Also could you direct me towards what code files to look for amendments in the backend for better understanding as I am unfamiliar with the codebase!

Primarily `BaseModel` and `TimeSeriesDataSet` in `pytorch-forecasting`.

Spinachboul:
> > I am not quite familiar with the abbreviation DSIPTS, could you please give some reference for it?
> 
> There is a link directly at the top of the mentioned issue #1736
> 
> > Also could you direct me towards what code files to look for amendments in the backend for better understanding as I am unfamiliar with the codebase!
> 
[...]

PR #1746, opened 2025-01-03
[ENH] clean-up refactor of `TimeSeriesDataSet`

fkiraly:
This PR carries out a clean-up refactor of `TimeSeriesDataSet`. No changes are made to the logic.

This is in preparation for major work items impacting the logic, e.g., removal of the `pandas` coupling (see https://github.com/sktime/pytorch-forecasting/issues/1685), or a 2.0 rework (see https://github.com/sktime/pytorch-forecasting/issues/1736).
In general, a clean state would make these easier.

Work carried out:

* clear, and complete docstrings, in numpydoc format
* separating logic, e.g., for parameter checking, data formatting, default handling
* reducing cognitive complexity and max indentations, addressing "code smells"
* linting

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -544,7 +544,7 @@ class TimeSeriesDataSet(Dataset):
         self._data_properties = self._data_properties(data)
 
         # target normalizer
-        self.target_normalizer = self._target_normalizer(
+        self.target_normalizer = self._set_target_normalizer(
             self._data_properties, self.target_normalizer
         )
 
@@ -837,7 +837,7 @@ class TimeSeriesDataSet(Dataset):
         else:
             return max([max(lag) for lag in self._lags.values()])
 
-    def _target_normalizer(self, data_properties, target_normalizer):
+    def _set_target_normalizer(self, data_properties, target_normalizer):
         """Determine target normalizer.
 
         Determines normalizers for variables based on self.target_normalizer setting.
@@ -887,7 +887,7 @@ class TimeSeriesDataSet(Dataset):
             f"class TorchNormalizer but found {target_normalizer}"
         )
         assert not self.multi_target or isinstance(
-            self.target_normalizer, MultiNormalizer
+            target_normalizer, MultiNormalizer
         ), (
             "multiple targets / list of targets requires MultiNormalizer as "
             f"target_normalizer but found {target_normalizer}"
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -922,9 +922,7 @@ class TimeSeriesDataSet(Dataset):
                 else:
                     transformer = None
                 if self.max_encoder_length > 20 and self.min_encoder_length > 1:
-                    normalizers.append(
-                        EncoderNormalizer(transformation=transformer)
-                    )
+                    normalizers.append(EncoderNormalizer(transformation=transformer))
                 else:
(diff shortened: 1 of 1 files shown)


pytorch_forecasting/data/timeseries.py
```diff
@@ -87,32 +94,53 @@ def _find_end_indices(
...
+    ------
+    ValueError
+        if tensor contains NAs or infinite values
+    AssertionError
+        if tensor has incorrect dimension, see above
     """
     if isinstance(names, str):
         names = [names]
-        assert tensor.ndim == 1
+        assert tensor.ndim == 1, names
```
fnhirwa, line 127: Should we be concise with the assertion error here?

pytorch_forecasting/data/timeseries.py
```diff
@@ -418,56 +517,61 @@ def __init__(
...
         self.lags = lags
-        self._lags = {} if lags is None else deepcopy(lags)
+        self._lags = _coerce_to_dict(lags)
 
         # add_encoder_length
         if isinstance(add_encoder_length, str):
-            assert (
-                add_encoder_length == "auto"
-            ), f"Only 'auto' allowed for add_encoder_length but found {add_encoder_length}"
+            msg = (
```
fnhirwa, line 549: Somehow the check for bool instance was ommited here!

fnhirwa (changes requested):
Overall, this clean-up looks good to me 😊

A one point that we should consider:

* I think we can start handling exceptions using `if` conditions instead of `assert` as `if` is optimized for runtime, and for some cases for example if a user is running a python program using optimizer flags(`python -O` or `python -OO`) the `assert` conditions are skipped.

```python
def divide(a, b):
    if b == 0:
        raise ValueError("Denominator cannot be zero.")
    return a / b
print(divide(1, 0) # ValueError: Cannot divide by zero when we do python -O script.py
```
When we use assert

```python
def divide(a, b):
    assert b != 0, "Denominator cannot be zero."
    return a / b
print(divide(2/0)) # ZeroDivisionError: division by zero with python -O
```
For the second case, the exception we wanted to handle was skipped and we got an unexpected  exception which can affect the exceptions handling in general for the case of our codebase.

pytorch_forecasting/data/timeseries.py
fkiraly: well, I added it for diagnostics as it was a fiddly case. Will probably replace with some `raise` later, see below.

pytorch_forecasting/data/timeseries.py
fkiraly: can you explain? Not sure if I get what you mean.

pytorch_forecasting/data/timeseries.py
fnhirwa: This is the same case as raising precise exceptions and errors, it will get solved later as you said above.

pytorch_forecasting/data/timeseries.py
fnhirwa: It is a minor missing conditional check for bool. It shows that `add_encoder_length` can be either `bool` or `"auto"` but only the `str` instance is being checked here.

fkiraly: oh, I see. The `bool` check is in `_check_params` now, and the `str` check I left there because it has a check fused to a parameter change (which we should later address, but that would be change to logic)

fnhirwa: I get it now.
Thanks

PR #1750, opened 2025-01-08
[DOC] fix `gradient_clip` value in tutorials to ensure reproducible outputs similar to the committed cell output

gbilleyPeco:
Update gradient_clip=0.01 to gradient_clip=0.1.

### Description

I've been following along with the Interpretable forecasting with N-Beats tutorial found here https://pytorch-forecasting.readthedocs.io/en/latest/tutorials/ar.html, copy-pasting the code exactly, and I am getting different results compared to what is shown on the documentation page. 

The first discrepancy between my results and the tutorial is when using lightning.pytorch.tuner.Tuner to find the optimal learning rate. The suggested learning rate and resulting plot are different. 

Perhaps more of a problem, however, is that my model fails to learn (training loss doesn't decrease), even when increasing the epochs. When I run the code from the Evaluate Results section, all of the resulting plots show a flat predicted curve.

I believe I've found the problem. Testing on my side confirms. 

Code cells [6] and [14] have the parameter gradient_clip=0.01, which is the source of the problem. The N-HiTS tutorial uses gradient_clip=0.1, so I tried that out, which fixes the problem. The outputs don't match the tutorial outputs exactly, but they're much closer than what gradient_clip=0.01 yields.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


yarnabrina:
Hi @gbilleyPeco , thanks for the PR. Is there any specific objective reason to prefer 0.1 over 0.01, except may be the fact 0.01 is too small? Can you please update the goal of the PR in the description section?

fkiraly (changes requested):
Thanks!

Would you mind copying the description from discord why we are doing this in the first post in this PR, or an issue that you reference?

Sohaib-Ahmed21:
@gbilleyPeco thanks for pointing. Was also working around to get correct predictions and it worked.

fkiraly (approved):
Looks reasonable with the explanation.

Issue #1741, opened 2024-12-26
[ENH] Kolmogorov Arnold Block for NBeats network


benHeid:
**Is your feature request related to a problem? Please describe.**
The following paper implements a KAN block in NBEats. 

https://arxiv.org/pdf/2412.17853

**Describe the solution you'd like**
Implement exchangeable blocks for nbeats. 

Check if the same can also be used for nhits.

Sohaib-Ahmed21:
@benHeid @fkiraly I want to work on this issue. Kindly assign me.

Sohaib-Ahmed21:
> https://arxiv.org/pdf/2412.17853

@benHeid  The paper utilizes **Domain-Adversarial Training of Neural Networks (DANN)** to learn domain-invariant features. Should DANN be incorporated into the implementation, or is it sufficient to simply switch between MLP and KAN blocks? Since DANN operates at a higher level and can be integrated into any model, it might be better suited for the base design.

Currently, my approach involves using a flag (param) in NBeats: when the flag is set to True, KAN blocks are used; otherwise, MLP blocks are used. This enables an interchangeable implementation.

Would appreciate your thoughts or suggestions on this approach.

benHeid:
I would focus only on implementing KAN blocks. 

The paper seems to have two contributions, combining NBEATS with KAN blocks and then also the DANN for zero-shot forecasting, if I understand it correctly.

Sohaib-Ahmed21:
> I would focus only on implementing KAN blocks.
> 
> The paper seems to have two contributions, combining NBEATS with KAN blocks and then also the DANN for zero-shot forecasting, if I understand it correctly.

Thanks for your input. 

Btw what are  your thoughts about having DANN in sktime for zero-shot forecasting tasks where we have primary and secondary domains in training?

benHeid:
To be honest, I am not really familiar with DANN and currently, I am not that convinced by DANN especially if you consider that the foundation models achieve something similar without doing DANN. Furthermore, I fear that this might be more complicated to implement it in a proper way. However, if you would like to implement it and you have an idea on how this would look like from an architectural perspective you are welcome to do it. But before starting to implement it, I suppose that we need to discuss API design for that. 

What are your thought with regard to DANN?

Sohaib-Ahmed21:
> However, if you would like to implement it and you have an idea on how this would look like from an architectural perspective you are welcome to do it. But before starting to implement it, I suppose that we need to discuss API design for that.
[...]

PR #1751, opened 2025-01-13
[ENH] Kolmogorov Arnold Block for NBeats

Sohaib-Ahmed21:
### Description
Fixes: #1741 

This PR adds Kolmogorov Arnold(KAN) Blocks in NBeats and also does refactoring of NBeats. Implementation of KAN blocks' layers is taken from [original paper code](https://github.com/KindXiaoming/pykan/tree/master). 

### Changes in Structure

* Introduced the `NBEATSKAN` module, which enables usage of **KAN blocks** within the `NBEATS` architecture.

* Integrated `KANLayer` logic, implemented in `kan_layer.py`, which handles KAN-specific operations such as:

  * **Spline coefficient computation**,
  * **Grid initialization and updates**, etc.

* Imported `KANLayer` to `submodules.py` for block operations, allowing `NBEATSKAN` to delegate block-level behavior through `use_kan=True`.

* Added the `NBEATSAdapter` class to encapsulate **common methods** shared by both `NBEATS` and `NBEATSKAN`, including:

  * Standard training, forward logic, etc.
  * **Excludes** block initialization (`__init__`), which is separately defined in each class to maintain architectural flexibility.
  
### GridUpdateCallback

When training **KAN-based models**, the grid can be **iteratively refined** during training for better performance.

To support this, logic from the original [`[pykan](https://github.com/KindXiaoming/pykan)`](https://github.com/KindXiaoming/pykan) implementation has been adapted to define a **custom callback** named `GridUpdateCallback`. This callback automatically updates the grid at specified training steps, improving model accuracy and convergence.

This callback has been **tested successfully** and demonstrates **improved results** in practice.

An example usage is provided in:
`examples/nbeats_with_kan.py`

### Checklist

Make sure to have fun coding!


Sohaib-Ahmed21:
@fkiraly @benHeid can you kindly review/merge it so I integrate **NBEATSX** modification in NBEATS without conflicts as I have asked @julian-fong and he is not working on NBEATSX.

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -30,6 +30,19 @@ def __init__(
         expansion_coefficient_lengths: Optional[List[int]] = None,
         prediction_length: int = 1,
         context_length: int = 1,
+        use_kan: bool = False,
```
benHeid: Can you ad these parameters at the end of the old parameters (directly before kwargs)? This improves the backward compatibility e.g. in case a user provides arguments instead of key-word arguments.

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -30,6 +30,19 @@ def __init__(
...
+        num_grids: int = 5,
+        k: int = 3,
+        noise_scale: float = 0.5,
+        scale_base_mu: float = 0.0,
+        scale_base_sigma: float = 1.0,
+        scale_sp: float = 1.0,
+        base_fun: callable = torch.nn.SiLU(),
```
benHeid: Please avoid all mutable objects as defaults. Since changing these mutable objects might lead to side-effects that are difficult to debug.

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -30,6 +30,19 @@ def __init__(
...
+        grid_eps: float = 0.02,
+        grid_range: List[int] = [-1, 1],
```
benHeid: Please avoid all mutable objects as defaults. Since changing these mutable objects might lead to side-effects that are difficult to debug.

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -47,48 +60,86 @@ def __init__(
 
         Based on the article
         `N-BEATS: Neural basis expansion analysis for interpretable time series
-        forecasting <http://arxiv.org/abs/1905.10437>`_. The network has (if used as ensemble) outperformed all
-        other methods
-        including ensembles of traditional statical methods in the M4 competition. The M4 competition is arguably
-        the most
-        important benchmark for univariate time series forecasting.
+        forecasting <http://arxiv.org/abs/1905.10437>`_. The network has (if
```
benHeid: Why are that many docstrings changed? Was this a linter?

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -47,48 +60,86 @@ def __init__(
...
                 interpretable mode: [3]
             prediction_length: Length of the prediction. Also known as 'horizon'.
-            context_length: Number of time units that condition the predictions. Also known as 'lookback period'.
+            context_length: Number of time units that condition the predictions.
+                Also known as 'lookback period'.
                 Should be between 1-10 times the prediction length.
-            backcast_loss_ratio: weight of backcast in comparison to forecast when calculating the loss.
-                A weight of 1.0 means that forecast and backcast loss is weighted the same (regardless of backcast and
-                forecast lengths). Defaults to 0.0, i.e. no weight.
+            num_grids :  Parameter for KAN layer. the number of grid intervals = G.
```
benHeid: Unsure I have overseen the description of `use_kan`. Please check.

benHeid: For all parameters that are only used if use_kan == True, please add a statement that these are ignored if kan is not set to True.

pytorch_forecasting/models/nbeats/kan_layer.py
```diff
@@ -0,0 +1,528 @@
+import numpy as np
```
benHeid: The license of the original implementation is MIT. So in theory it is okay to copy the file. However, please add some credits at the top of the file.

Alternatively, we could think about adding KAN as a dependency. 

@fkiraly do you have any additions on that matter?

benHeid (changes requested):
Not fully reviewed yet. Will continue in the next days. But I share my current comments so that you already receive some feedbacks. 

When training KANs, the grid can be iteratively be refined. I wonder, if there is a way to implement this also here. However, this might probably more difficult and require changes to the trainer. So probably out of scope for this PR. Do you have opinions on that?

pytorch_forecasting/models/nbeats/_nbeats.py
Sohaib-Ahmed21: yes before lines in docstring had >88 characters but linter didn't allow this much length before committing this time.

pytorch_forecasting/models/nbeats/_nbeats.py
Sohaib-Ahmed21: Oh I forgot to add, will add it.

pytorch_forecasting/models/nbeats/kan_layer.py
Sohaib-Ahmed21: Okay, I will add the appropriate credits at the top of the file. Additionally, I agree that adding KAN as a dependency—perhaps as a soft dependency—seems like a good idea, especially considering its increasing relevance in time series forecasting.

Sohaib-Ahmed21:
> Not fully reviewed yet. Will continue in the next days. But I share my current comments so that you already receive some feedbacks.

Thanks! Will address these reviews soon.

Sohaib-Ahmed21:
> When training KANs, the grid can be iteratively be refined. I wonder, if there is a way to implement this also here. However, this might probably more difficult and require changes to the trainer. So probably out of scope for this PR. Do you have opinions on that?

I'll explore this and share my thoughts.

Sohaib-Ahmed21:
@benHeid I have addressed the reviews. Kindly review the updated PR.

> When training KANs, the grid can be iteratively be refined. I wonder, if there is a way to implement this also here. However, this > might probably more difficult and require changes to the trainer. So probably out of scope for this PR. Do you have opinions on > that?

To address this, I have taken logic from original implementation of **pykan** library and made custom Callback i.e. `GridUpdateCallback` which updates grid after specified steps. This has been tested and working fine. Your thoughts will be helpful in this regard.

Sohaib-Ahmed21:
@benHeid @fkiraly  I have addressed the reviews.  Kindly review it so I proceed integrating NBEATSX without conflicts.

pytorch_forecasting/models/nbeats/kan_layer.py
benHeid: @fkiraly pinging you again to check if this is okay for you :)

pytorch_forecasting/models/nbeats/grid_callback.py
benHeid, line 1: I would appreciate if you can add an example on how to use it. Either in the docstring or in a notebook. Otherwise it is implemented but unclear on how to use it.

pytorch_forecasting/models/nbeats/kan_layer.py
```diff
@@ -0,0 +1,367 @@
...
+            grid_adaptive = x_pos[ids, :].permute(1, 0)
+            margin = 0.00
+            h = (
+                grid_adaptive[:, [-1]] - grid_adaptive[:, [0]] + 2 * margin
+            ) / num_interval
+            grid_uniform = (
+                grid_adaptive[:, [0]]
+                - margin
+                + h
+                * torch.arange(
```
benHeid: Ensure that the output of torch.arange is on the same device as h and margin. E.g., if everything is on cuda or mps, this will fail since the default is that the tensor from torch.arange is on cpu.

benHeid (commented):
I only have one last comment. But I also would like to hear @fkiraly opinion on the license of the kan_layer

pytorch_forecasting/models/nbeats/kan_layer.py
fkiraly: which package are we exactly planning to add as a soft dep?

If it is a single layer, I think copying it over and including the license is perhaps better for now, because we do not have machinery to manage soft dependencies (like `scikit-base` or similar).

The proposed design in here https://github.com/sktime/enhancement-proposals/pull/39 would allow that, but right now I think this would require a significant amounts of custom code to handle.

Or, is there an easy way that I am not seeing how the soft dependency import would work for part of the NN?

Sohaib-Ahmed21: > which package are we exactly planning to add as a soft dep?
> 
It is `pykan` library , reference https://pypi.org/project/pykan/

> If it is a single layer, I think copying it over and including the license is perhaps better for now, because we do not have machinery to manage soft dependencies (like `scikit-base` or similar).
> 
> The proposed design in here [sktime/enhancement-proposals#39](https://github.com/sktime/enhancement-proposals/pull/39) would allow that, but right now I think this would require a significant amounts of custom code to handle.
> 
> Or, is there an easy way that I am not seeing how the soft dependency import would work for part of the NN?

yes it is a single layer, only used in `NBEATS`. Also the library `pykan` has much more, but we only need this. I have implemented what you already suggested above. I have copied it over and included the license.

fkiraly: ok, makes sense, as long as the license points to the original source and is provided in full form (assuming it hsa the usual reqiurement to reproduce)

fkiraly (changes requested):
I do have one major request though: This PR seems to be changing the parametrization of `NBEATS` in major ways, and it essentially alters the structure.

May I hence request to, instead of modifying the existing class, to add a new class?
There may be common code that could be refactored, I just think that it is better to keep the class `NBEATS` - a common method - as a conceptual reference to the original NBEATS algorithm.

Variants and modifications I would provide as new classes - there are many modifications of NBEATS; so it seems like a suboptimal design decision to merge one of those (albeit a popular one) in the reference implementation.

Sohaib-Ahmed21:
> I do have one major request though: This PR seems to be changing the parametrization of `NBEATS` in major ways, and it essentially alters the structure.
> 
> May I hence request to, instead of modifying the existing class, to add a new class? There may be common code that could be refactored, I just think that it is better to keep the class `NBEATS` - a common method - as a conceptual reference to the original NBEATS algorithm.
> 
> Variants and modifications I would provide as new classes - there are many modifications of NBEATS; so it seems like a suboptimal design decision to merge one of those (albeit a popular one) in the reference implementation.

Sounds good, thanks for the review! I will make separate class for NBEATS using KAN keeping the original algorithm separate and will update the PR.

fkiraly:
FYI @benHeid, thoughts/opinions?

benHeid:
I see your point. However, I fear that we will introduce a lot of duplicated code. Thus, we might think about refactoring strategies.

But I am also happy with a new class.

Sohaib-Ahmed21:
> I see your point. However, I fear that we will introduce a lot of duplicated code. Thus, we might think about refactoring strategies.
> 
> But I am also happy with a new class, since this reduces also the complexity.

I'll try to reuse as much code as possible like submodules, etc. So should I move towards new class implementation?

Sohaib-Ahmed21:
@fkiraly @benHeid kindly review the udpated changes.  Thanks!

fkiraly (changes requested):
LGTM - could we make sure that the new modules in `nbeats` are private? That gives us more flexibility for downwards compatible changes, later.

Sohaib-Ahmed21:
> LGTM - could we make sure that the new modules in `nbeats` are private? That gives us more flexibility for downwards compatible changes, later.

Thanks for the review, I have made the changes and I think it's ready for merge now.

Sohaib-Ahmed21:
@fkiraly this seems ready for merge. Kindly review/merge it. Thanks!

Sohaib-Ahmed21:
The failing test seems unrelated to PR. Also, all reviews have been addressed here. Kindly review/merge it. Thanks!

Sohaib-Ahmed21:
This PR seems ready for merge. @fkiraly @benHeid kindly review/merge it. Thanks!

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+            If True, k points are extended on both ends. If False, no extension
+            (zero boundary condition). Default: True
+
+    Returns:
+    --------
+        spline values : 3D torch.tensor
+            shape (batch, in_dim, G+k). G: the number of grid intervals,
+            k: spline order.
+
+    Example
```
benHeid: This example seems not to work. I guess it is code from the original implementation.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
+# The following implementation of KANLayer is inspired by the pykan library.
+# Reference: https://github.com/KindXiaoming/pykan/blob/master/kan/KANLayer.py
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+
+def B_batch(x, grid, k=0, extend=True):
```
benHeid: Please rename the method to lower case. (Normally, we name all methods with lower cases)

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    evaludate x on B-spline bases
```
benHeid: Type in `evaludate`

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+    mask = torch.zeros(in_dim, out_dim)
+    mask[all_connection[:, 0], all_connection[:, 1]] = 1.0
+
+    return mask
+
+
+class KANLayer(nn.Module):
+    """
+    KANLayer class
+    """
```
benHeid: Please provide a description of the parameters.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+        self,
+        in_dim=3,
+        out_dim=2,
+        num=5,
+        k=3,
+        noise_scale=0.5,
+        scale_base_mu=0.0,
+        scale_base_sigma=1.0,
+        scale_sp=1.0,
+        base_fun=torch.nn.SiLU(),
```
benHeid: Please do not use mutables as default parameters. E.g. use None as default and initials base_fun within the method with SiLU if it is None.
Otherwise, this can introduce side-effects.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+        grid_eps=0.02,
+        grid_range=[-1, 1],
```
benHeid: lists are mutable too. So also use None as default and initialise with the list if the parameter is None.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+            sparse_init : bool
+                if sparse_init = True, sparse initialization is applied.
+
+        Returns:
+        --------
+            self
+
+        Example
+        -------
+        >>> from kan.KANLayer import *
```
benHeid: This example is probably not working correctly, since it is just copied from the original implementation.

pytorch_forecasting/models/nbeats/_kan_layer.py
benHeid: Please remove the extend=True parameter, this is not used.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+    except Exception as e:
+        print(f"lstsq failed with error: {e}")
+
+    return coef
+
+
+def extend_grid(grid, k_extend=0):
+    """
+    extend grid
+    """
```
benHeid: Please describe the parameters

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+        grid = torch.cat([grid[:, [0]] - h, grid], dim=1)
+        grid = torch.cat([grid, grid[:, [-1]] + h], dim=1)
+
+    return grid
+
+
+def sparse_mask(in_dim, out_dim):
+    """
+    get sparse mask
+    """
```
benHeid: Please describe the parameters

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+                fan out x into activations, shape (number of sampels,
+                output dimension, input dimension)
+            postacts : 3D torch.float
+                the outputs of activation functions with preacts as inputs
+            postspline : 3D torch.float
+                the outputs of spline functions with preacts as inputs
+
+        Example
+        -------
+        >>> from kan.KANLayer import *
```
benHeid: This example is probably not working.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+        >>> print(model.grid.data)
+        """
+
+        batch = x.shape[0]
+        x_pos = torch.sort(x, dim=0)[0]
+        y_eval = coef2curve(x_pos, self.grid, self.coef, self.k)
+        num_interval = self.grid.shape[1] - 1 - 2 * self.k
+
+        def get_grid(num_interval):
+            ids = [int(batch / num_interval * i) for i in range(num_interval)] + [-1]
```
benHeid: Please provide a doctoring

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+            grid_adaptive = x_pos[ids, :].permute(1, 0)
+            margin = 0.00
```
benHeid: Why is the margin 0.00. To me, it seems that it can be removed.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def update_grid_from_samples(self, x, mode="sample"):
+        """
+        update grid from samples
+
+        Args:
+        -----
+            x : 2D torch.float
+                inputs, shape (number of samples, input dimension)
+
```
benHeid: Please describe all parameters.

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,364 @@
...
+        y = coef2curve(x_eval=x, grid=self.grid, coef=self.coef, k=self.k)
+        y = (
+            self.scale_base[None, :, :] * base[:, :, None]
+            + self.scale_sp[None, :, :] * y
+        )
+        y = self.mask[None, :, :] * y
+        y = torch.sum(y, dim=1)
+        return y
+
+    def update_grid_from_samples(self, x, mode="sample"):
```
benHeid: I guess you can remove the mode="sample" since this method is only used with the default parameters. Thus, you can also remove the if condition where mode is used.

pytorch_forecasting/models/nbeats/_nbeats.py
```diff
@@ -139,296 +147,9 @@ def __init__(
...
                         forecast_length=prediction_length,
-                        dropout=self.hparams.dropout,
+                        dropout=dropout,
                     )
                 else:
                     raise ValueError(f"Unknown stack type {stack_type}")
 
                 self.net_blocks.append(net_block)
-
-    def forward(self, x: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
```
benHeid: This method is copied into _nbeats_adapter or?

pytorch_forecasting/models/nbeats/_nbeats_adapter.py
benHeid, line 1: Is there anything implemented by you or just copied?

pytorch_forecasting/models/nbeats/sub_modules.py
```diff
@@ -45,31 +53,111 @@ def __init__(
...
+            units: The number of units in the mlp/kan layers.
+            thetas_dim: The dimension of the parameterized output for the block.
+            num_block_layers: Number of fully connected mlp/kan layers. Default: 4.
+            backcast_length: The length of the backcast. Defines how many time units
+                from the past are used to predict the future. Default: 10.
+            forecast_length: The length of the forecast, i.e., the number of time steps
+                ahead to predict. Default: 5.
+            dropout: The dropout rate applied to the fully connected mlp layers to
+                prevent overfitting. Default: 0.1.
+            kan_params (dict): Parameters specific to the KAN layer
```
benHeid: Under the hood, there is a default behaviour for these parameter. Would be cool, if you can describe it here. Or even set some default values if someone sets `use_kan` to true.

pytorch_forecasting/models/nbeats/sub_modules.py
```diff
@@ -84,7 +172,50 @@ def __init__(
...
+            backcast_length: The length of the backcast. Defines how many time units
+                from the past are used to predict the future. Default: 10.
+            forecast_length: The length of the forecast, i.e., the number of time steps
+                ahead to predict. Default: 5.
+            nb_harmonics: The number of harmonics in the seasonal function (relevant for
+                seasonal models). Default: None (no seasonality).
+            min_period: The minimum period used for seasonal patterns. Default: 1.
+            dropout: The dropout rate applied to the fully connected mlp layers to
+                prevent overfitting. Default: 0.1.
+            kan_params (dict): Parameters specific to the KAN layer
```
benHeid: See comment

benHeid (changes requested):
Sorry for the late response. 

I guess in the copied modules there are some issues mainly regarding documentation. And I am also unsure if we are using all of the options in the code. 

A big plus would be if you could comment the code that you copied a bit. Since it is hard to understand

Sohaib-Ahmed21:
@benHeid thanks for the detailed review, I'll address it as soon as I get time and update the PR.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Description seems to be already present for `KANLayer` parameters. Can you please confirm?

pytorch_forecasting/models/nbeats/_nbeats.py
Sohaib-Ahmed21: Yes, it has been copied there.

pytorch_forecasting/models/nbeats/_nbeats_adapter.py
Sohaib-Ahmed21: The code is copied from [_nbeats.py](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nbeats/_nbeats.py). Only the `__init__` method is modified, as it's customized in child classes depending on whether `KAN` variant is used or not.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Enhanced example with detailed comments on how to run and tested it.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Enhanced example with detailed comments on how to run and tested it.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Enhanced example with detailed comments on how to run and tested it.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: You're right, removed it!

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Yes, this was used with default params, removed the unused code!

pytorch_forecasting/models/nbeats/sub_modules.py
Sohaib-Ahmed21: Refactored the comment and initialized `kan_params` with defaults if `use_kan` is true and `kan_params` is None.

Sohaib-Ahmed21:
@benHeid I have addressed the reviews. Kindly review the updated PR,  thanks!

fkiraly (changes requested):
Nice!

Could you kindly also add some explanation in the PR fist post about the changes to the structure, and what goes in the common base class? What is going on with the grid callback?

Also requesting review by @phoeenniixx and @PranavBhatP

Optionally, could you check in the `layers` module whether we can use that to avoid duplicate layer specifications?

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,413 @@
...
+import torch
+import torch.nn as nn
+
+
+def b_batch(x, grid, k=0):
+    """
+    evaluate x on B-spline bases
+
+    Args:
+    -----
```
phoeenniixx: Can you please use `Parameters` like in `sktime` and follow the numpydoc style docstrings, No need to add colon (:)

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,413 @@
...
+        base_fun=None,
+        grid_eps=0.02,
+        grid_range=None,
+        sp_trainable=True,
+        sb_trainable=True,
+        sparse_init=False,
+    ):
+        """'
+        Initialize a KANLayer
+
```
phoeenniixx: Can you move this docstring above just after the class definition?

pytorch_forecasting/models/nbeats/_nbeatskan.py
```diff
@@ -0,0 +1,235 @@
...
+        forecasting <http://arxiv.org/abs/1905.10437>`_. The network has (if
+        used as ensemble) outperformed all other methods including ensembles of
+        traditional statical methods in the M4 competition. The M4 competition is
+        arguably the most important benchmark for univariate time series forecasting.
+
+        The :py:class:`~pytorch_forecasting.models.nhits.NHiTS` network has recently
+        shown to consistently outperform N-BEATS.
+
+        Args:
+            stack_types: One of the following values: “generic”, “seasonality" or
```
phoeenniixx: Please use numpydoc style doctsring

pytorch_forecasting/models/nbeats/_kan_layer.py
```diff
@@ -0,0 +1,413 @@
...
+    -------
+    The following is an example from the original `pykan` library, adapted here
+    for illustration within the PyTorch Forecasting integration.
+
+    Install the `pykan` package first:
+    >>> pip install pykan
+    Then use:
+
+    >>> from kan.spline import B_batch
+    >>> import torch
```
phoeenniixx: Can we have an example for `pytorch_forecasting`'s implementation as I think it will be helpful for users using this implementation :)

pytorch_forecasting/models/nbeats/_nbeatskan.py
```diff
@@ -0,0 +1,235 @@
...
+    ):
+        """
+        Initialize NBeats Model - use its :py:meth:`~from_dataset` method if possible.
+
+        Based on the article
+        `N-BEATS: Neural basis expansion analysis for interpretable time series
+        forecasting <http://arxiv.org/abs/1905.10437>`_. The network has (if
+        used as ensemble) outperformed all other methods including ensembles of
+        traditional statical methods in the M4 competition. The M4 competition is
+        arguably the most important benchmark for univariate time series forecasting.
```
phoeenniixx: I think it would be good to mention the paper for `KAN` blocks here and how it works (just an overview) and how it is different from the original `NBeats`

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: These are the `utils` called by internal code only, so I think current examples will be fine, thoughts?

pytorch_forecasting/models/nbeats/_nbeatskan.py
```diff
@@ -0,0 +1,235 @@
...
+                else:
+                    raise ValueError(f"Unknown stack type {stack_type}")
+
+                self.net_blocks.append(net_block)
+
+    def update_kan_grid(self):
+        """
+        Updates grid of KAN layers when using KAN layers in NBEATSBlock.
+        """
+        for block in self.net_blocks:
```
phoeenniixx, line 282: Where exactly is this function used? I dont see any calls in `NBeatsKAN` or `NBeatsAdapter`. It would be helpful if you could explain that

Sohaib-Ahmed21: I have made a example in `examples` folder for using `GridUpdateCallback`, will refer that here.

pytorch_forecasting/models/nbeats/_nbeatskan_pkg.py
```diff
@@ -0,0 +1,39 @@
...
+    def get_model_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models import NBeatsKAN
+
+        return NBeatsKAN
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Return testing parameter settings for the trainer."""
+        return [{"backcast_loss_ratio": 1.0}]
```
phoeenniixx: Can we have some more params? Just to see any edge cases or anything?

pytorch_forecasting/models/nbeats/_nbeatskan.py
phoeenniixx: Thanks, its clear now

pytorch_forecasting/models/nbeats/_kan_layer.py
phoeenniixx: I think then if you are adding egs, some egs how the `ptf` implementation can be used will also be helpful, not necessarily here, maybe in `NBeatsKAN`? Its just a suggestion -non blocking

pytorch_forecasting/models/nbeats/_nbeatskan.py
Sohaib-Ahmed21: Referred the example in the docstring!

pytorch_forecasting/models/nbeats/_nbeatskan.py
Sohaib-Ahmed21: Mentioned the reference with details.

pytorch_forecasting/models/nbeats/_kan_layer.py
Sohaib-Ahmed21: Added `examples` folder's example link in docstring of `NBEATSKAN`.

Sohaib-Ahmed21:
> Could you kindly also add some explanation in the PR fist post about the changes to the structure, and what goes in the common base class? What is going on with the grid callback?

Updated the first post to include explanations on the structure changes, common base class, and grid callback as requested.

> Optionally, could you check in the layers module whether we can use that to avoid duplicate layer specifications?

I checked it and currently blocks are reused from `sub_modules.py` and are not duplicated.

Sohaib-Ahmed21:
@fkiraly @benHeid kindly review/merge this pull request, thanks!

phoeenniixx:
> > Optionally, could you check in the layers module whether we can use that to avoid duplicate layer specifications?
> 
> I checked it and currently blocks are reused from `sub_modules.py` and are not duplicated.

I think we can also move some of the layers (like `KAN` layer) to the `layers` right? I think these layers can be reused by other models as well? 

Also, a question for @fkiraly: Imo, we could also move networks present in `sub_modules` to `layers`, or is it better to keep them here along with `NBeats` only, as other models might not use them?

Sohaib-Ahmed21:
> I think we can also move some of the layers (like `KAN` layer) to the `layers` right? I think these layers can be reused by other models as well?

Yes, in the future,  `KANLayer` might be reused...I think then [`GridUpdateCallback`](https://github.com/Sohaib-Ahmed21/pytorch-forecasting/blob/5a31a583ab8108d44843e83093e852d81511e516/pytorch_forecasting/models/nbeats/_grid_callback.py#L4) should also be moved to `layers`. Thoughts?

phoeenniixx:
> > I think we can also move some of the layers (like `KAN` layer) to the `layers` right? I think these layers can be reused by other models as well?
> 
> Yes, in the future, `KANLayer` might be reused...I think then [`GridUpdateCallback`](https://github.com/Sohaib-Ahmed21/pytorch-forecasting/blob/5a31a583ab8108d44843e83093e852d81511e516/pytorch_forecasting/models/nbeats/_grid_callback.py#L4) should also be moved to `layers`. Thoughts?

Yes this can also be added!
I think we need to decide which parts of this PR are reusable and can be moved to `layers`

But this is a `callback` and not essentially a neural layer, so I think moving it to `layers` may not be a good idea, rather a similar folder for `callback`s? Not sure, what do you think?

Sohaib-Ahmed21:
> Yes this can also be added! I think we need to decide which parts of this PR are reusable and can be moved to `layers`
> 
> But this is a `callback` and not essentially a neural layer, so I think moving it to `layers` may not be a good idea, rather a similar folder for `callback`s? Not sure, what do you think?

What I was essentially thinking was that `KANLayer` in terms of my experience is not used much across the library...so the current implementation also sounds logical to me. Maybe we can ask from @benHeid also about what to include or the current implementation sounds good?

phoeenniixx:
> What I was essentially thinking was that `KANLayer` in terms of my experience is not used much across the library...so the current implementation also sounds logical to me.

I agree, it is not used anywhere right now, but we may need it in future, then it will be handy to have it in `layers` - A single API point for reusable layers.

As you said, maybe @benHeid, @fkiraly can also provide us a new perspective...

fkiraly:
> Also, a question for @fkiraly: Imo, we could also move networks present in `sub_modules` to `layers`, or is it better to keep them here along with `NBeats` only, as other models might not use them?

I think ultimately all layers should move to `layers`, although moving layers not changed or affected by this PR is out of scope. But I would move new layers, or layers that now get imported by both nbeats variants to there.

fkiraly (changes requested):
@Sohaib-Ahmed21, not sure if you saw my last message - could you kindly let us know if you are still working on this? If so:

* please mofe layers to the `layers` folder
* add a `pkg` for the new KAN based NBEATS so it is automatically collected by the test framework and tested

Sohaib-Ahmed21:
> @Sohaib-Ahmed21, not sure if you saw my last message - could you kindly let us know if you are still working on this? If so:
> 
> * please move layers to the `layers` folder
> * add a `pkg` for the new KAN based NBEATS so it is automatically collected by the test framework and tested

Hi @fkiraly, thanks for the catch-up. I'm still on it, will do by this weekend. Got busy with the job lately.

fkiraly:
> Got busy with the job lately.

Nice, hope that´s going well!

For the test framework, @phoeenniixx has coded it recently so should be able to help (Aryan Saini on discord) if you get stuck

fkiraly:
FYI @Sohaib-Ahmed21, I have upgraded your code to the most recent package/test structure, so the tests should now genuinely run for the `NBeatsKAN` model. Previously, this was prevented because it was missing the (newly introduced) `get_cls` and `_pkg` methods

Sohaib-Ahmed21:
> FYI @Sohaib-Ahmed21, I have upgraded your code to the most recent package/test structure, so the tests should now genuinely run for the `NBeatsKAN` model. Previously, this was prevented because it was missing the (newly introduced) `get_cls` and `_pkg` methods

Thanks @fkiraly , kindly let me know if something else is needed here in this PR.

fkiraly (changes requested):
the tests are running now and this looks like a genuine failure in the new code

Sohaib-Ahmed21:
> the tests are running now and this looks like a genuine failure in the new code

Okay, I'll look into and fix it.

fkiraly:
FYI, I am working on a `check_estimator` utility that should help you with running tests selectively from a python prompt:
https://github.com/sktime/pytorch-forecasting/pull/1954

This is still WiP but might be helpful

fkiraly:
ah, seems to work now! Can you explain why it was failing?

Sohaib-Ahmed21:
> ah, seems to work now! Can you explain why it was failing?

It failed because the test data used before could give `TweedieLoss` invalid targets; now the dataloader ensures valid non-negative targets, so the loss runs without errors.

Sohaib-Ahmed21:
@fkiraly kindly review/merge this pull request, thanks!

pytorch_forecasting/models/nbeats/__init__.py
```diff
@@ -1,17 +1,17 @@
...
-    NBEATSGenericBlock,
-    NBEATSSeasonalBlock,
-    NBEATSTrendBlock,
-)
+from pytorch_forecasting.models.nbeats._nbeatskan import NBeatsKAN
+from pytorch_forecasting.models.nbeats._nbeatskan_pkg import NBeatsKAN_pkg
 
 __all__ = [
     "NBeats",
-    "NBEATSGenericBlock",
```
fkiraly, line 13: to avoid breaking existing code, we need to keep exporting these in the current location, so import them from the networks location.

We should also add a `# todo v2: remove imports` to those, at a major we can break.

pytorch_forecasting/models/nbeats/sub_modules.py
```diff
@@ -1,219 +0,0 @@
...
-        stop = backcast_length + forecast_length - 1
-    lin_space = np.linspace(
-        start / norm, stop / norm, backcast_length + forecast_length, dtype=np.float32
-    )
-    b_ls = lin_space[:backcast_length]
-    f_ls = lin_space[backcast_length:]
-    return b_ls, f_ls
-
-
-class NBEATSBlock(nn.Module):
```
fkiraly, line 38: same comment here - I would also import the blocks here

fkiraly (changes requested):
Looks good to me!

Very small thing, I think we should restore the exports of the layers, to avoid breaking user code that imported these previously public classes.

fkiraly (approved):
Looks good now, great addition!

PR #1755, opened 2025-01-16
[ENH] speculative design for raw data container `DataSet`

fkiraly:
This PR contains a speculative design for a data container `TimeSeries`.

This follows the idea to split the current `TimeSeriesDataSet` in two parts:

* a "raw data set" object, representing a collection of time series or a single time series - contained in this PR as docstring
* a "resampled data set" which is based on a raw data set and additional resampling instructions - not contained in this PR

The idea of a split was discussed in https://github.com/sktime/pytorch-forecasting/issues/1736, but also already occurs in the comments of @jdb78 in the current `TimeSeriesDataSet`.

* Choices for input and output fields are based roughly on the DSIPTS design, and splitting `TimeSeriesDataSet` in two sets of concerns, one raw data specific, one resampling/batching specific.
* names are chosen that they are descriptive and short, making usage typing efficient.

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -2685,25 +2685,52 @@ class TimeSeries(Dataset):
     Sampling returns a dictionary, which always has following str-keyed entries:
 
     * group: tensor of shape (n_groups)
-    * time: tensor of shape (n_timepoints)
-    * time_fut: tensor of shape (n_timepoints_future)
-    * target: tensor of shape (n_timepoints, n_targets)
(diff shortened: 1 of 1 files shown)


fkiraly:
Some self-review:

* the output format seems a bit complex - can we reduce this to less tensors and still keep handling of categorical and future data?
* where exactly is the separation into past and future data needed? Do we already need this in the raw data layer, or can we keep it as metadata? Do we need to allow for `data_future` to be passed?

fkiraly:
@agobbifbk, @phoeenniixx, @PranavBhatP, we can close this, right?

PR #1757, opened 2025-01-22
[ENH] speculative design for raw data container `DataSet` v2

fkiraly:
This PR contains a speculative design for a data container `TimeSeries`.

This follows the idea to split the current `TimeSeriesDataSet` in two parts:

* a "raw data set" object, representing a collection of time series or a single time series - contained in this PR as docstring
* a "resampled data set" which is based on a raw data set and additional resampling instructions - not contained in this PR

The idea of a split was discussed in https://github.com/sktime/pytorch-forecasting/issues/1736, but also already occurs in the comments of @jdb78 in the current `TimeSeriesDataSet`.

Iteration on https://github.com/sktime/pytorch-forecasting/pull/1755 upon feedback of @thawn that the output should be simpler, as few tensors as possible.

This design returns a near-minimal set, making the choice to:

* separate endogenous from exogenous
* separate future from past data
* otherwise store metadata as list of str or int

This is near-minimal, as at least two tensors are needed, and this results in three. Two tensors are needed since the data is the union of two rectangular array, but not itself a rectangular array (because some features are not seen in the future). The "future features" are optimal though.

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -2685,25 +2685,52 @@ class TimeSeries(Dataset):
     Sampling returns a dictionary, which always has following str-keyed entries:
 
     * group: tensor of shape (n_groups)
-    * time: tensor of shape (n_timepoints)
-    * time_fut: tensor of shape (n_timepoints_future)
-    * target: tensor of shape (n_timepoints, n_targets)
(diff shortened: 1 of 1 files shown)


fkiraly:
@agobbifbk, @phoeenniixx, @PranavBhatP, we can close this, right?

PR #1763, opened 2025-02-05
[DOC] minor fixes in documentation

fkiraly:
Some minor fixes in documentation


yarnabrina (approved):
Approved for sktime typos, but what is happening with all these GHA warnings about numpy v2?

fkiraly:
Which warnings? I cannot see any, so link or explanation how to see them would be appreciated, @yarnabrina.

Issue #1752, opened 2025-01-13
[BUG] TimeSeriesDataSet.to_dataloader batch_size, RuntimeError


gbilleyPeco:
**Describe the bug**

**To Reproduce**

```python
import warnings

warnings.filterwarnings("ignore")

import lightning.pytorch as pl
from lightning.pytorch.callbacks import EarlyStopping
import matplotlib.pyplot as plt
import pandas as pd
import torch

from pytorch_forecasting import Baseline, DeepAR, TimeSeriesDataSet
from pytorch_forecasting.data import NaNLabelEncoder
from pytorch_forecasting.data.examples import generate_ar_data
from pytorch_forecasting.metrics import MAE, SMAPE, MultivariateNormalDistributionLoss

data = generate_ar_data(seasonality=12.0, timesteps=48, n_series=100, seed=42)
data["static"] = 2
data["date"] = pd.Timestamp("2020-01-01") + pd.to_timedelta(data.time_idx, "D")
data = data.astype(dict(series=str))

# create dataset and dataloaders
max_encoder_length = 24
max_prediction_length = 12

training_cutoff = data["time_idx"].max() - max_prediction_length

context_length = max_encoder_length
prediction_length = max_prediction_length

training = TimeSeriesDataSet(
    data[lambda x: x.time_idx <= training_cutoff],
    time_idx="time_idx",
    target="value",
    categorical_encoders={"series": NaNLabelEncoder().fit(data.series)},
    group_ids=["series"],
    static_categoricals=[
        "series"
    ],  # as we plan to forecast correlations, it is important to use series characteristics (e.g. a series identifier)
    time_varying_unknown_reals=["value"],
    max_encoder_length=context_length,
    min_encoder_length=1,
    max_prediction_length=prediction_length,
    min_prediction_length=1,
)

validation = TimeSeriesDataSet.from_dataset(training, data, min_prediction_idx=training_cutoff+1)
batch_size = 64
# synchronize samples in each batch over time - only necessary for DeepVAR, not for DeepAR
train_dataloader = training.to_dataloader(
    train=True, batch_size=batch_size, num_workers=0, batch_sampler="synchronized"
)
val_dataloader = validation.to_dataloader(
    train=False, batch_size=batch_size, num_workers=0, batch_sampler="synchronized"
)

baseline_predictions = Baseline().predict(val_dataloader, trainer_kwargs=dict(accelerator="cpu"), return_y=True)
[...]

PR #1786, opened 2025-02-26
[BUG] pytorch-forecasting#1752 Fixing

RUPESH-KUMAR01:
### Description

This PR is towards the issue #1752 . concat_sequences concat the timesteps of each batch. But our goal is to not concat time_stamps but to concat the batches. 
### Checklist

- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/data/timeseries.py
+++ b/pytorch_forecasting/data/timeseries.py
@@ -2513,7 +2513,7 @@ class TimeSeriesDataSet(Dataset):
 
         Parameters
         ----------
-        train : bool, optional, default=Trze
+        train : bool, optional, default=True
             whether dataloader is used for training (True) or prediction (False).
             Will shuffle and drop last batch if True. Defaults to True.
         batch_size : int, optional, default=64
--- a/pytorch_forecasting/models/base/_base_model.py
+++ b/pytorch_forecasting/models/base/_base_model.py
@@ -387,7 +387,7 @@ class PredictCallback(BasePredictionWriter):
             if self.return_decoder_lengths:
                 output["decoder_lengths"] = torch.cat(self._decode_lengths, dim=0)
             if self.return_y:
-                y = concat_sequences([yi[0] for yi in self._y])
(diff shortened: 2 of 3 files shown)


fkiraly (changes requested):
Great - could we add a test to ensure we have fixed the bug?

RUPESH-KUMAR01:
## **Modifications in `test_temporal_fusion_transformer.py`**  

### **Changes Made:**  
1. **Changed `fast_dev_run` to `2`**  
   - Previously, only **one batch** was run, which might have hidden errors.  
   - Now, **two batches** will run to ensure better testing.  

2. **Added an assertion to check `y` shape after concatenation**  
   - Ensures that **predictions match target dimensions** after processing.  

### **Updated Code Snippet:**  
```python
predictions = net.predict(
    val_dataloader,
    return_index=True,
    return_x=True,
    return_y=True,
    fast_dev_run=2,  # 🔹 Now runs for two batches
    trainer_kwargs=trainer_kwargs,
)

if isinstance(predictions.output, torch.Tensor):
    assert predictions.output.shape == predictions.y[0].shape, "shape of predictions should match shape of targets"
else:
    for i in range(len(predictions.output)):
        assert predictions.output[i].shape == predictions.y[0][i].shape, "shape of predictions should match shape of targets"
```

I am not familiar with tests, but while debugging, I found where the tests failed with my previous approach and modified the function with extra constraints.
Location of the file where I modified:
tests\test_models\test_temporal_fusion_transformer.py
function: _integration

If the changes are sufficient I will add the modifications to the PR.

fkiraly (changes requested):
Nice, thanks for fixing!

Some questions:

It looks like you are changing the tests, not just the code in `_utils`.

Can you explain that?

Would the original test fail?

If not, we should not change existing tests, but add new ones for the bug.

Hspix:
The fix is pending？

RUPESH-KUMAR01:
I tested the code initially before making changes in tests. The tests I modified are to add more constraints to the initial tests. The test modifications will ensure that the batch_size and time_stamps of both output and y match.

PR #1811, opened 2025-04-06
[ENH] EXPERIMENTAL PR: D1 and D2 layer for v2 refactor

phoeenniixx:
### Description

This PR implements the basic skeleton of D1 and D2 layer for v2.

See https://github.com/sktime/pytorch-forecasting/issues/1736 for discussion and design.


phoeenniixx:
FYI @fkiraly @agobbifbk, I have removed the imputation as I just found one thing related to handling missing data:
`constant_fill_strategy` that too handles only the missing timesteps and doesnot handle nans, 
Also see: https://pytorch-forecasting.readthedocs.io/en/latest/faq.html#creating-datasets

Here it says, the nans should be handled before-hand

phoeenniixx:
Thanks for the comments @agobbifbk, I will reply all of them shortly, here are some replies:
> I thought we decide to remove the nan fills in the getitem, or did I miss something?

I have removed the imputation from D2 layer as discussed, the only nan fills are in D1  layer `__getitem__` which is imo required when we concatenate the past and future data as we are creating a new datapoint `x_merged` in this case, and if there is any gap between the past and future data, it needs to be addressed to change them to tensors, should I remove that as well?

> target_scale is something for back-compatibility? Which is the purpose of this key?

yes the `__init__` of datamodule for now contains a the params copied from the previous verison of `TimeSeriesDataset` so that we can have an idea what things we need to add and finally reach the back-compatibility.

> Why the _ProcessedEncoderDecoderDataset expects a EncoderDecoderTimeSeriesDataModule type?

Because we agreed on calling the `_preprocess_data` inside the `__getitem__` and not outside this class, so to access this we need `EncoderDecoderTimeSeriesDataModule`, we cannot use `super()` here, as that is possible only if we are inheriting the class, but here `_ProcessedEncoderDecoderDataset` is a nested class, and not inheriting it, is there any other way without passing `EncoderDecoderTimeSeriesDataModule`, not a python expert here, so I may have missed something. Should I make the `_preprocess_data` standalone function outside the class, in that way we need not to pass  `EncoderDecoderTimeSeriesDataModule` to the nested class, but i am not sure we should do that.

agobbifbk:
> I have removed the imputation from D2 layer as discussed, the only nan fills are in D1 layer `__getitem__` which is imo required when we concatenate the past and future data as we are creating a new datapoint `x_merged` in this case, and if there is any gap between the past and future data, it needs to be addressed to change them to tensors, should I remove that as well? Maybe the solution can be the is_valid check like you suggested?

In my opinion imputing is something not correlated with the D1 layer, or, if you want, you can set a flag (default False or put a warning when it is triggered) and just skip the samples that contains NaNs

> Because we agreed on calling the `_preprocess_data` inside the `__getitem__` and not outside this class, so to access this we need `EncoderDecoderTimeSeriesDataModule`, we cannot use `super()` here, as that is possible only if we are inheriting the class, but here `_ProcessedEncoderDecoderDataset` is a nested class, and not inheriting it, is there any other way without passing `EncoderDecoderTimeSeriesDataModule`, not a python expert here, so I may have missed something. Should I make the `_preprocess_data` standalone function outside the class, in that way we need not to pass `EncoderDecoderTimeSeriesDataModule` to the nested class, but i am not sure we should do that.

I see, but in this way you are copying 3 times the same object, where only the windows is changed. In my opinion with minimal effort you can use the same get_item for the three DataLoaders without duplicating possibly large dataframes. 

Happy to continue the discussion also during the technical meeting, if needed we can arrange one in the next days!

phoeenniixx:
> Happy to continue the discussion also during the technical meeting, if needed we can arrange one in the next days!

Yes, please! IF possible, we can meet atleast once this week before my exams starts, because I think I need a discussion to understand all your points, and I will be able to learn new aspects which can be helpful for me to carry on the work.

fkiraly (changes requested):
My main comment: I think we need to start adding tests now, otherwise the devlopment work is getting too complex.

If you have time before the exams, that is what I would focus on - add tests in all three PR, and ensure they are stacked (keep them updated).

phoeenniixx:
Hi @fkiraly, @agobbifbk  , I have added the tests for D1, D2 layer (still to make changes suggested by @agobbifbk ), but while working on the tests, I realized that the `len` we are returning of `TimeSeries` is more problematic that I thought, for now I am returning 1 group per `__getitem__`  and this even leads to 0 len of  `val_dataset` as there might not be enough `groups`, so I think we should add windows to d1 layer as well, but that leads to another question:
Should 1 window should contain only 1 group or can it contain any number of groups? What I mean is like now I am masking and getting the group, should the windows be added after this? or should I create the windows without masking and directly to the raw data?

Also is there any other way (better way) than using windows here? 

(for now I have commented out the tests checking this train-val splits)

agobbifbk:
Hello @phoeenniixx ! I think that in the suggested changes you can find the answer to this question. 
The train-validation-test split in time series forecasting usually is performed temporal-wise. If we have just one group I think the most common strategy is to use the first x% of the data as training set, the consecutive y% as validation and the rest as test set. The possibility to split by group is something useful (global forecaster for example) but in a limited range of application. The temporal split is more important (you can do it using the percentage as first implementation, adding also temporal  ranges in the future). Hope this can help!

fkiraly:
> I think that in the suggested changes you can find the answer to this question.

Where are these suggested changes? I cannot see these on this PR.

fkiraly (approved):
I have made small structural changes, enabling a merge where using the class will raise a clear warning message to users, and where we can develop on top of this.

* moved the new class `TimeSeries` to a new private module
* added a clear warning message in the new class
* moved the current `TimeSeriesDataSet` class to a private module
* neither of the above impact import locations

The idea is to gradually reduce the "babylonian" stack of PR we currently have, and chip away at complexity by merging gradually (while warning users clearly of experimental nature)

phoeenniixx:
> Hello @phoeenniixx ! I think that in the suggested changes you can find the answer to this question. The train-validation-test split in time series forecasting usually is performed temporal-wise. If we have just one group I think the most common strategy is to use the first x% of the data as training set, the consecutive y% as validation and the rest as test set. The possibility to split by group is something useful (global forecaster for example) but in a limited range of application. The temporal split is more important (you can do it using the percentage as first implementation, adding also temporal ranges in the future). Hope this can help!

Thanks for the clarification, I  understand this, but I am a little unsure how to have this split? I mean should I sort the data first and then make the split (in this way all data points from one group will be together) or just split it regardless of the group and just split it without sorting?

fkiraly:
@agobbifbk, @phoeenniixx, I think we need to aim for feature parity first, i.e., new architecture and exactly the same parameters as before.

Are the requests by @agobbifbk related to adding new parameters that were not here before, or is the statement that we are not supporting sth that was previously supported?

If the latter, can you give a detail explanation which param we are missing?

If the former, I think we should do it separately after the refactor, in a new PR.

phoeenniixx:
> Are the requests by @agobbifbk related to adding new parameters that were not here before, or is the statement that we are not supporting sth that was previously supported?

He has requested some changes to params, but as you suggested we are not adding those params for now, but there are some changes with the logic as well he suggested, most of which can be met in the future iterations, but I think we should at least handle the splitting for now as returning a whole group in one `__getitem__` is not a good idea imo. what do you think?
> If the latter, can you give a detail explanation which param we are missing?

For now, I have tried to keep the params from the `__init__` of the original `TimeSeriesDataset` in our version (although they are divided, some are present in `TimeSeries` and some in the datamodule, depending on the relevance). But the params in datamodule is more of a placeholder for now, as we are not using them as I think the normalisation etc can be added afterwards once we narrow down on the final design.

fkiraly:
> what do you think?

Could you kindly be precise and list or link the changes requested? I do not see them above, but maybe I am missing them.

> But the params in datamodule are more of a placeholders for now, as we are not using them as I think the normalisation etc can be added afterwards once we narrow down on the final design.

So the API is final but some parameters are there and do nothing?
If so, it would be important to add this to the docs to warn users and developers.

phoeenniixx:
> Could you kindly be precise and list or link the changes requested?

@agobbifbk requested these changes: https://github.com/sktime/pytorch-forecasting/pull/1811#issuecomment-2801830248
while adding tests, I faced an issue with the current `__getitem__` return: https://github.com/sktime/pytorch-forecasting/pull/1811#issuecomment-2816727092

As discussed, we can make the changes suggested by @agobbifbk, in future iterations, but the issue with `__getitem__` return should be addressed soon ig.

Regarding the params not being used:
there are params like `target_normalizer`, `scaler` etc that are just new features that can be added to the data_module (these are also marked as TODO in the comments) and can be added in future once we have a stable API.

agobbifbk:
> > Hello @phoeenniixx ! I think that in the suggested changes you can find the answer to this question. The train-validation-test split in time series forecasting usually is performed temporal-wise. If we have just one group I think the most common strategy is to use the first x% of the data as training set, the consecutive y% as validation and the rest as test set. The possibility to split by group is something useful (global forecaster for example) but in a limited range of application. The temporal split is more important (you can do it using the percentage as first implementation, adding also temporal ranges in the future). Hope this can help!
> 
> Thanks for the clarification, I understand this, but I am a little unsure how to have this split? I mean should I sort the data first based on groups and then make the split (in this way all data points from same group will be together) or just split it regardless of the group and just split it without sorting?

Sorting data is something that you want to do in any case otherwise the windowing process can produce inconsistent data. I think we need to have a technical discussion on this point: in my opinion the temporal split is more important respect to the group split. In case of percentage, the most natural way to proceed in this first iterations is to apply the split independently for each group: if groups 1 has 1000 timestamp (and here we need to be sure that these points are consecutive) and the split is 50-20-30 I expect to use the first 500 for the training. If group 2 has 2000 timestamp I expect to use its first 1000 points in the training. Someone can complain that in this way the model will focus more on the group 2, but we can adjust it in a second iteration, using a weight in the sampler. With Sandeep we are in a good spot, maybe we can compare the implementations and see which are the best practices implemented in both the implementation and try to finalize this.

fkiraly:
@agobbifbk, @phoeenniixx, I think we need to have a round of prioritization then, when are you available both? We should also invite @PranavBhatP.

My opinion:

* features currently *not* in the (old) `TimeSeriesDataSet` must be postponed, for software engineering reasons - reach feature parity in a first tranche, add features later. Else, the complexity of the PR becomes too large to handle (for us as a group).
* features currently in the (old) `TimeSeriesDataSet` should be included, except where we can "phase" them in a modular fashion.
* top priority should be give to features required for an end-to-end prototype that cannot be subtracted in a modular way without compromising the end-to-end prototype substantially.

Now which features there are and where they go is currently unclear - so we need to make a list, ensure we have a very concrete understanding of the requirements for each, and then sort them into buckets.

What do you think?

fkiraly:
13 UTC? We have a big agenda item already - planning for ESoC and new mentee cohort - but we can have this as the second big item?

fkiraly:
Another useful piece of info: today, I attended a deep dive into `torch-spatiotemporal` by @mtorabirad.

Interestingly, they have a similar design, where "D2" is a lightning data module, and transformations happen there. I think that validates the decision to put the transformations in D2.

Their "D1" also does some temporal resampling, though I am not sure whether that is a good idea.

An eye on that would be appreciated, to inform our design.

fkiraly (dismissed):
Based on planning discussions on May 12: we will merge this PR to unlock dependencies (e.g., https://github.com/sktime/pytorch-forecasting/pull/1830 of @PranavBhatP) and further development - acknowledging that there is still work to do before the release:

* quick end-to-end prototype and jupyter notebook with usage
* adding a robust test suite
* implementation of "placeholder features" that are not there yet, e.g., scalers/transformers
* managing splitting (global vs temporal) as highlighted by @agobbifbk
* internal refactor to make the code less monolithic, e.g., moving out utility functions and helper classes in separate modules

I made some minimal changes:

* made both dataset and dataloader dataclass-like, by ensuring the `__init__` args are never mutated
* simplified `TimeSeries.__getitem__`
* added "beta" warning in new classes

PR #1812, opened 2025-04-06
Refactor model

phoeenniixx:
### Description

This PR implements the tft model based on the new data pipeline. This PR is branched off from #1811

### DIFF
--- a/pytorch_forecasting/data/data_module.py
+++ b/pytorch_forecasting/data/data_module.py
@@ -1,0 +1,633 @@
+#######################################################################################
+# Disclaimer: This data-module is still work in progress and experimental, please
+# use with care. This data-module is a basic skeleton of how the data-handling pipeline
+# may look like in the future.
+# This is D2 layer that will handle the preprocessing and data loaders.
+# For now, this pipeline handles the simplest situation: The whole data can be loaded
+# into the memory.
+#######################################################################################
+
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from lightning.pytorch import LightningDataModule
+from sklearn.preprocessing import RobustScaler, StandardScaler
+import torch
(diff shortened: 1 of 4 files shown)


phoeenniixx:
Hi @fkiraly, a doubt related to github: as my base of this stack is merged, will deleting that branch will cause problems with this branch as well? I dont think so, but still want to confirm

fkiraly:
> will deleting that branch will cause problems with this branch as well? I dont think so, but still want to confirm

No, the commits are still there on this branch. You just need to merge `main` into this branch asap, to avoid future issues with the merge to `main`, of this branch (I see you have already done this so all is fine!)

fkiraly:
See here the developer guide for chains/stacks of PR: https://www.sktime.net/en/latest/developer_guide/git_workflow.html#working-on-a-chain-of-dependent-tasks

fkiraly:
... it does not say what to do when you merge though. Should we add a minimal update that explains it in a subsection? Just merge A to `main`, and then merge `main` into B and propagate through the chain.

phoeenniixx:
> ... it does not say what to do when you merge though. Should we add a minimal update that explains it in a subsection? Just merge A to `main`, and then merge `main` into B and propagate through the chain.

Yes maybe this will help some confused souls like mine :)

fkiraly (changes requested):
Great!

May I just ask for minor changes:

* in the base model, include the warning message like in the data modules. Make a reference to `self` to also give hte user the name of the model
* make the new modules private
* I would suggest to use `v2` as an abbreviation, i.e., `_base_model_v2` and `_tft_v2`.

fkiraly (approved):

phoeenniixx:
Hi @fkiraly, I think this PR is ready to merge

Issue #1823, opened 2025-04-28
[BUG] utils.concat_sequences concatenation should be on dim=0 ?


cngmid:
**Describe the bug**
The utility function `concat_sequences` should concatenate a list of tensors on its sample dimension. With tensors *"of which first index are samples and second are timesteps"*, the concatenation should be on `dim=0`. 
The current code has `dim=1`.

This utility function is only used in the `BaseModel.predict()` method and creates the following issues.

**To Reproduce**
```
from pytorch_forecasting.data.examples import generate_ar_data
from pytorch_forecasting import Baseline, TimeSeriesDataSet
from pytorch_forecasting.data import NaNLabelEncoder

data = generate_ar_data(seasonality=10.0, timesteps=100, n_series=2, seed=42)
data["static"] = 2
data["date"] = pd.Timestamp("2020-01-01") + pd.to_timedelta(data.time_idx, "D")
data.head()

# create dataset and dataloaders
max_encoder_length = 20
max_prediction_length = 5

training_cutoff = data["time_idx"].max() - max_prediction_length

context_length = max_encoder_length
prediction_length = max_prediction_length

training = TimeSeriesDataSet(
    data[lambda x: x.time_idx <= training_cutoff],
    time_idx="time_idx",
    target="value",
    categorical_encoders={"series": NaNLabelEncoder().fit(data.series)},
    group_ids=["series"],
    # only unknown variable is "value" - and N-Beats can also not take any additional variables
    time_varying_unknown_reals=["value"],
    max_encoder_length=context_length,
    max_prediction_length=prediction_length,
)

batch_size = 71
train_dataloader = training.to_dataloader(
    train=True, batch_size=batch_size, num_workers=0)

baseline_model = Baseline()
predictions = baseline_model.predict(
    train_dataloader, return_x=True, return_y=True,
    trainer_kwargs=dict(logger=None, accelerator="cpu")
)
predictions.output.size(), predictions.y[0].size()
```
Output:
```
(torch.Size([142, 5]), torch.Size([71, 10]))
```

**Expected behavior**
The predicted output and the target should have the same size:
```
(torch.Size([142, 5]), torch.Size([142, 5]))
```

In fact if `batch_size` is set to 70, so that the last batch has a smaller size, the code above returns the error:
```
File C:\ProgramData\miniconda3\envs\sq3\lib\site-packages\pytorch_forecasting\utils\_utils.py:275, in concat_sequences(sequences)
[...]

PR #1827, opened 2025-05-08
[BUG] fix incorrect concatenation dimension in `concat_sequences`

cngmid:
### Description

This PR suggest a fix [1823](https://github.com/sktime/pytorch-forecasting/issues/1823)

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,7 +122,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,7 +122,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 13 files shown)


cngmid:
After my first commit, there was one test failure of the following test with python 3.13
`tests/test_models/test_temporal_fusion_transformer.py::test_prediction_with_dataloder_raw`

I removed `python-version: 3.13` from `.github/workflows/test.yml` and excluded a numpy warning in `pytest.ini`. After the new commit, a new failure arise on the Docs build.

Hopefully the reviewers will fix that.

fkiraly (changes requested):
Thanks!

Could you revert your changes to the `test.yml`?  There is only one combination failing, there is no need to remove all of 3.13 tests. I will take care of that separately, just make sure you do not change the file.

Regarding the skip in the `pytest.ini`, can you explain why we are skipping that?

cngmid:
"Regarding the skip in the pytest.ini, can you explain why we are skipping that?"
It was just to reduce the size of the test log. You are right. We cannot ignore it. I put it back.

fkiraly:
ok, but what exactly are we skipping? Why did you suggest to do that in the first place?

cngmid:
Sorry for any confusion. Let me clarify two points:

1.  **Regarding numpy 2.0 warnings:** I am no longer skipping these, nor am I suggesting they should be skipped. My previous temporary measure to skip them was purely for convenience.
2.  **Regarding current proposals:** The only change I am proposing with respect to the main branch is a one-line modification in the file `pytorch_forecasting\utils\_utils.py` (specifically at line 275).

Hope this clears things up.

fkiraly (approved):
Yes, thanks.

What would be great is if you could add a test as well - if you are not sure how to add tests, simply paste code that fails prior and now passes.

(to add this as a test, add it as a function `test_something` inside a module starting with "test")

fkiraly (changes requested):
How about the code in https://github.com/sktime/pytorch-forecasting/issues/1823?

This is already a great test case:

* add the original code with an assert for the correct shape
* add the modification that fails, as well. Simply paste it, and the test passes if the code does not fail.

fkiraly (approved):
Thanks!

Issue #1829, opened 2025-05-12
[ENH] Refactor the `data_module` layer in `ptf-v2`


phoeenniixx:
**Is your feature request related to a problem? Please describe.**
For now the `EncoderDecoderTimeSeriesDataModule` code is a little messy, that contains a nested `dataset` class (`_ProcessedEncoderDecoderDataset`) and preprocessing and splitting functions. It makes the code a little hard to understand and maintain as we move to more complex implementations

**Describe the solution you'd like**
We should make it more modular:
- Move the nested `_ProcessedEncoderDecoderDataset` class out of the `data_module` class.
- Create separate modules for preprocessing and splitting etc
- Call references of these implementations in the `data_module` class

This will make the code a little more clean and modular and easy to maintain.

**Additional context**
Please look at how the `tsl` handles these splitting and preprocessing using different modules

phoeenniixx:
FYI @agobbifbk, @fkiraly , @PranavBhatP

PranavBhatP:
Hi, @phoeenniixx, after going through the code for the `EncoderDecoderDataModule`, I did notice that many parts of this module are reusable and can be adapted for usage by other implementations of data modules for v2.0, so I think it's a necessary step to do this early on before we introduce a lot duplication in code.

fkiraly:
> I did notice that many parts of this module are reusable and can be adapted for usage by other implementations of data modules for v2.0, so I think it's a necessary step to do this early on before we introduce a lot duplication in code.

Interesting. I think it could be instructive if you try to draft this in a hacky way, but then we do not merge this but take this as a study motivating a clean refactor.

fkiraly:
Another point I noticed why finalizing https://github.com/sktime/pytorch-forecasting/pull/1811 is that you the loaders and data modules are not dataclass-like, i.e., some `__init__` arguments get mutated through coercions or defaults.

It is general best practice to make specification-like classes also dataclass-like, so I tried to move the processed attributes to private versions, but this causes test failures. So instead, I did that and then I added overwrites in "todo" sections in the `__init__` so the code does not break.

My question would be, how much problems does it cause if we try to make the modules and loaders dataclass-like? Is it just a matter of changing tests, or would we be changing a public contract that you had in mind, @phoeenniixx?

fkiraly:
> yes, ig all we need to do is just make some changes to tests and code and it will become dataclass-like :)

[...]

PR #1832, opened 2025-05-13
[ENH] EXPERIMENTAL PR: make the `data_module` dataclass-like

phoeenniixx:
This PR makes the `data_modulel` dataclass-like
See discussion in #1829

### DIFF
--- a/pytorch_forecasting/data/data_module.py
+++ b/pytorch_forecasting/data/data_module.py
@@ -165,11 +165,11 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
 
         # overwrite __init__ params for upwards compatibility with AS PRs
         # todo: should we avoid this and ensure classes are dataclass-like?
-        self.min_prediction_length = self._min_prediction_length
-        self.min_encoder_length = self._min_encoder_length
-        self.categorical_encoders = self._categorical_encoders
-        self.scalers = self._scalers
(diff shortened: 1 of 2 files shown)


phoeenniixx:
Hi @fkiraly, I tried to change the tests, is this what is intended? or should I need to make some more changes?

fkiraly:
> Hi @fkiraly, I tried to change the tests, is this what is intended? or should I need to make some more changes?

Yes, this may be sufficient - but are you sure we do not use the attributes in https://github.com/sktime/pytorch-forecasting/pull/1812?

phoeenniixx:
> > Hi @fkiraly, I tried to change the tests, is this what is intended? or should I need to make some more changes?
> 
> Yes, this may be sufficient - but are you sure we do not use the attributes in #1812?

we can merge this branch to #1812, then i think this should work?, just few adaptations may be required and most importantly the attributes are used with the help of `metadata` there, which use the private variables only, so ig this is safe

phoeenniixx:
So should we merge this @fkiraly ?

fkiraly:
> So should we merge this @fkiraly ?

If this should be merged, you should turn it to a "ready for review" pull request, it is still a draft

phoeenniixx:
yes, sorry didn't notice it was draft, i have made the final changes, please review

fkiraly (approved):

PR #1834, opened 2025-05-14
[BUG] EXPERIMENTAL PR: Solve the bug in `data_module`

phoeenniixx:
This PR solves the bug in `data_module` where the `static_categorical_features` and `static_continuous_features` were not correctly calculated in `__getitem__` of nested class

### DIFF
--- a/pytorch_forecasting/data/data_module.py
+++ b/pytorch_forecasting/data/data_module.py
@@ -519,8 +519,38 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
                 "decoder_mask": decoder_mask,
             }
             if data["static"] is not None:
-                x["static_categorical_features"] = data["static"].unsqueeze(0)
-                x["static_continuous_features"] = torch.zeros((1, 0))
+                st_cat_values_for_item = []
+                st_cont_values_for_item = []
+                raw_st_tensor = data.get("static")
+
+                static_col_names = self.data_module.time_series_metadata["cols"]["st"]
+                for i, col_name in enumerate(static_col_names):
+                    feature_value = raw_st_tensor[i]
+                    if (
+                        self.data_module.time_series_metadata["col_type"].get(col_name)
+                        == "C"
+                    ):
+                        st_cat_values_for_item.append(feature_value)
+                    else:
+                        st_cont_values_for_item.append(feature_value)
+
+                if st_cat_values_for_item:
+                    x["static_categorical_features"] = torch.stack(
+                        st_cat_values_for_item
+                    ).unsqueeze(0)
+                else:
+                    x["static_categorical_features"] = torch.zeros(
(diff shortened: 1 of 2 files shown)


phoeenniixx:
This solves the bug that was found while writing the integration tests for `tft` model in #1812

fkiraly (approved):
It fixes the bug, so it is approved.

But, the new code seems unnecessarily complicated?
Is it not possible to avoid the loops by simply using array operations smartly?

phoeenniixx:
Sure I'll try to find a better way

fkiraly:
the "basic principle" is to replace sth like 

```python
my_list = [] * len(my_vector)
for i, x in enumerate(my_vector):
    my_list[i] = 42*x
```

by something like

```python
my_list = list(42*my_vector)
```

phoeenniixx:
> the "basic principle" is to replace sth like
> 
> ```python
> my_list = [] * len(my_vector)
> for i, x in enumerate(my_vector):
>     my_list[i] = 42*x
> ```
> 
> by something like
> 
> ```python
> my_list = list(42*my_vector)
> ```

so we can use this approach by creating a mask right? then this mask can easily be mutiplied to get the final `st_cat_values_for_item`?
Like 
```python
is_categorical_mask = [
                self.data_module.time_series_metadata["col_type"].get(col_name) == "C"
                for col_name in static_col_names
            ]
is_continuous_mask = [not b for b in is_categorical_mask]
st_cat_values_for_item = raw_st_tensor[is_categorical_mask] 
st_cont_values_for_item = raw_st_tensor[is_continuous_mask]
```

fkiraly:
yes, exactly

fkiraly:
plus, naively, I would expect something like 

```python
is_categorical_mask =  self.data_module.time_series_metadata["col_type"] == "C"
```

to work.

phoeenniixx:
> plus, naively, I would expect something like
> 
> ```python
> is_categorical_mask =  self.data_module.time_series_metadata["col_type"] == "C"
> ```
> 
> to work.

We just need the `static_categorical` in this case, not all categorical
and i am not sure this comparison works with `dict`, I am sure similar way is possible with `pandas`, but with `dict`?

fkiraly (approved):
Yes, I think this is better!

phoeenniixx:
> Hello! I have a comment on the current situation: we are assuming that the data are already sorted and there is no gap in the temporal index. Since the output is taken and windowing in the d2 layer, we need to be sure that all the data is equally spaced. Maybe it is better to check it somehow during the `getitem`, and in a second moment to enlarge the dataset and put nans in the holes. What do you think?

Yes! ig we should do this, but I think we could add it in next iterations? First complete a basic e2e prototype? We have added the warning, so the users know the code still has loop holes.

agobbifbk:
Ok it was just a comment, we need to remember this :-)

Issue #1548, opened 2024-03-21
How does one initialise a network without `from_dataset`?


chrism2671:
### Expected behavior
Code:

```
from pytorch_forecasting.models import NBeats, BaseModel
from pytorch_lightning import Trainer, LightningModule

model = NBeats()
trainer.fit(NBeats, train, valid)

```
Result:

```
TypeError                                 Traceback (most recent call last)
[<ipython-input-90-0010063a6148>](https://localhost:8080/#) in <cell line: 1>()
----> 1 trainer.fit(NBeats, train,valid)

1 frames
[/usr/local/lib/python3.10/dist-packages/pytorch_lightning/utilities/compile.py](https://localhost:8080/#) in _maybe_unwrap_optimized(model)
    130         return model
    131     _check_mixed_imports(model)
--> 132     raise TypeError(
    133         f"`model` must be a `LightningModule` or `torch._dynamo.OptimizedModule`, got `{type(model).__qualname__}`"
    134     )

TypeError: `model` must be a `LightningModule` or `torch._dynamo.OptimizedModule`, got `type`
```

I'm trying to use the naked networks without the rest of the stuff around pytorch_forecasting. 

I've read the source code I do believe this should work; but I must be doing something stupid.

Is it possible to add an example or FAQ of how to use pytorch_forecasting without `from_dataset`?

benHeid:
You are using the old `pytorch_lightning`. I suppose you need to import `lightning.pytorch` instead. 

TODO:
* [ ] check if we can add a documentation on how to use models without the `from_dataset` method and if it works :) 

are you having a good location, where the extension of the documentation would fit? Would you like to contribute it?

MeherBhaskar:
Hi @benHeid ,
I made the changes discussed in this issue and raised a PR here - https://github.com/sktime/pytorch-forecasting/pull/1837
Please take a look when you get a chance and share your thoughts.

PR #1837, opened 2025-05-19
[ENH] Allow direct initialization of `NBeats` model without `from_dataset`

MeherBhaskar:
### Description

This PR fixes issue [#1548 ](https://github.com/sktime/pytorch-forecasting/issues/1548) where the NBeats model fails when initialized without using the `from_dataset` method. The changes enable direct initialization while maintaining backward compatibility and adding proper validation.

## Changes
- Enhanced `__init__` method with comprehensive parameter validation
- Added clear documentation for both initialization methods (direct and from_dataset)
- Added validation for parameter list lengths to match stack types
- Updated `from_dataset` method documentation to clarify its role
- Added unit tests for direct initialization

Now users can:
1. Use `from_dataset` (recommended) for standard time series forecasting
2. Initialize directly for custom use cases
3. Get clear error messages when parameters are invalid

Example of direct initialization:
```python
model = NBeats(
    stack_types=["trend", "seasonality"],
    num_blocks=[3, 3],
    num_block_layers=[3, 3],
    widths=[32, 512],
    sharing=[True, True],
    expansion_coefficient_lengths=[3, 7],
    prediction_length=24,
    context_length=72,
)
```
### Checklist

- [x] Added/modified tests

Make sure to have fun coding!


fkiraly (changes requested):
Thanks!

Quick question, why are we changing the old test? Does this mean existing code breaks?

We should avoid to break working code of users.

MeherBhaskar:
Hi @fkiraly,
The change to update the widths from [4, 4, 4] to [4, 4] was made to match the default number of stacks used by the N-BEATS model (["trend", "seasonality"])

Passing a list of length 3 was causing the following error:
```
ValueError: Length of widths (3) must match length of stack_types (2)
```
Let me know if you'd like to handle this differently!

fkiraly:
Can you explain why the `[4, 4, 4]` was previously passing them in the tests? If it worked previously - and seems to have been standard expectation, an dnow we change it, we might break someone's code.

jobs-git:
@MeherBhaskar can I request that we also allow training the model this way:

```python
output = model(x)
```

jobs-git:
> Hi @fkiraly, The change to update the widths from [4, 4, 4] to [4, 4] was made to match the number of stacks used by the N-BEATS model (["trend", "seasonality"])
> 
> Passing a list of length 3 was causing the following error:
> 
> ```
> ValueError: Length of widths (3) must match length of stack_types (2)
> ```
> 
> Let me know if you'd like to handle this differently!

I encountered a similar issue in nhits too! Apparently, even if you just have two hidden layers, its asking hidden_size as 3 elements! What happens was that, there is a weird implementation in the NHiTS block for MLP, it adds the input_feature which are features of the input layer to the hidden size! This was forcibly needed to be able to fix the input_feature of the first hidden layer as supplied by the input layer. That is why I have this pull request to avoid this confusion: https://github.com/sktime/pytorch-forecasting/pull/1864

Note that, i also saw this same style from the reference code of the paper authors, so it was nobody's fault. NBeaTS might also have this same confusion?

Ref: Check out the `n_theta_hidden` here: https://github.com/cchallu/n-hits/blob/main/src/models/nhits/nhits.py

fkiraly:
@phoeenniixx, @PranavBhatP, review against the v1 interface would be appreciated!

Issue #1838, opened 2025-05-20
[ENH] Testing Framework for `ptf-v2`


phoeenniixx:
We should add test framework for `ptf-v2`. The starting point would be #1780, But as it is mainly focused on v1, we might need to refactor some components from it. 
This will also be used to test how the models present in original implementations need to be refactored to adapt them for v2.

**For more context** see discussion in #1780

phoeenniixx:
FYI @agobbifbk, @fkiraly , @PranavBhatP

Issue #1679, opened 2024-09-21
[ENH] tagged objects and retrieval


fkiraly:
Related to this issue: https://github.com/sktime/pytorch-forecasting/issues/1582

Would it be useful to tag objects and make them available for retrieval?

The package contains neural networks and custom layers.

If we think, yes, then the "how" is the next question. Some thoughts:
* `torch` objects do not mutate their init args, so they would be compliant with `skbase`, which brings its own tagging and retrieval system. Though I am not entirely sure whether we have compatibility, or whether we want to "weigh down" the torch objects.
* alternatively, we could add records that are `BaseObjects` that point to the `torch` object. That may be safer, as the torch objects themselves would not inherit non-torch logic, which might be safer for upwards compatibility with `torch`. The idea here is a barebones `BaseObject` with tags, an `__init__` with same params as the object, and with a `build` method that constructs the object instead.

PR #1841, opened 2025-05-20
[ENH] Test framework for `ptf-v2`

phoeenniixx:
### Description

This PR solves #1838
Implements Test framework for v2


phoeenniixx:
Hi @fkiraly, here I branched off from #1812 and also merged your PR #1780, I will need some days to actually get my head straight for this PR

fkiraly (commented):
Here is a thought:

tags and properties will be the same for v1.

Idea: could you complete the v1 framework first? Separate PR, and all it does is adding the models. Given that a lot of code is shared, it might make it easier to write the "integration" part for v2.

phoeenniixx:
> could you complete the v1 framework first? Separate PR, and all it does is adding the models. Given that a lot of code is shared, it might make it easier to write the "integration" part for v2.

I can do that, but this may shift the "cutoff time" for this PR will move forward from this week :) 
but I agree that working on v1 will help me understand the code and lib more deeply and will help in integration for v2 

So I think we should move ahead with this? I'll keep this PR on hold and raise a new one branched off from your PR and get the v1 models first

phoeenniixx:
Also where can I find the whole list of tags and properties with meanings like I dont understand some of them like : `info:compute` and `capability:cold_start`
and I assume there are more tags as well which are not mentioned rn in `DeepARMetadata` ?

fkiraly:
> Also where can I find the whole list of tags and properties with meanings like I dont understand some of them like : `info:compute` and `capability:cold_start`

I got them from here:
https://pytorch-forecasting.readthedocs.io/en/latest/models.html

Although there is not a clearly defined meaning that I could find.

> and I assume there are more tags as well which are not mentioned rn in `DeepARMetadata` ?

Perhaps - currently, there are no tags whatsoever in `pytorch-forecasting` - so we should think about which tags would make sense.

For the start, while we build the framework, I think it is ok to have documentation only in the code what the tags mean - see https://github.com/sktime/pytorch-forecasting/issues/1679

Long-term we should document what they mean, and have them in the docs, like here:
https://www.sktime.net/en/latest/api_reference/tags.html

fkiraly:
> but the original implementation, it is causing some problems

Can we not leave the original implementation as-is for now, and ensure it is not collected in the v2 tests?

After all, the tests are meant for v2 for now.

fkiraly:
>  I tried adding `TFT` to the tests

Quick questions:

* is there a new, second test class for v2 tests?
* which exact model classes are tested by which exact tests?
* which combination of model class and test class fail?

phoeenniixx:
>Can we not leave the original implementation as-is for now, and ensure it is not collected in the v2 tests?

Actually I tried to bring the original implementation to v2, to see how much effort is required, I haven't changed the main implementation in `_tft.py` rather, created a new file `_tft_ver2` which has the same code just some changes that I have mentioned in comments in the file as well

This is how I have done changes:
- `_data_scenarios.py`: created 2 new functions - `data_with_covariates_v2` and `make_dataloaders_v2`, here I am not using exactly stallion-data as for now the data pipeline can handle only the numeric data
- `_conftest.py`: same changes as `_data_scenarios.py`
- `test_all_estimators_v2.py`: contains the `_integration` func for v2 and `TestAllPtForecastersV2`
- `_base_model_v2.py`: changed the `loss` from `nn.Module` to `Metric`
- `_tft_ver2.py` : contains almost the same implementation as original `_tft.py` just some refactors to make it compatible with v2 - here is where the error is causing problems
- `_tft_v2_metadata`: contains 2 metadata classes - `TFTMetadata` for the TFT implementation in `_tft_v2` that is not as robust as original `_tft` and `TemporalFusionTransformerMetadata` that is metadata class for original implementation with some refactors (`_tft_ver2`)

I have not changed any of the existing code, or code on your branch of #1780, just added some code for v2

phoeenniixx:
> * is there a new, second test class for v2 tests?

yes, `TestAllPtForecastersV2` in `test_all_estimators_v2.py`
> * which exact model classes are tested by which exact tests?

only the `_tft_v2` and `_tft_ver2` files are being tested for `_integration` in `test_all_estimators_v2.py`.

> * which combination of model class and test class fail?

the `_tft_ver2` fails as mentioned above

I think I can make it more clear in a tech meet about what exactly the problems I am facing

phoeenniixx:
Hi @fkiraly, i just wanted to make sure that how should we test `data_module`? As the data_module instance is created in `make_dataloaders_v2` and not with the model, so, it may remain same with different models, So should I create a global `get_data_module` that returns the `data_module` and then tests if `metadata` is passed correctly or not?

 And what tests to include?
- Check if `metadata` is passed correctly across layers and to the dataloaders 
- Check if the dataloaders are created correctly, 

other than that do we need anything else?

fkiraly:
> So should I create a global `get_data_module` that returns the `data_module` and then tests if `metadata` is passed correctly or not?

I think it makes sense to have two types of tests:

* the D2 data modules individually, I would for now make loose testes here
* integrated with the model, I would add a `get_data_module` or similar method to the metadata class. Via this it can then be retrieved for an integration test.
* Perhaps a "data module test config" method also needs to be added, if the config varies?

> And what tests to include?

The things you mention make sense - also, simple input/output contract tests.

phoeenniixx:
> the D2 data modules individually, I would for now make loose testes here

Will the available tests ([test_d1](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_data/test_d1.py), [test_data_module](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_data/test_data_module.py) ) be sufficient?

> Perhaps a "data module test config" method also needs to be added, if the config varies?

Yes the config may vary based on the transformers, scalers etc, so we can add this method -  in the metadata class only?

fkiraly:
FYI, I have merged #1780 to prevent conflicts through divergence.

As a kind reminder, to avoid conflicts it make sense to merge the lower PR in a stack regularly:
https://www.sktime.net/en/stable/developer_guide/git_workflow.html#working-on-a-chain-of-dependent-tasks

I am now aware that this may be a WiP fork for prototyping purposes - still some conflicts have accumulated and should probably get resolved, so tests can run.

phoeenniixx:
> I would ask this question in a bigger group. I will have to think about it.

Maybe we can discuss this in one of our tech sessions?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -14,11 +14,23 @@
...
+    QuantileLoss,
+    convert_torchmetric_to_pytorch_forecasting_metric,
+)
+
 
 class BaseModel(LightningModule):
     def __init__(
         self,
-        loss: nn.Module,
+        loss: Metric = SMAPE(),
```
PranavBhatP: Just a minor change request, I think it would be better to define a default value for loss in the actual model which is  inheriting from `BaseModel` and leave the parameter without a default value inside the `BaseModel`, since `BaseModel` should be model-agnostic.

phoeenniixx: Thanks for pointing it out, but I think we should do this in another PR? As this is making changes to the already available code?

PranavBhatP: Yes, I think this would be better to put in as a separate PR. We can stick to nn.Module for now.

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -262,6 +262,21 @@ def _integration(
 class TestAllPtForecasters(PackageConfig, BaseFixtureGenerator):
     """Generic tests for all objects in the mini package."""
 
+    def _all_objects(self):
```
fkiraly: why is this needed? I thought the `_all_objects` filter would work without this.

fkiraly: (only the `object_type_filter` needs to be added)

phoeenniixx: actually I tried that, but `TestAllPtForecasters` collects ALL the models irrespective of the filter, the filter is added in `TestAllPtForecastersV2` and not to  `TestAllPtForecasters`, then how would `TestAllPtForecasters` know that it has to skip `ptf-v2`?

phoeenniixx: can we use `object_type_filter` here as well in some way? It will not accept the `ptf-v2`? I am new to this, so if it is possible please tell me :)

fkiraly: There is already filter logic in `BaseFixtureGenerator._all_objects`, it is supposed to use the class tag and the attribute `object_type_filter` to subset.

If it works as intended, `all_objects` subsets to the right type.

if not, then there is a bug either with the filter logic, or `all_objects`.

phoeenniixx: Yes I found my mistake, thanks!

fkiraly (changes requested):
Nice!

May I suggest to have two private base classes, one for v1 and one for v2? This will make it easier to maintain. Further, note that setting the tag `pft-v1` will overwrite the current object type `forecaster_pytorch`.
May I instead suggest `["forecaster_pytorch", "forecaster_pytorch_v1"]` - this will ensure the respective forecaster has both types.

phoeenniixx:
> May I suggest to have two private base classes, one for v1 and one for v2?

You are talking about `_BasePtForecaster` right? For now `_BasePtForecaster` can handle v2 as well, so if I add a new one for v2, it will have the same code...

fkiraly:
>> May I suggest to have two private base classes, one for v1 and one for v2?

>You are talking about `_BasePtForecaster` right? For now `_BasePtForecaster` can handle v2 as well, so if I add a new one for v2, it will have the same code...

Yes - I suggest to inherit for now instead of using the same class. Make a common base class for both, and the v1 / v2 classes only have the different tag setting for now.

phoeenniixx:
so you are saying I create a new base class from which v1, v2 `_BasePtForecaster` inherit?
I think the only change in both of them will be 
```python
class _BasePtForecaster(_BaseObject):
    """Base class for all PyTorch Forecasting forecaster metadata.

    This class points to model objects and contains metadata as tags.
    """

    _tags = {
        "object_type": "forecaster_pytorch",
    }

```
this `object_type` tag - either `forecaster_pytorch` or `forecaster_pytorch_v2` (I'll change it from `ptf-v2` to make it consistent)

Then I have a question: Should I move the current `_BasePtForecaster` logic to `_BaseObject` and then it `_BasePtForecasterV1` and `_BasePtForecasterV2` will inherit from `_BaseObject`?

phoeenniixx:
The new design will look something like this:
```python
class _BaseObject(_SkbaseBaseObject):
    # Existing `_BasePtForecaster` logic
    pass

class BasePtForecasterV1(_BaseObject):

    _tags = {
        "object_type": "forecaster_pytorch",
    }

class BasePtForecasterV2(_BaseObject):


    _tags = {
    }
```

fkiraly:
I would do it this way, only slightly different:
* one more class to leave "space" for things like metadata for layers
* change the type string of the v1 forecaster

```python
class _BaseObject(_SkbaseBaseObject):
    # as current

class _BasePtForecaster(_BaseObject):
    # Existing `_BasePtForecaster` logic
    pass

class _BasePtForecasterV1(_BasePtForecaster):

    _tags = {
        "object_type": "forecaster_pytorch_v1",
    }

class _BasePtForecasterV2(_BasePtForecaster):

    _tags = {
    }
```

phoeenniixx:
Should I do this in this PR only or raise a new one (as I'll be making changes to V1 as well?)

fkiraly:
> Should I do this in this PR only or raise a new one (as I'll be making changes to V1 as well?)

Strictly speaking, it is as you say: best practice would be to introduce the intermediate class in a separate PR and stack on it, to keep changes localized. You are right!

However, the change is so small that I think it is not a big problem if we do not split it off.

pytorch_forecasting/tests/test_all_estimators_v2.py
```diff
@@ -0,0 +1,145 @@
...
+
+    assert batch_y.shape[2] == metadata["target"]
+
+    net = estimator_cls(
+        metadata=metadata,
+        loss=nn.MSELoss(),
+        **kwargs,
+    )
+
+    try:
```
fkiraly: in a test, is try without except just silent execution of code? If I understand correctly, if this code fails, nothing happens - so why run it in the first place?

fkiraly (changes requested):
Looks good!

One question in there about the try block without except, in the tests.

pytorch_forecasting/tests/test_all_estimators_v2.py
phoeenniixx: Well I took inspiration from v1 `_integration`, so that both versions could remain as similar as possible
See: https://github.com/sktime/pytorch-forecasting/blob/4a34931e499c2b59de3939fcffcaabd75204b045/pytorch_forecasting/tests/test_all_estimators.py#L212-L234

I had to remove `net.predict` as rn v2 implementation doesnot have it (I may add it in further iterations)

fkiraly: I do not have a problem with what is inside the `try` block, but that there is a `try` block. We are suppressing all exceptions as far as I understand, so nothing is tested - the test is just as good as if this block would not exist.

Please let me know if you disagree.

phoeenniixx: That makes sense! I will remove it
Just a follow up question: So is this required in V1? Because as much as I can understand, there as well we donot use `except`, at first I thought it was a design choice as this `try`- `finally` block was present in almost all tests in V1 (before test framework as well), but as you said it can work without this block as well?

fkiraly: > Just a follow up question: So is this required in V1? Because as much as I can understand, there as well we donot use `except`, at first I thought it was a design choice as this `try`- `finally` block was present in almost all tests in V1 (before test framework as well), but as you said it can work without this block as well?

I now realize this was already there in v1!

Now it is a different story in v1, there it was *pre-existing*, I just moved the code in a pure refactor.

Along the same lines though, I doubt the usefulness of this code block in v1 - perhaps we should simply remove it and see what happens?

FYI @jdb78, it would be much appreciated if you could shed light on this - why is there a `try` block in the tests that silences all exceptions? Or am I wrong in believing that it silences all exceptions?

phoeenniixx: > Along the same lines though, I doubt the usefulness of this code block in v1 - perhaps we should simply remove it and see what happens?

Maybe we can try it in a draft PR? 
I can raise one to see if it breaks anything...

pytorch_forecasting/tests/test_all_estimators_v2.py
```diff
@@ -0,0 +1,144 @@
...
+    assert metadata["decoder_cat"] == 0
+    assert metadata["static_categorical_features"] == 0
+    assert (
+        metadata["static_continuous_features"] == 2
+    )  # 2 (agency_encoded, sku_encoded)
+    assert metadata["target"] == 1
+
+    batch_x, batch_y = next(iter(train_dataloader))
+
+    assert batch_x["encoder_cont"].shape[2] == metadata["encoder_cont"]
```
PranavBhatP: I see that the test_all_estimators for v2 is specific to the `EncoderDecoderDataModule`. I might be out of context here, but is it decided as to how will we handle the testing of models with separate dataloaders? I noticed your comment on using a separate `get_data_module` function inside `make_dataloaders_v2` to extract the specific d2 layer for that model. Will this require a separate tag on the model? Would be grateful for updates on this and the approach you are planning to take.

phoeenniixx: As discussed, these changes are meant to happen in #1868
> I noticed your comment on using a separate `get_data_module` function inside `make_dataloaders_v2` to extract the specific d2 layer for that model. Will this require a separate tag on the model?

We'll be using the strategy in #1859, and create `_get_test_datamodules_from` inside the metadata class (see comment: https://github.com/sktime/pytorch-forecasting/pull/1868#issuecomment-2936265136, and the further conversation) and we will not be using tags, that discussion for now is left for the future to determine if we even need tags or not.

PranavBhatP: Thanks for the update! The rest of this PR looks good! No other reviews from my side.

phoeenniixx:
Hi @fkiraly, @PranavBhatP, I tried adding  `_get_test_datamodule_from` to the metadata class, please review it and tell me if I am on a right path

phoeenniixx:
- I added the whole initialisation logic of the datamodule to `_get_test_datamodule_from` so that any kind of data module can be added to tests without making any changes to `_integration`
-  changed the `make_dataloaders_v2` to `make_datasets_v2` as now it is returning `TimeSeries` dataset in place of whole dataloaders and datamodule - now `_get_test_datamodule_from` returns the datamodules and dataloaders.
- now the `_integration` test only checks if `metadata` is a `dict`.

The open issue is: how to do `metadata` testing for different datamodules?
By `metadata` testing I mean testing if the `metadata` is calcualted correctly and passed correctly to dataloaders and model for initialisation.

fkiraly:
> The open issue is: how to do `metadata` testing for different datamodules?
By `metadata` testing I mean testing if the `metadata` is calcualted correctly and passed correctly to dataloaders and model for initialisation.

Hmmm - here is an idea: each model (intermediate base class) could have a `_check_metadata` that is called from the test.
It passes if the metadata is of expected type, and fails otherwise with a descriptive exception.

For performance reasons, I would not carry out the check in usage, just in the test.

For the metadata generated by D1, this is universal and can simply be tested by D1 tests - all D1 should produce the same metadata.

pytorch_forecasting/models/temporal_fusion_transformer/tft_v2_metadata.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @classmethod
+    def get_model_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models.temporal_fusion_transformer._tft_v2 import TFT
+
+        return TFT
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
```
PranavBhatP: I have a suggestion to improve this. Suppose we have 10 different models dependent on a single data module. In this case, to follow DRY principles and to prevent redundancy in the metadata file for each model, we can probably include the code inside this function as a part of a utility function called _build_encoder_decoder_datamodule(trainer_kwargs), `_build_tslib_datamodule` etc..

```python
@classmethod
def get_test_datamodule_from(cls, trainer_kwargs):
   return  _build_encoder_decoder_datamodule(**trainer_kwargs) # return the final dict output.
```
and so on for different types of data modules. I am making this suggestion on the assumption that the code inside this _get_test_datamodule is only dependent on the datamodule. Again, this doesn't affect functionality, it can be refactored when we move to add utility functions to make the code cleaner in several parts of the codebase.

Any thoughts @phoeenniixx @fkiraly ?

phoeenniixx:
> each model (intermediate base class) could have a `_check_metadata` that is called from the test.
It passes if the metadata is of expected type, and fails otherwise with a descriptive exception.

Sorry I didn't quite understand what you mean here.
- If you mean we add `_check_metadata` to the base class `_BasePtForecasterV2` (or its child class) how to add different tests for different `metadata` - Imagine one model is from `EncoderDecoderDataModule` and other from `TSLibDataModule` then, how would `_check_metadata` would know which metadata to test? We'll need to add the testing logic of both the data modules to the same method, but what if there is one more data module in future? I think that will make the method little bit messy?
```python
class _BasePtForecasterV2(_BasePtForecaster):
    def _check_metadata():
          if EncoderDecoderDataModule:
                    return test_enc_dec_dm()
          elif TSLibDataModule:
                     return test_tslib_dm()
          elif:
               ... # so on?

```
- If you mean each MODEL has `_check_metadata` then there will be duplicate code for all models having same datamodule?

These are my initial thoughts, but I still need your help to understand what exactly is the idea, I'm a bit lost here😅

pytorch_forecasting/models/temporal_fusion_transformer/tft_v2_metadata.py
phoeenniixx: yes I think we can do this! Thanks! just there can be the issues like @fkiraly faced in v1 that each model expects a sightly different kind of dataloader, so we might need to add that logic in future but that can be handled using `kwargs` ig

Maybe to keep the test utils separate, we donot add these utilities to `utils` rather to `tests.utils`?

fkiraly:
> If you mean each MODEL has `_check_metadata`

That is what I mean

> then there will be duplicate code for all models having same datamodule?

... unless you add the `check_metadata` to a common intermediate base class that all models using that data module inherit from.
(which I did not say but thought would be implied by this design)

phoeenniixx:
Hi @fkiraly, do we need to add something else to this PR? We can add `metadata` testing in some other PR?

fkiraly:
I think this is ready - the conflicts with #1886 need to be resolved though.

phoeenniixx:
done

fkiraly (approved):
I made some small changes:

* deduplicate code copy-pasted twice in `_conftest` - why did you put the same code in two files?
* revert change to `_BasePtForecaster` name, the name remains the same until we move to v2
* changed "metadata" filename to pkg

PR #1850, opened 2025-05-27
[ENH] Precompute data to accelerate training in GPU

### LINKED ISSUE #715: gpus=1, but works on CPU mostly. Is it normal?
- PyTorch-Forecasting version: 0.9.1
- PyTorch version: 1.9.0a0+c3d40fd
- Python version: 3.8.10
- Operating System: Ubuntu

I've ran tutorial code ('Demand forecasting with the Temporal Fusion Transformer') and used gpus=1, but mostly CPU have been used for training/optimization of hyperparameters. Is it normal? 

Снимок экрана 2021-10-28 в 7 52 30 PM

Снимок экрана 2021-10-28 в 7 52 47 PM

### LINKED ISSUE #922: GPU not being utilised
- PyTorch-Forecasting version: 0.9.2
- PyTorch version: 1.10.0
- Python version: 3.9.7
- Operating System: Ubuntu

### Expected behavior

I tried to use the N-beats tutorial to do some forecasting. I used Paperspace to do the training on a GPU

### Actual behavior

However, When I try to train the model the GPU usage never goes higher then 1%

### Code to reproduce the problem

Screenshot 2022-03-18 at 4 57 22 PM

Screenshot 2022-03-18 at 4 35 27 PM

### LINKED ISSUE #991: Slower trainning on Linux under 0.10.1
## 🐛 Bug

code：https://pytorch-forecasting.readthedocs.io/en/latest/tutorials/stallion.html
When max_epochs=1 and use cpu, it costs **9.6s** on win 10 and **382.79s** on linux.
Using **AdvancedProfiler** from pytorch-lightning, there is difference in **SingleDeviceStrategy.validation_step** between platforms

windows:
```
   Profile stats for: [Strategy]SingleDeviceStrategy.validation_step

   504226 function calls (490331 primitive calls) in 0.638 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.000    0.000    0.638    0.319 strategy.py:338(validation_step)
        2    0.000    0.000    0.638    0.319 base_model.py:411(validation_step)
        2    0.000    0.000    0.486    0.243 __init__.py:517(create_log)
        2    0.000    0.000    0.458    0.229 base_model.py:429(create_log)
        2    0.000    0.000    0.443    0.222 base_model.py:684(log_prediction)
        2    0.000    0.000    0.288    0.144 __init__.py:679(plot_prediction)
        2    0.000    0.000    0.159    0.079 figure.py:3189(tight_layout)
        2    0.000    0.000    0.157    0.078 tight_layout.py:251(get_tight_layout_figure)
        2    0.000    0.000    0.156    0.078 tight_layout.py:19(_auto_adjust_subplotpars)
        2    0.000    0.000    0.155    0.078 writer.py:641(add_figure)
        4    0.000    0.000    0.155    0.039 _base.py:4570(get_tightbbox)
   958/14    0.004    0.000    0.152    0.011 module.py:1104(_call_impl)
   ...
```

Linux:
```
   Profile stats for: [Strategy]SingleDeviceStrategy.validation_step
    461785 function calls (447573 primitive calls) in 52.269 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.000    0.000   52.270   26.135 strategy.py:338(validation_step)
        2    0.000    0.000   52.270   26.135 base_model.py:411(validation_step)
        2    0.000    0.000   52.139   26.069 __init__.py:517(create_log)
        2    0.000    0.000   52.119   26.059 base_model.py:429(create_log)
        2    0.000    0.000   52.106   26.053 base_model.py:684(log_prediction)
        2    0.000    0.000   45.007   22.504 __init__.py:679(plot_prediction)
        2    0.000    0.000   44.801   22.400 base_model.py:725(plot_prediction)
        2    0.000    0.000   44.781   22.391 pyplot.py:1321(subplots)
        2    0.000    0.000   44.742   22.371 pyplot.py:686(figure)
        2    0.000    0.000   44.742   22.371 pyplot.py:324(new_figure_manager)
        2    0.000    0.000   39.713   19.857 backend_bases.py:3487(new_figure_manager)
        2    0.000    0.000   39.709   19.854 _backend_tk.py:940(new_figure_manager_given_figure)
      ### 250   33.181    0.133   33.186    0.133 {method 'call' of '_tkinter.tkapp' objects}
    ...
```

Cost of **plot_prediction** on linux is more than win 10.

- PyTorch Lightning Version: 1.6.3
- PyTorch forecasting Version: 0.10.1
- PyTorch Version: 1.11.0
- Python version: 3.8.13
- OS : centos 7.6 and win 10
- How you installed PyTorch: pip

### LINKED ISSUE #1426: NHiTS tutorial extremely slow if accelerator changed from 'cpu' to 'auto' or 'gpu'
### Expected behavior

I hoped to get a speed improvement when changing the NHiTS tutorial demo accelerator mode from `cpu` to `gpu`.

### Actual behavior

After changing to `gpu` (and also `auto`), the demo runs extremely slowly.

### Code to reproduce the problem

Change the accelerator in the NHiTS demo from `cpu` to `auto`, then re-run it.

Are there any updates to NHiTS demo, or advice on how to get it working fast in multi-GPU environment?

### LINKED ISSUE #1846: [ENH] 2000% performance improvement by directly without using data loader
Data Loader from pytorch was designed for accessing files from Disk, however, this results to a very slow and very low utilization of high performance accelerators resulting to slower completion and inefficient training. 

This can be alleviated by manually batching via for loop, and can result to about 2000% speed up! So that is very much performance improvement.

De-couple the pytorch from pytorch lighting to enable manual batching

n/a

**Additional context**
n/a

### LINKED ISSUE #1849: [BUG] Slow model training with GPU
**Describe the bug**
As the title says

**To Reproduce**
Use CUDA

```python
<Paste your code here>
```

**Expected behavior**
Faster training with most of the GPU utilized

**Additional context**
na

**Versions**

</details>

### LINKED ISSUE #1860: I suggest you re-implement the logic of data loading
the current implementation performance is poor. dataload process consumed 99% performance time while fitting.
please don't use traditional way of handling large group of time series, which organize multi-dimensions data in a 2-dimensions pd.DataFrame and manupulate it. use tensor please!
use DataFrame as input data type only, and manupulate it always by tensor or np.array, restore every usefull dimension from df columns. if DataFrame manupulation is inevitable, please use AI to find a vectorized way
never pre-process data on time series wise, thounds and thounds times of time cost there.

### LINKED ISSUE #8278: [BUG] Low GPU utilization
**Describe the bug**
GPU is only partially used (4%) even if we have larger amount of data

**To Reproduce**

```python
# import packages
import torch
from sktime.forecasting.base import ForecastingHorizon
from sktime.forecasting.pytorchforecasting import PytorchForecastingNHiTS
from sktime.utils._testing.hierarchical import _make_hierarchical
from sklearn.model_selection import train_test_split
# generate random data
data = _make_hierarchical(
    hierarchy_levels=(64, 512), max_timepoints=100, min_timepoints=100, n_columns=3
)

data_tensor_bfloat16 = torch.tensor(data.values, dtype=torch.bfloat16)
print (data_tensor_bfloat16)


max_prediction_length = 5
fh = ForecastingHorizon(range(1, max_prediction_length + 1), is_relative=True)


x = data[["c0", "c1"]]
y = data["c2"].to_frame()
X_train, X_test, y_train, y_test = train_test_split(
    x, y, test_size=0.2, train_size=0.8, shuffle=False
)
len_levels = len(y_test.index.names)
y_test = y_test.groupby(level=list(range(len_levels - 1))).apply(
    lambda x: x.droplevel(list(range(len_levels - 1))).iloc[:-max_prediction_length]
)

model = PytorchForecastingNHiTS(
    model_params= {
        #"n_blocks": [10,10],
        "activation": "ReLU",
        "hidden_size": 512,
        "n_blocks": [1, 1, 1],
        "n_layers": 2,
    },
    trainer_params={
        "max_epochs": 5, 
        "limit_train_batches": 1,
    },
    train_to_dataloader_params = {"batch_size": 1024},
    validation_to_dataloader_params = {"batch_size": 1024},

)

model.fit(y=y_train, X=X_train, fh=fh) # doctest skip
y_pred = model.predict(fh, X=X_test, y=y_test)
print(y_test)
```

**Expected behavior**

**Additional context**

**Versions**
<details>

```bash
machine: Linux-6.8.0-53-generic-x86_64-with-glibc2.39

Python dependencies:
          pip: 25.0.1
       sktime: 0.37.0
      sklearn: 1.5.2
       skbase: 0.8.3
        numpy: 1.26.4
        scipy: 1.15.2
       pandas: 2.2.2
   matplotlib: 3.10.1
       joblib: 1.4.2
        numba: 0.61.0
  statsmodels: 0.14.4
     pmdarima: 2.0.4
statsforecast: 1.7.8
      tsfresh: 0.21.0
      tslearn: 0.6.3
        torch: 2.7.0
   tensorflow: 2.19.0
```

</details>

jobs-git:
### Description
Fixes: https://github.com/sktime/pytorch-forecasting/issues/1849; Fixes: https://github.com/sktime/pytorch-forecasting/issues/1426; Fixes: https://github.com/sktime/pytorch-forecasting/issues/1860; Fixes: https://github.com/sktime/pytorch-forecasting/issues/922; Fixes: https://github.com/sktime/pytorch-forecasting/issues/715; Fixes: https://github.com/sktime/sktime/issues/8278

Closes: https://github.com/sktime/pytorch-forecasting/issues/1846; Closes: https://github.com/sktime/pytorch-forecasting/issues/991

Supersedes: https://github.com/sktime/pytorch-forecasting/pull/806

The `__getitem__()` and `_collate_fn()` methods in `TimeSeriesDataset` are performance bottlenecks. They run computation extensive branching logic on every call, which temporarily stalls compute devices and slows down training.

To improve in these, the following major changes were made:

- introduced a pre-compute function `__precompute__` that performs the needed calculations conducted by `__getitem__` and `_collate_fn()` then stores batch data to `self.precollate_cache`, the `__precompute__` does not assume its own data order, but follows the `idx` provided by `BatchSamplers`
- introduced fast collate function `__fast_collate_fn__` which performs retrieval of batch data from `self.precollate_cache`
- renamed `__getitem__` to `__item_tensor__` so it can be used in pre-compute
- modify the `__getitem__` to accommodate the usage of old code and fast path when `precompute=True`
- during `precompute=True`, `__getitem__`  will return `None` as only the `return` value of `__fast_collate_fn__` is needed by `DataLoader` to retrieve necessary batch to train models

Note: `TimeSeriesDataset` defaults to slow path, so the old method is used.

### Caveat and Limitations
- This feature stores precomputed data in RAM. Therefore, sufficient memory must be available; otherwise, an out-of-memory (OOM) error may occur.

    Super batching could be considered if `precompute=True` is required. However, implementing that is beyond the scope of this PR.

     Recommend to use: `precompute=False` to use the original slow path.

- For extremely large datasets, FSDP (Fully Sharded Data Parallel) may be used with PyTorch Lightning for distributed training. This, too, is outside the scope of this PR.

     Recommend to use: `precompute=False` to use the original slow path.

### Checklist
- [x] Accelerate `__getitem__()`
- [x] Accelerate `collate_fn()`
- [x] Add timeseries test with `precompute=True`
- [x] Add timeseries test to check output from `precompute=True` is consistent to default behaviour (`precompute=False`)
- [x] Amended changelog for large changes (and added myself there as contributor)
- [x] Added/modified tests


fkiraly (commented):
Nice!

FYI @phoeenniixx, @PranavBhatP, @xandie985 - is this something we need to take into account in v2? We should also think about adding profiling checks.

Code quality (linting) is failing, this can be fixed by using pre-commit or automating formatting. The `pytorch-forecasting` does not have this in the docs (we should add it), but this is the same as in `sktime`:
https://www.sktime.net/en/stable/developer_guide/coding_standards.html

fkiraly:
FYI @agobbifbk

agobbifbk:
Agree, most of the time you can fit your data in memory and we should include the precomputation possibility in the d2 layer. We should have already the correct indexes computed, it is just a matter of create the tensors according to those indexes. 

> When testing with performing tuning with pytorch-lighting it is not able to perform hyperparameter optimization with this info: Failed to compute suggestion for learning rate because there are not enough points. Increase the loop iteration limits or the size of your dataset/dataloader.

When you say `hyperparameter optimization` you mean the learning rate using `trainer.tune` or do you really mean the hyperparameters of the model?

jobs-git:
> Agree, most of the time you can fit your data in memory and we should include the precomputation possibility in the d2 layer. We should have already the correct indexes computed, it is just a matter of create the tensors according to those indexes.
> 
> > When testing with performing tuning with pytorch-lighting it is not able to perform hyperparameter optimization with this info: Failed to compute suggestion for learning rate because there are not enough points. Increase the loop iteration limits or the size of your dataset/dataloader.
> 
> When you say `hyperparameter optimization` you mean the learning rate using `trainer.tune` or do you really mean the hyperparameters of the model?

This is not an issue anymore, I missed removing the fit function (conducting during my testing) I placed prior to the Tuner.

phoeenniixx:
Hi @jobs-git, I see you are facing some linting issues, may I suggest you try setting up pre-commit in your local repo, the process is similar to `sktime` as mentioned by @fkiraly 
https://www.sktime.net/en/stable/developer_guide/coding_standards.html#setting-up-local-code-quality-checks

pytorch_forecasting/data/timeseries/_timeseries.py
```diff
@@ -480,6 +480,10 @@ def __init__(
         """Timeseries dataset holding data for models."""
         super().__init__()
 
+        self.precompute_cache = []
+        self.precompute_idx = 0
+        self.precompute = False
```
fkiraly: should this be an `__init__` arg so the user can control it?

jobs-git: That can be done, I can move the `precompute` param from to_dataloader to to `__init__`

How about the other highlighted variables? These are more like internally used, so those I think need not be added to `__init__` arg

fkiraly: > How about the other highlighted variables?

They are not "settings", are they? Which ones do you mean?

jobs-git: These two:

```
        self.precompute_cache = []
        self.precompute_idx = 0
```
are intended to be variables, but they need to be set in `__init__` function as these are being accessed within the TimeSeriesDataset that is not easy to pass from method to method.

The `precompute_cache` is being accessed by two methods that don't directly communicate with each other, while the `precompute_idx` is a counter that TimeSeriesDataset does not have control into as its being iterated by Pytorch DataLoader via `__getitem__()` calls, but we need that variable so we can track which item gets to be fetch next from `precompute_cache`.

fkiraly: these look like internal variables, I do not think the user will want to configure this.

fkiraly:
is this ready for review?

jobs-git:
> is this ready for review?

Yes, please. Inputs are welcome.

fkiraly (changes requested):
Thanks - while we wait for reviews, can you kindly add tests? For `TimeSeriesDataSet` in isolation, and in integration with the networks?

pytorch_forecasting/data/timeseries/_timeseries.py
```diff
@@ -2342,6 +2347,51 @@ def __getitem__(self, idx: int) -> tuple[dict[str, torch.Tensor], torch.Tensor]:
...
+        """
+        Get sample for model
+
+        Args:
+            idx (int): index of prediction (between ``0`` and ``len(dataset) - 1``)
+
+        Returns:
+            tuple[dict[str, torch.Tensor], torch.Tensor]: x and y for model
+        """
+        if self.precompute:
```
fkiraly, line 2429: can you explain why this is correct if `self.precompute`?

The index `idx` is not used at all, so this looks wrong.

fkiraly (changes requested):
Thanks - can you explain the logic in the `self.precompute` branch of `__getitem__`? It does not look correct to me.

pytorch_forecasting/data/timeseries/_timeseries.py
jobs-git: Its being used under `__item_tensor__` which is just a rename of the original `__getitem__` its input param is not changed including it accepting the `idx`. 

But the mechanism at which it gets `idx` is different.

jobs-git: When `self.precompute=True`, `__item_tensor__` gets the `idx` **first**, so when this happens, the **second time** we receive `idx` in `__getitem__` it will now be a different set. This will change the actual behavior of the original `__getitem__` so we dont use it, instead we follow the order from `self.precompute_cache` which contains the correct/original order of `idx` from the **first call**.

When `self.precompute=True`, the order at which the `idx` gets obtained by `__item_tensor__` is as follows:
1. create `Dataloader` by calling `to_dataloader` 
2. this calls `__precompute__`
3. `__precompute__` follows `TimeSynchronizedBatchSampler` to get `idx`
4. passes `idx` **first sequences** to `__item_tensor__`

Caveat:
So when Dataloader generates its own `idx` from its own `Sampler` it is not used, if we use the Dataloader `idx` we will NOT get the benefit of this PR, as that `idx` is being generated real time i guess. 

So, instead of retrieving from `cache`, the effect is we will be back to computing original `__getitem__` on each call, which is the slow path all over again, starving the GPU, technically is the source of slowness.

However, if a user wishes to use the `idx` of the `Dataloader`, the default will just skip precomputation, and get the original slow path from `__getitem__` -> `__item_tensor__` - which is the same vanilla `__getitem__` anyway.

jobs-git:
Details has been added to the PR

pytorch_forecasting/data/timeseries/_timeseries.py
fkiraly: > When `self.precompute=True`, `__item_tensor__` gets the `idx` **first**, so when this happens, the **second time** we receive `idx` in `__getitem__` it will now be a different set.

Sorry, I still do not get it. Can you explain which dunders are called from external, and which are only used internally? I thought that `__getitem__` gets called - so when it does not use `idx`, the information is not passed on.

jobs-git: **TLDR**
Does this PR use `idx` from `Dataloader`? **YES** and **NO**. If using defaults (`precompute=False`), then **YES**. But if using accelerated path (`precompute=True`), then **NO**.

**EXPLANATIONS**

- **Vanilla**
  ```bash
  Dataloader -> pass idx -> __getitem__() -> use idx
  ```
- **PR - Revised vanilla - slow path (`precompute=False`)**
  ```bash
  Dataloader -> pass idx -> __getitem__() -> pass idx -> 
  __itemtensor__() -> use idx
  ```
  Notes:
  - `__getitem__()` is revised so it can switch to **vanilla** and **accelerated** based on `precompute` value, in this case, its using vanilla.
  - the `__itemtensor__()` is just the renamed `__getitem__()` so its the usual calculation path when `precompute=False`

  Verdict: 
  - `idx` from `Dataloader` is used, as usual.


- **PR - Accelerated path (`precompute=True`)**

    Two steps happen here
    - a. Pre computation routine
    
      ```bash
      to_dataloader () -> __precompute__() -> generates idx from TSDS -> 
      loop on every idx -> pass each idx -> __itemtensor__() -> use idx -> 
      save to data to cache -> move to the next idx -> repeat
      ```
    
      This is essentially how Dataloader uses `__getitem__()`. It calls `__getitem__() `and pass `idx` to it, repeats until all `idx` exhaust the data. But, what I did in this PR is used the **FOR LOOP** to retrieve `idx` in advance and exhaust the data so we can pre-calculate it without calling the `Dataloader`.
    
    - b. `Dataloader` call
    
      Since we have cached the result of `__itemtensor__()` in advanced. Meaning `__itemtensor__()` has the pre-computed the data on each specific `idx`. We just retrieve it.
    
      ```bash
      Dataloader -> pass idx -> __getitem__() -> idx ignored ->  use cache sequence
      ```
      Note:
      - notice, we dont need to call the `__itemtensor__()` this time since it already did its job
  
      Verdict: 
      - `idx` from `Dataloader` is not essential anymore, since another `idx` source has been used, specifically from `TimeSynchronizedBatchSampler`

fkiraly: @jobs-git, apologies, I still do not get it.

> Since we have cached the result of itemtensor in advanced. meaning itemtensor has the precomputed the data on each specific idx. We just retrieve it.

As far as I understand: you are not retrieving the expected result at `idx`. You are retrieving the results in sequence, irrespective of which `idx` gets queried. Therefore, this is not the correct logic for `__getitem__`.

Please let me know if you disagree.

@phoeenniixx, @PranavBhatP, @agobbifbk, @fnhirwa - can you perhaps also have a look? I also do not fully understand the explanation.

jobs-git: For us to be able to accelerate this (`precomputation=True`), we need to query the `idx`'s in advance (not wait for `Dataloader` to give it one by one) to perform precomputation. This is what the PR does.

I am confused by your meaning of **sequence**, since `idx`'s from `TimeSynchronizedBatchSampler` is always **random** when `shuffle=True`, we just saved the result sequentially so retrieval is also sequence.

As mentioned previously, acceleration requires obtaining the `idx`'s in advance, otherwise, we can't perform pre-computation.

jobs-git: @fkiraly I think what you prefer implemented is obtaining the precomputed values from real-time `idx`? This is what I prefer too, until I saw that the algorithm seems to be **normalizing** on a given `idx`. 

That is when I realized we cant accelerate this if we wait for `Dataloader`, so we need to obtain idx's in advance from a different mechanism. Thus this PR.

May I ask what is the actual concern if the accelerated path uses a different mechanism to obtain random `idx`?

Of course, the exact training trajectory would likely be different, but a properly trained model should resolve to the same/similar outcome. If I could present prediction results, would that settle the concern for accelerated path using a different random `idx` generator? Just let me know.

fkiraly:
ok, I think I am starting to get it. @jobs-git, would it be possible to produce code for a basic usage examples?

Then, for each call of `__getitem__` and `__itemtensor__` , list which `idx` the different functions are called with? For the `precompute=False` and `precompute=True` case.

pytorch_forecasting/data/timeseries/_timeseries.py
```diff
@@ -2342,6 +2347,51 @@ def __getitem__(self, idx: int) -> tuple[dict[str, torch.Tensor], torch.Tensor]:
...
+            if self.precompute_idx >= len(self.precompute_cache):
```
fkiraly: could we move this logic into an extra private function `_retrieve_precomputed_and_increment_idx` (or similar)? Just to make this function easier to read, and clearer what is happening.

jobs-git:
> ok, I think I am starting to get it. @jobs-git, would it be possible to produce code for a basic usage examples?
> 
> Then, for each call of `__getitem__` and `__itemtensor__` , list which `idx` the different functions are called with? For the `precompute=False` and `precompute=True` case.

Yes, of course, its possible to add sample code, regarding the term `precompute`, is that ok? I think it should be have been `preprocess`, but I noticed there is a usage of `preprocess` already, so I did not use it to avoid confusion.

fkiraly:
@jobs-git, this explanation would be much appreciated! Alternatively, you could join the community meet-up on Friday (13 UTC) and explain there

jobs-git:
> @jobs-git, this explanation would be much appreciated! Alternatively, you could join the community meet-up on Friday (13 UTC) and explain there

Hi, apologies for late response, I found another bottleneck in TimeSeries so I was fixing it. I will push it soon, but with the updated version, we can now get 2100% performance improvement on the same sample. In big datasets that can fit in 1 VRAM, that I think that could translate to just needing 1x GPU from a previous needed 21x GPUs.

But as previously discussed, the behavior would be different. But I dont think this should be much concern on a properly trained model with appropriate amount of data. We could probably discuss achieving close to parity with old behaviour in another PR.

Anyway, I will push it soon.

fkiraly:
> We could probably discuss achieving close to parity with old behaviour in another PR.

We need exact parity for any previously working code - we cannot push a breaking change with a release - it would be great if we could also ensure this with a test. Namely, that `__getitem__` returns exactly the batch that previously was returned.

jobs-git:
> > We could probably discuss achieving close to parity with old behaviour in another PR.
> 
> We need exact parity for any previously working code - we cannot push a breaking change with a release - it would be great if we could also ensure this with a test. Namely, that `__getitem__` returns exactly the batch that previously was returned.

Got it, I will test this and add that as sample, and probably some demonstration on available real world samples from sktime too.

fkiraly:
@jobs-git, are you still working on this? Let us know if you need help!

jobs-git:
> @jobs-git, are you still working on this? Let us know if you need help!

nothing anymore, it's all complete

fkiraly:
I see, then you should notify reviewers by "re-request review"!

image

fkiraly (changes requested):
quick request: can you kindly move the precompute_bench notebooks to another branch? They have quite large size.

jobs-git:
> quick request: can you kindly move the precompute_bench notebooks to another branch? They have quite large size.

Sure, will separate the notebooks.

jobs-git:
The [commit](https://github.com/sktime/pytorch-forecasting/pull/1850/commits/7f979639d0f4e01554254169a39ef403dd75a3aa) to achieve close to parity algorithm with the old batching lowered the gain. So far, I can achieve about 900% in synthetic data.

Will create a separate PR that demonstrate its usage and benchmark performance.

PR #1853, opened 2025-05-28
[MNT] Replace Black with Ruff formatting and update configuration

Borda:
### Description

Switched from Black to Ruff for code formatting to simplify tooling. Updated `.pre-commit-config.yaml` and `pyproject.toml` to reflect the removal of Black and added Ruff-specific formatting configurations. This change ensures consistency and leverages Ruff's integrated formatting capabilities.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ line-ending = "auto"
 select = ["E", "F", "W", "C4", "S"]
 extend-select = [
   "I", # isort
+  "UP", # pyupgrade
   "C4", # https://pypi.org/project/flake8-comprehensions
 ]
 extend-ignore = [
--- a/pytorch_forecasting/data/samplers.py
+++ b/pytorch_forecasting/data/samplers.py
@@ -47,13 +47,11 @@ class GroupedSampler(Sampler):
         ):
             raise ValueError(
                 "batch_size should be a positive integer value, "
-                "but got batch_size={}".format(batch_size)
+                f"but got batch_size={batch_size}"
             )
         if not isinstance(drop_last, bool):
             raise ValueError(
-                "drop_last should be a boolean value, but got drop_last={}".format(
-                    drop_last
-                )
+                f"drop_last should be a boolean value, but got drop_last={drop_last}"
             )
         self.sampler = sampler
         self.batch_size = batch_size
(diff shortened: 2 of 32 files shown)


pytorch_forecasting/data/samplers.py
```diff
@@ -51,8 +51,9 @@ def __init__(
             )
         if not isinstance(drop_last, bool):
             raise ValueError(
-                "drop_last should be a boolean value, but got "
-                "drop_last={}".format(drop_last)
+                "drop_last should be a boolean value, but got " "drop_last={}".format(
```
Borda:
```suggestion
                "drop_last should be a boolean value, but got drop_last={}".format(
```

pyproject.toml
```diff
@@ -170,30 +177,6 @@ force-sort-within-sections = true
   "E501", # Line too long being fixed in #1746 To be removed after merging
 ]
 
-[tool.black]
-line-length = 88
-include = '\.pyi?$'
-exclude = '''
-(
-  /(
-      \.eggs         # exclude a few common directories in the
```
fkiraly, line 180: where did this go? Explanation appreciated.

pytorch_forecasting/data/samplers.py
```diff
@@ -51,8 +51,9 @@ def __init__(
             )
         if not isinstance(drop_last, bool):
             raise ValueError(
-                "drop_last should be a boolean value, but got "
-                "drop_last={}".format(drop_last)
+                "drop_last should be a boolean value, but got drop_last={}".format(
```
fkiraly: can we use `f"`?

pyproject.toml
Borda: it is in the Ruff section, see:
https://github.com/sktime/pytorch-forecasting/blob/de1a4f1231775d238acbdce491c8078f0a29f346/pyproject.toml#L132-L142

pytorch_forecasting/data/samplers.py
Borda: Yes, it would be better, but that could be done with adding a `UP` rule, see #1856

fkiraly:
FYI @fnhirwa, review would be appreciated due to `ruff`

PR #1855, opened 2025-05-29
[ENH] address `__array_wrap__` change incompatibility between `numpy 2` and early `torch` or `pandas` versions

### LINKED ISSUE #1764: [ENH] address `numpy 2` deprecations
Current test execution causes these warnings:

`Warning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated NumPy 2.0)`

We should diagnose the source and address the deprecation - noting that code should work under `numpy 1` as well as `numpy 2`.

fkiraly:
Fixes #1764.

The reason is that `numpy 2` added additional arguments to `__array_wrap__` which was interacting with `torch.Tensor.__array_wrap__` or `pandas.Series.__array_wrap__` that did not have said arguments on earlier versions, in `TorchNormalizer`.

The warning can be avoided by explicit coercion to `torch` prior to mixed usage - avoiding mixed types across packages entirely.

This also makes the coercions more explicit, enabling future refactors where we may want to move to more "clean" type assumptions instead of polymorphic union types.

### DIFF
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -712,18 +712,26 @@ class TorchNormalizer(
             center = center.view(*center.size(), *(1,) * (y.ndim - center.ndim))
             scale = scale.view(*scale.size(), *(1,) * (y.ndim - scale.ndim))
 
-        if isinstance(y, torch.Tensor):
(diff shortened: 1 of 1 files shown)


fkiraly:
Superseded by https://github.com/sktime/pytorch-forecasting/pull/1911

PR #1856, opened 2025-05-29
[MNT] Add 'UP' to extend-select for pyupgrade python syntax

Borda:
### Description

This PR updates the ruff hook in ./pre-commit-config.yaml to include the --extend-select=UP argument, adding the UP rule group so that PyUpgrade’s syntax checks (codes starting with UP) are enforced; this enables Ruff to detect and auto-fix PyUpgrade suggestions—such as modernizing f-strings and removing deprecated syntax.

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,6 +146,7 @@ target-version = "py39"
 select = ["E", "F", "W", "C4", "S"]
 extend-select = [
   "I", # isort
+  "UP", # pyupgrade
   "C4", # https://pypi.org/project/flake8-comprehensions
 ]
 extend-ignore = [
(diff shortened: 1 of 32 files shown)


fkiraly:
Nice! How does this make sure that we do not upgrade past the lower boundary of our support range, i.e., python 3.9?

Borda:
> Nice! How does this make sure that we do not upgrade past the lower boundary of our support range, i.e., python 3.9?

This is setting the lower bound https://github.com/sktime/pytorch-forecasting/blob/fda4a4c873926e359466b478aed92d3104f5f5d6/pyproject.toml#L143

fkiraly (approved):

PR #1859, opened 2025-05-29
[ENH] `DecoderMLP` metadata container for v1 tests

fkiraly:
Adds `DecoderMLP` metadata container for v1 tests.

Stacks on https://github.com/sktime/pytorch-forecasting/pull/1859


pytorch_forecasting/models/mlp/_decodermlp_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+        from pytorch_forecasting.metrics import (
+            MAE,
+            CrossEntropy,
+            MultiLoss,
+            QuantileLoss,
+        )
+
+        return [
+            {},
+            dict(train_only=True),
```
phoeenniixx: I think the error is due to this param, is `train_only` being sent to `kwargs` correctly?

fkiraly: yes, fixed it in the following way:

it looks like `train_only` was controlling the tests, and if `True` only a part of the tests was run. It does not contribute to coverage *via* tests, so I removed it entirely.
(it was not present in `_integration` to start with)

PR #1861, opened 2025-05-30
[ENH] refactor test metadata container to include data loader configs

fkiraly:
Refactors the test metadata container to include data loader configs which may vary, adds this in a private class method `_get_test_dataloaders_from`. Also refactors the `_integration` method in `TestAllEstimators` accordingly.

Also:

* Adds `NBeats` metadata container for v1 tests.
* Removes the private import from `test_tide`, of `_integration` from `test_all_estimators`

### DIFF
--- a/pytorch_forecasting/models/deepar/_deepar_metadata.py
+++ b/pytorch_forecasting/models/deepar/_deepar_metadata.py
@@ -123,3 +123,53 @@ class DeepARMetadata(_BasePtForecaster):
                 trainer_kwargs=dict(accelerator="cpu"),
             ),
         ]
+
+    @classmethod
+    def _get_test_dataloaders_from(cls, params):
+        """Get dataloaders from parameters.
+
+        Parameters
+        ----------
+        params : dict
+            Parameters to create dataloaders.
+            One of the elements in the list returned by ``get_test_train_params``.
+
+        Returns
+        -------
+        dataloaders : tuple of three torch.utils.data.DataLoader
+            List of dataloaders created from the parameters.
+            Train, validation, and test dataloaders, in this order.
+        """
+        trainer_kwargs = params.get("trainer_kwargs", {})
+        clip_target = params.get("clip_target", False)
+        data_loader_kwargs = params.get("data_loader_kwargs", {})
+
+        from pytorch_forecasting.metrics import NegativeBinomialDistributionLoss
+        from pytorch_forecasting.tests._conftest import make_dataloaders
+        from pytorch_forecasting.tests._data_scenarios import data_with_covariates
+
+        dwc = data_with_covariates()
+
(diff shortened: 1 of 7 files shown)


fkiraly:
Thanks @phoeenniixx, @PranavBhatP, it looks like it works now!

I have also opened an issue here, on the design topic related to this paramerer: https://github.com/sktime/pytorch-forecasting/issues/1865 - we should review whether this should be captured in a tag, for instance.

fkiraly:
Hm, not actually.

It seems the variable part is bigger than I originally cut, so the data loaders need to move into the metadata class.
Attempt in this update.

phoeenniixx:
I have an idea, not sure if it supports the "unified API" but it can make things a little bit easier for us:

What if we create a default `_integration` that optimally tests all models (like right now), But, if some models donot follow the "defaults" of this function and needs totally different params, then it can override the default `_integration` and can create its "own" `_integration`, that can be used for testing?

fkiraly:
> What if we create a default `_integration` that optimally tests all models (like right now), But, if some models donot follow the "defaults" of this function and needs totally different params, then it can override the default `_integration` and can create its "own" `_integration`

I do not think that is a good idea - it is more or less the current status quo. We could have just moved the entire test into the metadata class, already.

This is "cutting too wide" in my opinion.

phoeenniixx:
Yes that makes sense

phoeenniixx:
> @phoeenniixx, @PranavBhatP - do you think it is ok to merge this, or will this cause conflicts with your branches?

I think on my branch, I am only changing `test_all_estimators.py`, rest files are unchanged
The change in `test_all_estimators.py` is mainly just overriding `_all_objects` in `TestAllPtForecasters` so that it doesnot collects the `ptf-v2` objects - as we discussed on one of the tech meets

phoeenniixx:
> I do not think you need to override it, only add the filter tag. I commented in that PR.

I added the filter tag to `TestAllPtForecastersV2` but how to prevent `TestAllPtForecasters` from collecting all metadata classes using this filter tag? Because how would `TestAllPtForecasters` will know not to take `ptf-v2` objects?

fkiraly:
> Because how would `TestAllPtForecasters` will know not to take `ptf-v2` objects?

You add the same tag for v1 objects - `"forecaster_pytorch"` - to the v1 test class, so it will not collect the v2 classes.

phoeenniixx:
> > Because how would `TestAllPtForecasters` will know not to take `ptf-v2` objects?
> 
> You add the same tag for v1 objects - `"forecaster_pytorch"` - to the v1 test class, so it will not collect the v2 classes.

Then I'll have to make changes to all metadata classes... Sure I can do that as well
Thanks

fkiraly:
ok, will merge then to avoid conflicts and too high stacks

PR #1864, opened 2025-05-30
Clearly define the MLP as a class/nn.model

jobs-git:
In the nhits paper: https://arxiv.org/abs/2201.12886, shows that the MLP as component of the Block, the current code does integrates this already as series of appended layers. 

This pull request instead clearly defines the MLP as a class/nn.model.

### Description

This PR ...

### Checklist

- [ ] Added/modified tests

Make sure to have fun coding!


fkiraly:
seems like there are still some failures - are these caused by your changes?

fkiraly (approved):
Looks good now!

Issue #1614, opened 2024-08-23
[BUG] `test_none_reduction` fails on main


fkiraly:
The `test_none_reduction` test fails on main, with 

```
>                   reduced = torch.cat([global_state, local_state])
E                   RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
```

This seems like change in upcasting, and is resolved by aggregating via `mean` instead, which also seems to produce the correct result.

fkiraly:
This is the `pip list`

```
Package                                 Version
--------------------------------------- -----------
absl-py                                 2.1.0
accessible-pygments                     0.0.5
aiohappyeyeballs                        2.4.0
aiohttp                                 3.10.5
aiosignal                               1.3.1
alabaster                               1.0.0
alembic                                 1.13.2
appnope                                 0.1.4
astroid                                 3.2.4
asttokens                               2.4.1
async-timeout                           4.0.3
attrs                                   24.2.0
babel                                   2.16.0
backports.functools-lru-cache           2.0.0
beautifulsoup4                          4.12.3
black                                   24.8.0
bleach                                  6.1.0
certifi                                 2024.7.4
cfgv                                    3.4.0
charset-normalizer                      3.3.2
click                                   8.1.7
cmaes                                   0.11.1
colorlog                                6.8.2
comm                                    0.2.2
commonmark                              0.9.1
contourpy                               1.2.1
coverage                                7.6.1
cpflows                                 0.1.2
cycler                                  0.12.1
debugpy                                 1.8.5
decorator                               5.1.1
defusedxml                              0.7.1
dill                                    0.3.8
distlib                                 0.3.8
docutils                                0.21.2
exceptiongroup                          1.2.2
execnet                                 2.1.1
executing                               2.0.1
setuptools                              65.5.0
six                                     1.16.0
snowballstemmer                         2.2.0
soupsieve                               2.6
Sphinx                                  8.0.2
sphinxcontrib-applehelp                 2.0.0
sphinxcontrib-devhelp                   2.0.0
sphinxcontrib-htmlhelp                  2.1.0
sphinxcontrib-jsmath                    1.0.1
sphinxcontrib-qthelp                    2.0.0
sphinxcontrib-serializinghtml           2.0.0
SQLAlchemy                              2.0.32
stack-data                              0.6.3
statsmodels                             0.14.2
subprocess32                            3.5.4
sympy                                   1.13.2
tensorboard                             2.17.1
[...]

PR #1872, opened 2025-06-04
[BUG] Fix for the case when reduction is set to `none`

fnhirwa:
### Description

fixes https://github.com/sktime/pytorch-forecasting/issues/1614
This PR fixes the tests for the reduction set to none for the `MultiHorizonMetric` class.

Using `cat` needs the `default` tensor is expected to be of at least one dimension.


pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -817,16 +817,28 @@ class MultiHorizonMetric(Metric):
...
-        )
-        self.add_state(
-            "lengths",
-            default=torch.tensor(0),
-            dist_reduce_fx="sum" if reduction != "none" else "mean",
-        )
+        if reduction == "none":
+            self.add_state(
+                "losses",
+                default=torch.tensor([], dtype=torch.float),
```
fkiraly: would this better be addressed by dispatching on the argument?

```python
if reduction == "none":
    val_loss = val_len = []
else:
    val_loss = 0.
    v_len = 0
```

fnhirwa: Yeah, this one can work, updating it.

fkiraly (approved):

PR #1875, opened 2025-06-05
[WIP] [ENH] Add timeseries integration test for various models

jobs-git:
Add timeseries integration test for various models


fkiraly (changes requested):
This is nice, thanks!!

One change request, could you kindly add it instead to the class named `TestAllPtForecasters` as a new test?

To loop over all classes like you are doing, use the `object_class` fixture, like in `test_doctest_examples` - this will take values of the models.

(not all are retieved at the moment, they require a metadata class like `DeepARMetadata` - but adding more metadata records would also be a nice contribution later)

fkiraly:
(also, code formatting fails - I suggest to set up `pre-commit` locally - works like this https://www.sktime.net/en/stable/developer_guide/coding_standards.html)

PR #1878, opened 2025-06-08
[BUG] enable silenced TFT v2 tests

fkiraly:
The tests in the file `_test_tft_v2` were not run since it started with an underscore.

This PR removes the underscore to enable those tests to run.


fkiraly:
FYI @phoeenniixx, I think you added the file?

Issue #1870, opened 2025-06-02
[API] M layer design


fkiraly:
Issue to discuss the M layer design. For context, see design document here: https://github.com/sktime/enhancement-proposals/pull/39

My current proposed design, this is base on v1 and v2 metadata layer design. Long-term state:

* starts off the current "metadata" class, but named like the model, e.g., `TFT`. The current ligthning network is renamed `TFT_NN`
* `__init__` has all args of two objects: the loader (D2, e.g., `DecoderEncoderModule`) and the network (e.g., `TFT_NN`). minus data
* method `get_loader_class` gets the loader class (e.g., the class `DecoderEncoderModule`); `get_loader(data: TimeSeries)` produces an loader object, an instance of the `get_loader_class` return.
* method `get_nn_class` returns the nn class (e.g., `TFT_NN`); `get_nn(loader)` gets an instance of the nn class.
* finally, there is a method `init(data)`, which calls the above in sequence, and produces a pair of loader and nn, as if the two `get` methods were called in sequence.
* `__call__` dispatches to `init` 

So, a usage vignette could look like:

```python
from pytorch_forecasting.models import TFT

dataset = TimeSeries(...)
tft_cfg = TFT(
    max_encoder_length=30,
    max_prediction_length=1,
    batch_size=32,
    loss=nn.MSELoss(),
    logging_metrics=[MAE(), SMAPE()],
    optimizer="adam",
    hidden_size=64,
    num_layers=2,
    attention_head_size=4,
)

net, loader = tft_cfg()

trainer = Trainer(
    max_epochs=5,
    accelerator="auto",
    devices=1,
    enable_progress_bar=True,
    log_every_n_steps=10,
)

trainer.fit(net, loader)

etc
```

fkiraly:
FYI @agobbifbk, @PranavBhatP , @phoeenniixx , @xandie985

agobbifbk:
Nice!!!
Some comments:
- `logging_metrics=[MAE(), SMAPE()],` in my opinion should be managed by the `Trainer` with the argument `callbacks=...`
- with this design, are we loosing the flow 'D1->D2->M'? It seems that 'M' and 'D2' are create in the same function, can this  cause some problem to the user? We are mixing model parameters, loss and optimizer in the same place
[...]

PR #1888, opened 2025-06-12
[ENH] two-way linkage of model package classes and neural network classes

fkiraly:
Stacks on https://github.com/sktime/pytorch-forecasting/pull/1892

This PR adds a `pkg` property to models that is also tested to be inverse to `get_model_cls` interface of the package class.

To be discussed if this is a good idea - as previously discussed in the meet-up, the exact use case for the `pkg` method is unclear.


fkiraly:
FYI @agobbifbk, @phoeenniixx, @PranavBhatP , @xandie985

pytorch_forecasting/models/deepar/_deepar_pkg.py
phoeenniixx, line 1: will it make sense to move all `pkg` files to a new location maybe like `model_pkg` or something just to keep everything at one place and designate a specific place for such packages?

pytorch_forecasting/models/deepar/_deepar.py
```diff
@@ -35,6 +35,15 @@
 
 
 class DeepAR(AutoRegressiveBaseModelWithCovariates):
+    """DeepAR: Probabilistic forecasting with autoregressive recurrent networks."""
+
+    @classmethod
+    def _pkg(cls):
```
phoeenniixx, line 41: Also, I have a doubt about having model the access of this package - is it for eventual movement to M layer architecture you proposed in #1870, or is there any other reason?

pytorch_forecasting/models/deepar/_deepar_pkg.py
fkiraly: I think that is not good idea - like with python package, imo, the best place for "package metadata" (like `pyproject.toml`) is imo with the package contents, not in a central place.

From a design perspective, this is minimizing the "extension surface" - with your proposal, an extender has to add information in two places. This is worse than having to add two things in a single place.

pytorch_forecasting/models/deepar/_deepar.py
fkiraly: Good question - I thought it might be useful, because that way the model can access its tags programmatically.
The tags might be useful in the data loading and checking boilerplace, for instance.
(though I admit this is not a very strong argument)

And, yes, the M layer archtitecture might also profit from the "pointers".

pytorch_forecasting/models/deepar/_deepar_pkg.py
fkiraly: or did I maybe misunderstand: were you, or were you not, suggesting, to collect all `_pkg` files or classes in a single location?

phoeenniixx: yes this is what I was suggesting - to  move `_pkg` files to a single location - now I understand why this might be a bad idea, Thanks!!

fkiraly:
Based on our discussion, @phoeenniixx, now split up.

Rename and bugfixes to test class, minor docstring imporvements are now here: https://github.com/sktime/pytorch-forecasting/pull/1892

This PR now only contains the addition of the `pkg` method, which we can delay until we close it or we have a clear design purpose.

PR #1889, opened 2025-06-12
[ENH] `Tide` model in `v2` interface

phoeenniixx:
Adds `Tide` model from `dsipts` to `ptf-v2`

Created a new folder `tide_dsipts` in `tide` that contains all the necessary parts for the tide


phoeenniixx:
Hi @fkiraly, @agobbifbk, I would appreciate some reviews on the model, the tests pass but the issue is if we add a non empty `embs` - it leads to errors, is this a compatibility issue or some other issue with code?  
maybe while trying to  make it compatible with the data module i broke something? although I just tried to map `dspits` to `ptf` terminology 

Or maybe  I am just not entering correct params?

I did make some changes to `base_dsipts` and `utils` but that is I think just to make it a little bit "simpler" so that data module can handle it - just tried to make `compute_loss` of `base` a little simpler by removing the loss functions that we might not use for now....

phoeenniixx:
Failing param:
```python
       dict(
                embs=[3, 3],
                hidden_size=64,
                d_model=32,
                n_add_enc=3,
                n_add_dec=2,
                dropout_rate=0.1,
                activation="",
                data_loader_kwargs=dict(max_encoder_length=4, max_prediction_length=2),
            ),
```
Error message of the failing param (copied from the workflow):
```
self = <pytorch_forecasting.tests.test_all_estimators_v2.TestAllPtForecastersV2 object at 0x1176f6270>
object_metadata = <class 'pytorch_forecasting.models.tide.tide_dsipts.tide_v2_metadata.TIDEMetadata'>
trainer_kwargs = {'activation': '', 'd_model': 32, 'data_loader_kwargs': {'max_encoder_length': 4, 'max_prediction_length': 2}, 'dropout_rate': 0.1, ...}
tmp_path = PosixPath('/private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/pytest-of-runner/pytest-0/test_integration_Tide_2_0')

    def test_integration(
        self,
        object_metadata,
        trainer_kwargs,
        tmp_path,
    ):
        object_class = object_metadata.get_model_cls()
        dataloaders = object_metadata._get_test_datamodule_from(trainer_kwargs)
    
>       _integration(object_class, dataloaders, tmp_path, **trainer_kwargs)

pytorch_forecasting/tests/test_all_estimators_v2.py:97: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pytorch_forecasting/tests/test_all_estimators_v2.py:66: in _integration
    trainer.fit(
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py:561: in fit
    call._call_and_handle_interrupt(
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/call.py:48: in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py:599: in _fit_impl
    self._run(model, ckpt_path=ckpt_path)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py:1012: in _run
    results = self._run_stage()
              ^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py:1054: in _run_stage
    self._run_sanity_check()
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py:1083: in _run_sanity_check
    val_loop.run()
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/loops/utilities.py:179: in _decorator
    return loop_run(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/loops/evaluation_loop.py:145: in run
    self._evaluation_step(batch, batch_idx, dataloader_idx, dataloader_iter)
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/loops/evaluation_loop.py:437: in _evaluation_step
    output = call._call_strategy_hook(trainer, hook_name, *step_args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/trainer/call.py:328: in _call_strategy_hook
    output = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/lightning/pytorch/strategies/strategy.py:412: in validation_step
    return self.lightning_module.validation_step(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pytorch_forecasting/models/tide/tide_dsipts/base_dsipts.py:191: in validation_step
    y_hat = self(batch)
            ^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1751: in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1762: in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pytorch_forecasting/models/tide/tide_dsipts/tide.py:186: in forward
    emb_cat_past, emb_cat_fut = self.cat_categorical_vars(batch)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pytorch_forecasting/models/tide/tide_dsipts/tide.py:306: in cat_categorical_vars
    emb_cat_full = self.emb_cat_var(cat_full, self.device)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1751: in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1762: in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pytorch_forecasting/models/tide/tide_dsipts/sub_nn.py:58: in forward
    cat_n_embd = self.get_cat_n_embd(cat_vars)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pytorch_forecasting/models/tide/tide_dsipts/sub_nn.py:89: in get_cat_n_embd
    emb = layer(cat_vars[:, :, index])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1751: in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/module.py:1762: in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/modules/sparse.py:190: in forward
    return F.embedding(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = tensor([[0, 1, 2, 3, 4, 5],
        [0, 1, 2, 3, 4, 5]])
weight = Parameter containing:
tensor([[-1.7752e+00, -4.5896e-01,  9.9723e-01, -1.3401e+00, -3.1466e-01,
          2.9700e-01, ...00,  5.9457e-01, -1.1853e+00,
         -1.5251e+00, -6.5680e-01,  6.4112e-01,  1.7847e-01]],
       requires_grad=True)
padding_idx = -1, max_norm = None, norm_type = 2.0, scale_grad_by_freq = False
sparse = False

    def embedding(
        input: Tensor,
        weight: Tensor,
        padding_idx: Optional[int] = None,
        max_norm: Optional[float] = None,
        norm_type: float = 2.0,
        scale_grad_by_freq: bool = False,
        sparse: bool = False,
    ) -> Tensor:
        r"""Generate a simple lookup table that looks up embeddings in a fixed dictionary and size.
    
        This module is often used to retrieve word embeddings using indices.
        The input to the module is a list of indices, and the embedding matrix,
        and the output is the corresponding word embeddings.
    
        See :class:`torch.nn.Embedding` for more details.
    
        .. note::
            Note that the analytical gradients of this function with respect to
            entries in :attr:`weight` at the row specified by :attr:`padding_idx`
            are expected to differ from the numerical ones.
    
        .. note::
            Note that `:class:`torch.nn.Embedding` differs from this function in
            that it initializes the row of :attr:`weight` specified by
            :attr:`padding_idx` to all zeros on construction.
    
        Args:
            input (LongTensor): Tensor containing indices into the embedding matrix
            weight (Tensor): The embedding matrix with number of rows equal to the maximum possible index + 1,
                and number of columns equal to the embedding size
            padding_idx (int, optional): If specified, the entries at :attr:`padding_idx` do not contribute to the gradient;
                                         therefore, the embedding vector at :attr:`padding_idx` is not updated during training,
                                         i.e. it remains as a fixed "pad".
            max_norm (float, optional): If given, each embedding vector with norm larger than :attr:`max_norm`
                                        is renormalized to have norm :attr:`max_norm`.
                                        Note: this will modify :attr:`weight` in-place.
            norm_type (float, optional): The p of the p-norm to compute for the :attr:`max_norm` option. Default ``2``.
            scale_grad_by_freq (bool, optional): If given, this will scale gradients by the inverse of frequency of
                                                    the words in the mini-batch. Default ``False``.
            sparse (bool, optional): If ``True``, gradient w.r.t. :attr:`weight` will be a sparse tensor. See Notes under
                                     :class:`torch.nn.Embedding` for more details regarding sparse gradients.
    
        Shape:
            - Input: LongTensor of arbitrary shape containing the indices to extract
            - Weight: Embedding matrix of floating point type with shape `(V, embedding_dim)`,
              where V = maximum index + 1 and embedding_dim = the embedding size
            - Output: `(*, embedding_dim)`, where `*` is the input shape
    
        Examples::
    
            >>> # a batch of 2 samples of 4 indices each
            >>> input = torch.tensor([[1, 2, 4, 5], [4, 3, 2, 9]])
            >>> # an embedding matrix containing 10 tensors of size 3
            >>> embedding_matrix = torch.rand(10, 3)
            >>> # xdoctest: +IGNORE_WANT("non-deterministic")
            >>> F.embedding(input, embedding_matrix)
            tensor([[[ 0.8490,  0.9625,  0.6753],
                     [ 0.9666,  0.7761,  0.6108],
                     [ 0.6246,  0.9751,  0.3618],
                     [ 0.4161,  0.2419,  0.7383]],
    
                    [[ 0.6246,  0.9751,  0.3618],
                     [ 0.0237,  0.7794,  0.0528],
                     [ 0.9666,  0.7761,  0.6108],
                     [ 0.3385,  0.8612,  0.1867]]])
    
            >>> # example with padding_idx
            >>> weights = torch.rand(10, 3)
            >>> weights[0, :].zero_()
            >>> embedding_matrix = weights
            >>> input = torch.tensor([[0, 2, 0, 5]])
            >>> F.embedding(input, embedding_matrix, padding_idx=0)
            tensor([[[ 0.0000,  0.0000,  0.0000],
                     [ 0.5609,  0.5384,  0.8720],
                     [ 0.0000,  0.0000,  0.0000],
                     [ 0.6262,  0.2438,  0.7471]]])
        """
        if has_torch_function_variadic(input, weight):
            return handle_torch_function(
                embedding,
                (input, weight),
                input,
                weight,
                padding_idx=padding_idx,
                max_norm=max_norm,
                norm_type=norm_type,
                scale_grad_by_freq=scale_grad_by_freq,
                sparse=sparse,
            )
        if padding_idx is not None:
            if padding_idx > 0:
                assert padding_idx < weight.size(
                    0
                ), "Padding_idx must be within num_embeddings"
            elif padding_idx < 0:
                assert padding_idx >= -weight.size(
                    0
                ), "Padding_idx must be within num_embeddings"
                padding_idx = weight.size(0) + padding_idx
        else:
            padding_idx = -1
        if max_norm is not None:
            # Note [embedding_renorm contiguous]
            # `embedding_renorm_` will call .contiguous() on input anyways, so we
            # call it here and take advantage of the improved locality in the
            # `embedding` call below too.
            input = input.contiguous()
            # Note [embedding_renorm set_grad_enabled]
            # XXX: equivalent to
            # with torch.no_grad():
            #   torch.embedding_renorm_
            # remove once script supports set_grad_enabled
            _no_grad_embedding_renorm_(weight, input, max_norm, norm_type)
>       return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       IndexError: index out of range in self

/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages/torch/nn/functional.py:2551: IndexError
```

agobbifbk:
Uh I will look at the code but as far as I see from the error your are passing the `embs = [3,3]` but your data have an higher carnality e.g. I see max value 5. 
The `nn.Embedding` layer expects two input: the first is the maximum carnality in your case `6` and the second one is the dimension of the vector space you are putting that class. In DSIPTS the `embs` variable is something that comes from the data processing pipeline and is not something that the user should give as input. 

If it is not clear please drop me a pm and we try to go through together! 

RN, in PTF I don't think we are processing the categorical variables with any categorical encoder. 

@PranavBhatP I don't remember if there is an open issue related to categorical variables, if yes we should link it because we need to add this feature to the d1 layer metadata eventually. 

I will look at the rest of the PR but at least you know why you are getting this error :-)

agobbifbk:
One question before starting deeply debugging: you take the `Base` DSIPTS class and you adapted it for working with PTF right? 

This seems a nice workaround :-) 
```python
        if "x_num_past" not in batch:
            batch["x_num_past"] = batch["encoder_cont"]
        if "x_num_future" not in batch:
            batch["x_num_future"] = batch["decoder_cont"]
        if "x_cat_past" not in batch:
            batch["x_cat_past"] = batch["encoder_cat"]
        if "x_cat_future" not in batch:
            batch["x_cat_future"] = batch["decoder_cat"]
        if "idx_target" not in batch:
            batch["idx_target"] = [list(range(self.output_channels))]
```
Although I need to check:
1. are you separating `historical_target` data (as @PranavBhatP is doing) or all the past data are in `encoder_cont`? If you have `historical_target` better to use that 
2. `[list(range(self.output_channels))]` need to be checked too 

What about the loss functions? We need to check if the output is suitable for PTF, as you see I'm returning a 4d tensor. Moreover you can work also to let it work with the quantile loss, see how @PranavBhatP is doing it in `TimeXER` [here](https://github.com/sktime/pytorch-forecasting/blob/953214e1d400e5e69d6eb47292e032e13eb34633/pytorch_forecasting/models/timexer/_timexer_v2.py)]

phoeenniixx:
> One question before starting deeply debugging: you take the Base DSIPTS class and you adapted it for working with PTF right?

Yes, that is what I intend to do :)

> In DSIPTS the embs variable is something that comes from the data processing pipeline and is not something that the user should give as input.

Ohh! I did not know that - I think that would solve the problem... Thanks!!

>RN, in PTF I don't think we are processing the categorical variables with any categorical encoder.

Yes that is true, this is still a feature we need to add

phoeenniixx:
> are you separating `historical_target` data (as @PranavBhatP is doing) or all the past data are in `encoder_cont`? If you have `historical_target` better to use that

Yes, rn the `historical_target` in `encoder_cont`, but are we using it in `tide`? what is the key for `historical_target` in `dsipts` or is it even used here in `tide`?
 I tried to map the `EncoderDecoderDataModule`'s keys to `dsipts` as much as I could understand and the model works perfectly fine, jsut the problem was with `embs` which now I understand is that our pipeline is just not compatible with `embs` for now, right?

Also I don't think so `EncoderDecoderDataModule` can still handle `historical_target` - we still need to add it which ig we'll do in next iterations?

> [list(range(self.output_channels))] need to be checked too

from what I could understand from `idx_target` - it is the idx of target? so I thought the best way would be to create a list with values ranging from 0 to `self.output_channels - 1`. In that way all `target`s will have a idx?

> What about the loss functions?

well I took the liberty to change the `compute_loss` to return just the `loss` which is `nn.Module` as that is what our pipeline can handle rn (like if you see `BaseModel` for v2 it uses `nn.Module` losses only), I needed to make these changes jsut to keep it "simple" for now, we can add back these once our pipeline is mature enough? Also there is no `loss_type` param here in `ptf-v2` - so for now I just removed it altogether, we can add a similar functionality once this works as intended or have I done some breaking changes?

```python
    def compute_loss(self, x_dict, y_tensor, y_hat):
        """
        custom loss calculation

        :meta private:
        """

        if self.use_quantiles is False:
            initial_loss = self.loss(y_hat[:, :, :, 0], y_tensor)
        else:
            initial_loss = self.loss(y_hat, y_tensor)
        loss = initial_loss

        return loss
```

> Moreover you can work also to let it work with the quantile loss, see how @PranavBhatP is doing it in TimeXER [here](https://github.com/sktime/pytorch-forecasting/blob/953214e1d400e5e69d6eb47292e032e13eb34633/pytorch_forecasting/models/timexer/_timexer_v2.py)]

Yes I'll try to add the quantile loss and if I face any problem I'll ping you and @PranavBhatP :)

agobbifbk:
> > are you separating `historical_target` data (as @PranavBhatP is doing) or all the past data are in `encoder_cont`? If you have `historical_target` better to use that
> 
> Yes, rn the `historical_target` in `encoder_cont`, but are we using it in `tide`? what is the key for `historical_target` in `dsipts` or is it even used here in `tide`? I tried to map the `EncoderDecoderDataModule`'s keys to `dsipts` as much as I could understand and the model works perfectly fine, jsut the problem was with `embs` which now I understand is that our pipeline is just not compatible with `embs` for now, right?

In DSIPTS we have `x_num_past` that is the concatenation of numerical past and target past, here I see you are using the V1 version of PTF. I look into the code but I can not find where you build `encoder_cont`, can you point it in the d1 layer? We need to be sure that the index are correct, just a double check! 

> 
> Also I don't think so `EncoderDecoderDataModule` can still handle `historical_target` - we still need to add it which ig we'll do in next iterations?

This depends on what we decide in the `TSLIBDataModule`. It may be the case to insert them also here for being aligned or maybe it is too hard to refactor the rest of the code. If we are sure about the indexes (see above) maybe is better to not chage too much here...

> 
> > [list(range(self.output_channels))] need to be checked too
> 
> from what I could understand from `idx_target` - it is the idx of target? so I thought the best way would be to create a list with values ranging from 0 to `self.output_channels - 1`. In that way all `target`s will have a idx?

See above, it seems correct but depends in how the `encoder_cont` is built.

> 
> > What about the loss functions?
> 
> well I took the liberty to change the `compute_loss` to return just the `loss` which is `nn.Module` as that is what our pipeline can handle rn, I needed to make these changes jsut to keep it "simple" for now, we can add back these once our pipeline is mature enough? Also there is no `loss_type` param here in `ptf-v2` - so for now I just removed it altogether, we can add a similar functionality once this works as intended or have I done some breaking changes?

Sure! My comment was about the dimensions. The DSIP module return something that has always the shape (BS,length, channels, quantiles), in your case quantiles is 1 but you need to check how are implemented the loss functions in this version of PTF (are 4d allowed?). I try to read the functions implemented in `metrics` folder but it is not easy to answer to my doubts :-)  In any case since the model  supports the quantile loss I suggest to try to implement also that possibility, what do you think?
> 
> ```python
>     def compute_loss(self, x_dict, y_tensor, y_hat):
>         """
>         custom loss calculation
> 
>         :meta private:
>         """
> 
>         if self.use_quantiles is False:
>             initial_loss = self.loss(y_hat[:, :, :, 0], y_tensor)
>         else:
>             initial_loss = self.loss(y_hat, y_tensor)
>         loss = initial_loss
> 
>         return loss
> ```

phoeenniixx:
> In DSIPTS we have x_num_past that is the concatenation of numerical past and target past, here I see you are using the V1 version of PTF. I look into the code but I can not find where you build encoder_cont, can you point it in the d1 layer? We need to be sure that the index are correct, just a double check!

well this is the `__getitem__` of the nested dataset in data module:https://github.com/sktime/pytorch-forecasting/blob/57a6a1c77b81cba755a19199e6ecd3a7d8de296d/pytorch_forecasting/data/data_module.py#L382-L552
please see if this helps

> Sure! My comment was about the dimensions. The DSIP module return something that has always the shape (BS,length, channels, quantiles), in your case quantiles is 1 but you need to check how are implemented the loss functions in this version of PTF (are 4d allowed?). I try to read the functions implemented in metrics folder but it is not easy to answer to my doubts :-) In any case since the model supports the quantile loss I suggest to try to implement also that possibility, what do you think?

Well rn we are using `torch.nn` losses and not `metrices` in v2, but yes we intend to add metrices of v1 to v2 in near future, so I think for now it should work, about the shape, I changed the `compute_loss` a little in `Base`, there it actually uses the loss function of  `torch.nn` only with which v2 pipeline is compatible... Or do you mean something else here and I misunderstood your statement?

agobbifbk:
Not fully because ` encoder_cont = data["features"]`  point to `features` that I think links to `features = sample["x"]` that links to `x` that seems to come from this ` "x": torch.tensor(data_feat_vals),` and `data_feat_vals` comes from `feature_cols`
that are defined as:
```python
self.feature_cols = [
            col
            for col in data.columns
            if col not in [self.time] + self._group + [self.weight] + self._target
        ]
```
so my idea is that you can not say that the first N are related to the target, but I may have understand it wrong :-)

phoeenniixx:
Hmm I think I understand this is a difference between assumptions of the 2 packages? We may need to have a discussion on this..  I still need your help there :)

agobbifbk:
Sure!

phoeenniixx:
Hi @agobbifbk, I was trying to debug the `forward` and and i saw that in 
```python
        if self.aux_past_channels > 0:  # if we have more numerical variables about past
            aux_num_past = batch["encoder_cont"]
            assert self.aux_past_channels == aux_num_past.size(2), beauty_string(
                f"{self.aux_past_channels} LAYERS FOR PAST VARS AND "
                f"{aux_num_past.size(2)} VARS",
                "section",
                True,
            )
```
`self.aux_past_channels` is calculated as 
```python
self.aux_past_channels = self.past_channels - self.output_channels
```
but in `ptf` we already have `target` or `y` separate and we donot keep it in `feature_cols`
```python
self.feature_cols = [
            col
            for col in data.columns
            if col not in [self.time] + self._group + [self.weight] + self._target
        ]
```
so I think in `ptf` we should keep `self.aux_past_channels` same as `self.past_channels` or in other words it should be 
```python 
self.aux_past_channels = self.past_channels
```
and not 
```python
self.aux_past_channels = self.past_channels - self.output_channels
```
also as `tide` uses `target_past`, the only way to correctly calculate [`y_past`](https://github.com/sktime/pytorch-forecasting/blob/84367930615fad7cd14d10df96607601707e241f/pytorch_forecasting/models/tide/tide_dsipts/tide.py#L182) in `forward` is by calculating `target_past` in datamodule. I do this in this way (taking inspiration from @PranavBhatP's implementation) :
```python
target_past = data["target"][encoder_indices]
```
where `data["target"]` contains the `target` from d1 layer. So, it will return `y` just before `y_present` is. 
The basic logic is already there in data module [[source]](https://github.com/sktime/pytorch-forecasting/blob/1fc5542a6639f9d90e960392e3845a0be875847f/pytorch_forecasting/data/data_module.py#L433) used to calculate `target_scale`, we just need to add it to `collate_fn`. We can change the name to `target_pass` there...


So the new `forward` will look like this:
```python
    def forward(self, X: dict) -> float:
        if isinstance(X, tuple):
            x_batch, y_batch = X
            batch = x_batch
        else:
            batch = X

        if "x_num_past" not in batch:
            batch["x_num_past"] = batch["encoder_cont"]
        if "x_num_future" not in batch:
            batch["x_num_future"] = batch["decoder_cont"]
        if "x_cat_past" not in batch:
            batch["x_cat_past"] = batch["encoder_cat"]
        if "x_cat_future" not in batch:
            batch["x_cat_future"] = batch["decoder_cat"]

        y_past = batch["target_past"] # comes from D2 layer
        B = y_past.shape[0]

        emb_cat_past, emb_cat_fut = self.cat_categorical_vars(batch)

        emb_cat_past = torch.mean(emb_cat_past, dim=2)
        emb_cat_fut = torch.mean(emb_cat_fut, dim=2)

        if self.aux_past_channels > 0: 
            aux_num_past = batch["encoder_cont"]
            assert self.aux_past_channels == aux_num_past.size(2), beauty_string(
                f"{self.aux_past_channels} LAYERS FOR PAST VARS AND "
                f"{aux_num_past.size(2)} VARS",
                "section",
                True,
            )
            # -----------------------------------------------------------
            # if we never do
            # self.aux_past_channels = self.past_channels - self.output_channels and
            # rather jsut do self.aux_past_channels = self.past_channels, it passes,
            # is it because ptf donot have target in features_col? 
            # -----------------------------------------------------------
            aux_emb_num_past = torch.Tensor().to(self.device)
            for i, layer in enumerate(self.linear_aux_past):
                aux_emb_past = layer(aux_num_past[:, :, [i]]).unsqueeze(2)
                aux_emb_num_past = torch.cat((aux_emb_num_past, aux_emb_past), dim=2)
            aux_emb_num_past = torch.mean(aux_emb_num_past, dim=2)
        else:
            aux_emb_num_past = None  

        if (
            self.aux_fut_channels > 0
        ): 
            aux_num_fut = batch["x_num_future"].to(self.device)
            assert self.aux_fut_channels == aux_num_fut.size(2), beauty_string(
                f"{self.aux_fut_channels} LAYERS FOR PAST VARS AND "
                f"{aux_num_fut.size(2)} VARS",
                "section",
                True,
            )  
            aux_emb_num_fut = torch.Tensor().to(self.device)
            for j, layer in enumerate(self.linear_aux_fut):
                aux_emb_fut = layer(aux_num_fut[:, :, [j]]).unsqueeze(2)
                aux_emb_num_fut = torch.cat((aux_emb_num_fut, aux_emb_fut), dim=2)
            aux_emb_num_fut = torch.mean(aux_emb_num_fut, dim=2)
        else:
            aux_emb_num_fut = None  

        if self.aux_past_channels > 0:
            emb_past = torch.cat(
                (emb_cat_past, aux_emb_num_past), dim=2
            )  
            proj_past = self.feat_proj_past(emb_past, True)
        else:
            proj_past = self.feat_proj_past(emb_cat_past, True)

        if self.aux_fut_channels > 0:
            emb_fut = torch.cat((emb_cat_fut, aux_emb_num_fut), dim=2)
            proj_fut = self.feat_proj_fut(emb_fut, True)  
        else:
            proj_fut = self.feat_proj_fut(emb_cat_fut, True) 

        concat = torch.cat(
            (y_past.view(B, -1), proj_past.view(B, -1), proj_fut.view(B, -1)), dim=1
        )  
        dense_enc = self.first_encoder(concat)
        for lay_enc in self.aux_encoder:
            dense_enc = lay_enc(dense_enc)

        dense_dec = self.first_decoder(dense_enc)
        for lay_dec in self.aux_decoder:
            dense_dec = lay_dec(dense_dec)

        temp_dec_input = torch.cat(
            (dense_dec.view(B, self.future_steps, self.d_model), proj_fut), dim=2
        )
        temp_dec_output = self.temporal_decoder(temp_dec_input, False)
        temp_dec_output = temp_dec_output.view(
            B, self.future_steps, self.output_channels, self.mul
        )

        linear_regr = self.linear_target(y_past.view(B, -1))
        linear_output = linear_regr.view(
            B, self.future_steps, self.output_channels, self.mul
        )

        output = temp_dec_output + linear_output
        return output

```


Am I on the right track? I think this will solve the problem of "wrong calculations of target" you noticed ?

phoeenniixx:
The only issue with this approach is we need to add `target_past` to datamodule, that I think we can do now, as first e2e is already out and now we can work on improving the datamodule?

I think we can discuss more on this on Monday?

fkiraly (changes requested):
Can you kindly:

* move the base class to a private omdule in `base`?
* the neural networks to `layers`?

It also seems you did not stack properly and duplicated commits instead. This made it very difficult to merge, but I fixed it. Please be aware for next time.

agobbifbk:
> Hi @agobbifbk, I was trying to debug the `forward` and and i saw that in
> 
> ```python
>         if self.aux_past_channels > 0:  # if we have more numerical variables about past
>             aux_num_past = batch["encoder_cont"]
>             assert self.aux_past_channels == aux_num_past.size(2), beauty_string(
>                 f"{self.aux_past_channels} LAYERS FOR PAST VARS AND "
>                 f"{aux_num_past.size(2)} VARS",
>                 "section",
>                 True,
>             )
> ```
> 
> `self.aux_past_channels` is calculated as
> 
> ```python
> self.aux_past_channels = self.past_channels - self.output_channels
> ```
> 
> but in `ptf` we already have `target` or `y` separate and we donot keep it in `feature_cols`
> 
> ```python
> self.feature_cols = [
>             col
>             for col in data.columns
>             if col not in [self.time] + self._group + [self.weight] + self._target
>         ]
> ```
> 
> so I think in `ptf` we should keep `self.aux_past_channels` same as `self.past_channels` or in other words it should be
> 
> ```python
> self.aux_past_channels = self.past_channels
> ```
> 
> and not
> 
> ```python
> self.aux_past_channels = self.past_channels - self.output_channels
> ```
> 

Correct!

> also as `tide` uses `target_past`, the only way to correctly calculate [`y_past`](https://github.com/sktime/pytorch-forecasting/blob/84367930615fad7cd14d10df96607601707e241f/pytorch_forecasting/models/tide/tide_dsipts/tide.py#L182) in `forward` is by calculating `target_past` in datamodule. I do this in this way (taking inspiration from @PranavBhatP's implementation) :
> 
> ```python
> target_past = data["target"][encoder_indices]
> ```
> 
> where `data["target"]` contains the `target` from d1 layer. So, it will return `y` just before `y_present` is. The basic logic is already there in data module [[source]](https://github.com/sktime/pytorch-forecasting/blob/1fc5542a6639f9d90e960392e3845a0be875847f/pytorch_forecasting/data/data_module.py#L433) used to calculate `target_scale`, we just need to add it to `collate_fn`. We can change the name to `target_pass` there...
> 

Yes exactly, you are converging to the solution we have in the tslib d2 layer more or less

> So the new `forward` will look like this:
> 
> ```python
>     def forward(self, X: dict) -> float:
>         if isinstance(X, tuple):
>             x_batch, y_batch = X
>             batch = x_batch
>         else:
>             batch = X
> 
>         if "x_num_past" not in batch:
>             batch["x_num_past"] = batch["encoder_cont"]
>         if "x_num_future" not in batch:
>             batch["x_num_future"] = batch["decoder_cont"]
>         if "x_cat_past" not in batch:
>             batch["x_cat_past"] = batch["encoder_cat"]
>         if "x_cat_future" not in batch:
>             batch["x_cat_future"] = batch["decoder_cat"]
> 
>         y_past = batch["target_past"] # comes from D2 layer
>         B = y_past.shape[0]
> 
>         emb_cat_past, emb_cat_fut = self.cat_categorical_vars(batch)
> 
>         emb_cat_past = torch.mean(emb_cat_past, dim=2)
>         emb_cat_fut = torch.mean(emb_cat_fut, dim=2)
> 
>         if self.aux_past_channels > 0: 
>             aux_num_past = batch["encoder_cont"]
>             assert self.aux_past_channels == aux_num_past.size(2), beauty_string(
>                 f"{self.aux_past_channels} LAYERS FOR PAST VARS AND "
>                 f"{aux_num_past.size(2)} VARS",
>                 "section",
>                 True,
>             )
>             # -----------------------------------------------------------
>             # if we never do
>             # self.aux_past_channels = self.past_channels - self.output_channels and
>             # rather jsut do self.aux_past_channels = self.past_channels, it passes,
>             # is it because ptf donot have target in features_col? 
>             # -----------------------------------------------------------
>             aux_emb_num_past = torch.Tensor().to(self.device)
>             for i, layer in enumerate(self.linear_aux_past):
>                 aux_emb_past = layer(aux_num_past[:, :, [i]]).unsqueeze(2)
>                 aux_emb_num_past = torch.cat((aux_emb_num_past, aux_emb_past), dim=2)
>             aux_emb_num_past = torch.mean(aux_emb_num_past, dim=2)
>         else:
>             aux_emb_num_past = None  
> 
>         if (
>             self.aux_fut_channels > 0
>         ): 
>             aux_num_fut = batch["x_num_future"].to(self.device)
>             assert self.aux_fut_channels == aux_num_fut.size(2), beauty_string(
>                 f"{self.aux_fut_channels} LAYERS FOR PAST VARS AND "
>                 f"{aux_num_fut.size(2)} VARS",
>                 "section",
>                 True,
>             )  
>             aux_emb_num_fut = torch.Tensor().to(self.device)
>             for j, layer in enumerate(self.linear_aux_fut):
>                 aux_emb_fut = layer(aux_num_fut[:, :, [j]]).unsqueeze(2)
>                 aux_emb_num_fut = torch.cat((aux_emb_num_fut, aux_emb_fut), dim=2)
>             aux_emb_num_fut = torch.mean(aux_emb_num_fut, dim=2)
>         else:
>             aux_emb_num_fut = None  
> 
>         if self.aux_past_channels > 0:
>             emb_past = torch.cat(
>                 (emb_cat_past, aux_emb_num_past), dim=2
>             )  
>             proj_past = self.feat_proj_past(emb_past, True)
>         else:
>             proj_past = self.feat_proj_past(emb_cat_past, True)
> 
>         if self.aux_fut_channels > 0:
>             emb_fut = torch.cat((emb_cat_fut, aux_emb_num_fut), dim=2)
>             proj_fut = self.feat_proj_fut(emb_fut, True)  
>         else:
>             proj_fut = self.feat_proj_fut(emb_cat_fut, True) 
> 
>         concat = torch.cat(
>             (y_past.view(B, -1), proj_past.view(B, -1), proj_fut.view(B, -1)), dim=1
>         )  
>         dense_enc = self.first_encoder(concat)
>         for lay_enc in self.aux_encoder:
>             dense_enc = lay_enc(dense_enc)
> 
>         dense_dec = self.first_decoder(dense_enc)
>         for lay_dec in self.aux_decoder:
>             dense_dec = lay_dec(dense_dec)
> 
>         temp_dec_input = torch.cat(
>             (dense_dec.view(B, self.future_steps, self.d_model), proj_fut), dim=2
>         )
>         temp_dec_output = self.temporal_decoder(temp_dec_input, False)
>         temp_dec_output = temp_dec_output.view(
>             B, self.future_steps, self.output_channels, self.mul
>         )
> 
>         linear_regr = self.linear_target(y_past.view(B, -1))
>         linear_output = linear_regr.view(
>             B, self.future_steps, self.output_channels, self.mul
>         )
> 
>         output = temp_dec_output + linear_output
>         return output
> ```
> 
> Am I on the right track? I think this will solve the problem of "wrong calculations of target" you noticed ?

It seems correct, unluckily you picked one of the model not written by me (but from one of my colleagues) so it is difficult to understand the rest of the code but I think you used the correct information at each step. There is one think to say: in DSIPTS we assume that `cat_past` and` cat_fut` variables are the same. Is it the case in PTF? We would like to change this for being more flexible but it can take a while since it is impacting a huge amount of lines of code :-) If in PTF they can be different we need to change something in `self.cat_categorical_vars(batch)`. 
Last question: the `embs` is an input required but AFAIK the values are not computed from the d1 layer RN, how are you managing them?

phoeenniixx:
>  There is one think to say: in DSIPTS we assume that cat_past and cat_fut variables are the same. Is it the case in PTF? 

`cat_past` is `encoder_cat` and `cat_fut` is `decoder_cat` I assume, in that case, they are not the same :
* `encoder_cat` : Uses ALL categorical features (both known and unknown)
* `decoder_cat` : Uses ONLY the categorical features marked as "known" 
See lines : 
https://github.com/sktime/pytorch-forecasting/blob/300fd90db71df492b3035d783216b2bc4a95e8e4/pytorch_forecasting/data/data_module.py#L449-L489

> If in PTF they can be different we need to change something in self.cat_categorical_vars(batch)

Ok, so we might need to make changes to this method as well? I will look into it!

> the `embs` is an input required but AFAIK the values are not computed from the d1 layer RN, how are you managing them?

I have for now made it default to `[]` (empty list), we can change it once D1/D2 layer are fully mature.

phoeenniixx:
Although, I might need some help with `cat_categorical_vars`, I am not very much familiar with the model and I might need some guidance :)

agobbifbk:
> > There is one think to say: in DSIPTS we assume that cat_past and cat_fut variables are the same. Is it the case in PTF?
> 
> `cat_past` is `encoder_cat` and `cat_fut` is `decoder_cat` I assume, in that case, they are not the same :
> 
> * `encoder_cat` : Uses ALL categorical features (both known and unknown)
> * `decoder_cat` : Uses ONLY the categorical features marked as "known"
>   See lines :
>   https://github.com/sktime/pytorch-forecasting/blob/300fd90db71df492b3035d783216b2bc4a95e8e4/pytorch_forecasting/data/data_module.py#L449-L489
> 
> > If in PTF they can be different we need to change something in self.cat_categorical_vars(batch)
> 
> Ok, so we might need to make changes to this method as well? I will look into it!

Yes in this case there will be something like `embs_past` and `embs_future` and the corresponding tensors should be processed by two different instance of `embedding_cat_variables`: one with the data of the past and the other with data of the future (length and embs). Or we can think to a unique module that has 2 instances of the embedding layers, something like:
```python
  self.cat_n_embd_past = nn.ModuleList([nn.Embedding(emb_dim, d_model) for emb_dim in self.cat_embeds_past] )
  self.cat_n_embd_fut = nn.ModuleList([nn.Embedding(emb_dim, d_model) for emb_dim in self.cat_embeds_future] )
```
And modifying the forward loop according 

> 
> > the `embs` is an input required but AFAIK the values are not computed from the d1 layer RN, how are you managing them?
> 
> I have for now made it default to `[]` (empty list), we can change it once D1/D2 layer are fully mature.

For now you are just using something like positional encoder for past and future. We can work on it together when we have a mature design of D1/D2 then! We will do once and reuse it among all the other models (and DSIPTS) :-)

phoeenniixx:
@agobbifbk, @fkiraly, I think this PR is ready for review ( I tried to keep the code almost same as [tide in dsipts](https://github.com/DSIP-FBK/DSIPTS_PTF/blob/main/dsipts/models/TIDE.py))

phoeenniixx:
(still need to add docstrings, the current docstrings are also copied from dsipts, I will add them today as well)

fkiraly:
> I think this PR is ready for review 

then please click the button "ready for review" (below), I almost did not see this

agobbifbk:
https://github.com/sktime/pytorch-forecasting/blob/89939e8b363914a8631ff8c5dd2b889b3a0557cd/pytorch_forecasting/models/tide/tide_dsipts/_tide_v2.py#L165

Here I have some doubts: are you sure all the keys are in the batch dict: `encoder_cont,decoder_cont,encoder_cat,decoder_cat`? I see in the D2 that you put some defaults if this information is missing (see https://github.com/sktime/pytorch-forecasting/blob/89939e8b363914a8631ff8c5dd2b889b3a0557cd/pytorch_forecasting/data/data_module.py#L491) 

This is has not been tested, for example here (https://github.com/sktime/pytorch-forecasting/blob/89939e8b363914a8631ff8c5dd2b889b3a0557cd/pytorch_forecasting/layers/_dsipts/_sub_nn.py#L58) I don't know what could be the behavior of this operation in case of a 0-shaped tensor. Are your tests finishing without any error?

phoeenniixx:
> Are your tests finishing without any error?

Yes,
From what I can understand, `embedding_cat_variables` is to get embeddings for categorical variables, but as we discussed, we are for now, not passing any categoricals, so I think this layer is not used, that's why there is no issue?

The data i am using, doesnot pass any categoricals:
https://github.com/sktime/pytorch-forecasting/blob/d931c3b34e7ab9b4da2d9d80eacc074b504bc2f3/pytorch_forecasting/tests/_data_scenarios.py#L189

agobbifbk:
Ok, I see. Even though there is the key in the dict there is no category to loop on. It should work :-)

phoeenniixx:
exactly :)

pytorch_forecasting/data/data_module.py
```diff
@@ -430,8 +430,8 @@ def __getitem__(self, idx):
             encoder_indices = slice(start_idx, start_idx + enc_length)
             decoder_indices = slice(start_idx + enc_length, end_idx)
 
-            target_scale = data["target"][encoder_indices]
-            target_scale = target_scale[~torch.isnan(target_scale)].abs().mean()
+            target_past = data["target"][encoder_indices]
```
fkiraly, line 433: why are we doing this? This seems like a significant change to the internal contract of `TimeSeriesDataSet`

fkiraly (changes requested):
* see above, it seems like there is a significant change to `EncoderDecoderTimeSeriesDataModule` in terms of the dict return. The `target_scale` variable might not be assigned, and its source is being changed. Can you please explain?
* Can you add the new layers in folders sorted by type of layer rather than by name of dependency (dsipts)?

pytorch_forecasting/data/data_module.py
phoeenniixx: We discussed it some months back, that there was no `target_past` in `EncoderDecoderDataModule` and some models do require it. `TSLibDataModule` already implements it as `history_target` (see [here](https://github.com/sktime/pytorch-forecasting/blob/1956bc82fde4a5585860918fcab1bf0b463a7206/pytorch_forecasting/data/_tslib_data_module.py#L156)). `Tide` also needs it, and we were already calculating `target_past` for the `target_scale`, so I just renamed the variable and added it to the return

(if you see line 509, we are still returning `target_scale` as well)

fkiraly: ok, I suppose this happens only in D2, so that does not compromise the original design, since D2 could return anything

phoeenniixx: exactly :)

PR #1897, opened 2025-06-19
[ENH] Refactor metrics to be polymorphic

phoeenniixx:
This PR refactors `base_metrics` to make them polymorphic.
Mainly Base class `Metric` is changed as the changes can be easily propagated to the child classes. Moreover other cases like `MultiHorizonMetric`, `MultiLoss` etc are also changed

### DIFF
--- a/pytorch_forecasting/metrics/base_metrics.py
+++ b/pytorch_forecasting/metrics/base_metrics.py
@@ -3,7 +3,7 @@ Base classes for metrics - only for inheritance.
 """
 
 import inspect
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Union
 import warnings
 
 from sklearn.base import BaseEstimator
@@ -87,7 +87,7 @@ class Metric(LightningMetric):
         """  # noqa: E501
         return encoder(dict(prediction=parameters, target_scale=target_scale))
 
-    def to_prediction(self, y_pred: torch.Tensor) -> torch.Tensor:
+    def _to_prediction_3d(self, y_pred: torch.Tensor) -> torch.Tensor:
         """
         Convert network prediction into a point prediction.
 
@@ -107,7 +107,28 @@ class Metric(LightningMetric):
                 y_pred = y_pred.mean(-1)
         return y_pred
 
-    def to_quantiles(
+    def to_prediction(self, y_pred: torch.Tensor) -> torch.Tensor:
+        """
+        Convert network prediction into a point prediction.
+
+        Args:
+            y_pred: prediction output of network.
+
+        Returns:
+            torch.Tensor: point prediction.
+        """
+        if y_pred.ndim == 4:
+            predictions = [
+                self._to_prediction_3d(y_pred[:, :, i, :])
+                for i in range(y_pred.shape[2])
+            ]
+            return torch.stack(predictions, dim=2)
+        elif y_pred.ndim == 3:
+            return self._to_prediction_3d(y_pred)
+        else:
+            return y_pred
+
+    def _to_quantiles_3d(
         self, y_pred: torch.Tensor, quantiles: list[float] = None
     ) -> torch.Tensor:
         """
@@ -136,6 +157,35 @@ class Metric(LightningMetric):
         else:
             raise ValueError(
                 f"prediction has 1 or more than 3 dimensions: {y_pred.ndim}"
+            )
+
+    def to_quantiles(
+        self, y_pred: torch.Tensor, quantiles: list[float] = None
+    ) -> torch.Tensor:
+        """
+        Convert network prediction into a quantile prediction.
+
+        Args:
+            y_pred: prediction output of network
+            quantiles (List[float], optional): quantiles for probability range. Defaults to quantiles as
+                as defined in the class initialization.
+
+        Returns:
+            torch.Tensor: prediction quantiles.
+        """  # noqa: E501
+        if y_pred.ndim == 4:
+            quantile_preds = [
+                self._to_quantiles_3d(y_pred[:, :, i, :], quantiles=quantiles)
+                for i in range(y_pred.shape[2])
+            ]
+            return torch.stack(quantile_preds, dim=2)
+        elif y_pred.ndim == 3:
+            return self._to_quantiles_3d(y_pred, quantiles=quantiles)
+        elif y_pred.ndim == 2:
+            return self._to_quantiles_3d(y_pred, quantiles=quantiles)
+        else:
+            raise ValueError(
+                f"prediction has unsupported number of dimensions: {y_pred.ndim}"
             )
 
     def __add__(self, metric: LightningMetric):
@@ -320,6 +370,35 @@ class MultiLoss(LightningMetric):
         """
         return len(self.metrics)
 
+    def _prepare_y_pred(
+        self, y_pred: Union[torch.Tensor, list[torch.Tensor]]
+    ) -> list[torch.Tensor]:
+        """
+        Ensure y_pred is a list of tensors.
+        """
+        if isinstance(y_pred, torch.Tensor) and y_pred.ndim == 4:
+            if y_pred.shape[2] != len(self.metrics):
+                raise ValueError(
+                    f"The number of targets in the 4D prediction "
+                    f"tensor ({y_pred.shape[2]}) "
+                    f"does not match the number of metrics in "
+                    f"MultiLoss ({len(self.metrics)})."
+                )
+            return list(torch.unbind(y_pred, dim=2))
+        elif isinstance(y_pred, (list, tuple)):
+            if len(y_pred) != len(self.metrics):
+                raise ValueError(
+                    f"The number of predictions in the list ({len(y_pred)}) "
+                    f"does not match the number of metrics in "
+                    f"MultiLoss ({len(self.metrics)})."
+                )
+            return y_pred
+        else:
+            raise TypeError(
+                "y_pred for MultiLoss must be a list of tensors or a single 4D tensor, "
+                f"but got {type(y_pred)}"
+            )
+
     def update(self, y_pred: torch.Tensor, y_actual: torch.Tensor, **kwargs) -> None:
         """
         Update composite metric
@@ -329,6 +408,7 @@ class MultiLoss(LightningMetric):
             y_actual: actual values
             **kwargs: arguments to update function
         """
+        y_pred = self._prepare_y_pred(y_pred)
         for idx, metric in enumerate(self.metrics):
             try:
                 metric.update(
@@ -372,6 +452,7 @@ class MultiLoss(LightningMetric):
         Returns:
             torch.Tensor: metric value on which backpropagation can be applied
         """
+        y_pred = self._prepare_y_pred(y_pred)
         results = []
         for idx, metric in enumerate(self.metrics):
             try:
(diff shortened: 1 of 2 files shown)


phoeenniixx:
FYI @agobbifbk, @fkiraly , @PranavBhatP

agobbifbk:
Very hard to debug, a lot of overhead for just computing the loss :-)  @fkiraly did you write it or do you know the author? Not sure to be able to say something useful here, sorry.

fkiraly (changes requested):
I am deeply uncomfortable with all the case distinctions here.
This is very difficult to parse.

What we need to work with:

* proper docstrings. Do not just say the output is a `torch.Tensor`, but be clear about dimension, size, and type. To not say the output is `y_pred` without any type etc. Same everywhere
* ensure that is also tested reliably. Test all changed and newly introduced methods with individual input/output pairs.

If we do not do this, we risk introducing a lot of bugs and brittleness. The code would be impossible to maintain in the current state.

Minor comment: It is easier to document clearly if you stick to the numpydoc format for docstring that we use in sktime.

phoeenniixx:
> proper docstrings.

I still need to add docstrings, for now, I just copied the docstrings of original methods, I will update them with numpydoc format and add clear dimensions

> ensure that is also tested reliably.

I tested the metrics using [test_metrics](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_metrics.py), and all the tests passed, but `test_all_estimators` showed a bug in `DistributionLoss` functions so, I will work on it and will add tests for 4D inputs as well

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -107,7 +107,28 @@ def to_prediction(self, y_pred: torch.Tensor) -> torch.Tensor:
                 y_pred = y_pred.mean(-1)
         return y_pred
 
-    def to_quantiles(
+    def to_prediction(self, y_pred: torch.Tensor) -> torch.Tensor:
+        """
+        Convert network prediction into a point prediction.
+
+        Args:
```
fkiraly: this docstring should be clarified - assumed input type, guaranteed output type

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -87,7 +87,7 @@ def rescale_parameters(
         """  # noqa: E501
         return encoder(dict(prediction=parameters, target_scale=target_scale))
 
-    def to_prediction(self, y_pred: torch.Tensor) -> torch.Tensor:
+    def _to_prediction_3d(self, y_pred: torch.Tensor) -> torch.Tensor:
```
fkiraly, line 90: this docstring should be clarified - assumed input type, guaranteed output type

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -138,6 +159,35 @@ def to_quantiles(
                 f"prediction has 1 or more than 3 dimensions: {y_pred.ndim}"
             )
 
+    def to_quantiles(
+        self, y_pred: torch.Tensor, quantiles: list[float] = None
+    ) -> torch.Tensor:
+        """
```
fkiraly: this docstring should be clarified - assumed input type, guaranteed output type

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -320,6 +370,35 @@ def __len__(self) -> int:
         """
         return len(self.metrics)
 
+    def _prepare_y_pred(
```
fkiraly, line 429: this docstring should be clarified - assumed input type, guaranteed output type

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -1046,7 +1158,9 @@ def loss(self, y_pred: torch.Tensor, y_actual: torch.Tensor) -> torch.Tensor:
         loss = -distribution.log_prob(y_actual)
         return loss
 
-    def to_prediction(self, y_pred: torch.Tensor, n_samples: int = 100) -> torch.Tensor:
+    def _to_prediction_3d(
+        self, y_pred: torch.Tensor, n_samples: int = 100
+    ) -> torch.Tensor:
```
fkiraly, line 1280: this docstring should be clarified - assumed input type, guaranteed output type

fkiraly:
> I still need to add docstrings, for now, I just copied the docstrings of original methods, I will update them with numpydoc format and add clear dimensions

I understand that this is already how it was - which is part of the reason, perhaps, that the code base is a bit confusing.

Because we are now changing this - and working in a group of people - I think we need to be super precise as to what the contracts are.

We are in a really central and crucial location of the package API, I think only clean documentation (from the start) and systematic testing ensures that we do not introduce problems that are hard to spot later.

phoeenniixx:
Hi, I tried to debug the errors, I was able to do so for most part but for these params in `_deepar_pkg`
https://github.com/sktime/pytorch-forecasting/blob/33041c74f5598be629cf02dbca87da2195fd622e/pytorch_forecasting/models/deepar/_deepar_pkg.py#L93-L99

I am a little clueless :
- for `ImplicitQuantileNetworkDistributionLoss`, just adding `input_size` = 98 in param works that means
```python
dict(
                loss=ImplicitQuantileNetworkDistributionLoss(hidden_size=8, input_size=98),
            ),
```
but what I am more concerned about is why this is not required in the current implementation? 
the error it shows is:
```
  emb_inputs = x.unsqueeze(-2) * (
            1.0 + cos_emb_tau
        )  # ... x n_quantiles x input_size
E       RuntimeError: The size of tensor a (98) must match the size of tensor b (16) at non-singleton dimension 3
```
it is from this line :
https://github.com/sktime/pytorch-forecasting/blob/33041c74f5598be629cf02dbca87da2195fd622e/pytorch_forecasting/metrics/distributions.py#L543-L547

at first i thought this is because the default value of `input_size` is 16 and that is causing the issue, and i changed it to 98 and it worked! but the problem is why this was not raised in the current implementation, also I tried using a debugger to see when this shape problem occurs, but i never found the exact point where it fails as in the debugger, it always shows `x` to be of shape [a,b,16] (before unsqueeze) which means there should never be a requirement of 98?

**Edit**: I ran the debugger again but now with more brakpoints, i found that at first (for most of the `y_pred`, the shape is correct of x - [a,b,16], but I dont know why, suddenly after many `y_pred`s it changes to [a,b,98].

- for `MultivariateNormalDistributionLoss`:
the tests pass if we use `target_normalizer` (which is the next param):
```python
 dict(
                loss=MultivariateNormalDistributionLoss(),
                data_loader_kwargs=dict(
                    target_normalizer=GroupNormalizer(
                        groups=["agency", "sku"], transformation="log1p"
                    )
                ),
                trainer_kwargs=dict(accelerator="cpu"),
            ),
```
I am not sure why it is not working without normalisation

this is the error:
```
        K.view(-1, m * m)[:, :: m + 1] += 1  # add identity matrix to K
>       return torch.linalg.cholesky(K)
E       torch._C._LinAlgError: linalg.cholesky: (Batch element 1): The factorization could not be completed because the input is not positive-definite (the leading minor of order 1 is not positive-definite).
```
so the normalisation makes it compatible, i understand that, but why not without it? and how my reshaping the inputs affecting things to become "not positive-definite"

phoeenniixx:
I have added the docstrings, please see if that helps or if I need to makethem more descriptive, 
Also please see if my changes are breaking the code ( rn I am a little confused with what is happening in these two params as mentioned above)

I think I should start adding tests, maybe that can help as well... so for now I will work on tests

PranavBhatP:
> so the normalisation makes it compatible, i understand that, but why not without it? and how my reshaping the inputs affecting things to become "not positive-definite"

I ran a quick search through co-pilot, it says that this kind of loss always requires normalization due to its existing implementation. There is no work-around to not using normalization when using this specific loss. Ig we should stick to this, not sure about the mathematical reason behind this.

phoeenniixx:
> > so the normalisation makes it compatible, i understand that, but why not without it? and how my reshaping the inputs affecting things to become "not positive-definite"
> 
> I ran a quick search through co-pilot, it says that this kind of loss always requires normalization due to its existing implementation. There is no work-around to not using normalization when using this specific loss. Ig we should stick to this, not sure about the mathematical reason behind this.

but then why it never failed before? I mean this param was added without normalisation right? and it passed at that time? then why now it fails?

fkiraly:
> Hi, I tried to debug the errors

FYI, unfortunately I do not know why these parameters were chosen as they were - I just transferred them from the pre-existing tests in 1.2.0 and prior, from the `tests` folder in root. I did notice that the test parameters were "brittle" in the sense that varying them may cause hard to diagnose failures, but I did not drill down on that.

fkiraly:
maybe sth to discuss in a tech meeting - to give any useful input, I would need to understand the exact context of what is failing and when.

pytorch_forecasting/metrics/base_metrics.py
```diff
@@ -87,15 +87,30 @@ def rescale_parameters(
...
+    def _to_prediction_3d(self, y_pred: torch.Tensor) -> torch.Tensor:
+        """Convert network prediction into a point prediction.
 
-        Args:
-            y_pred: prediction output of network
+        This is an internal helper method.
 
-        Returns:
-            torch.Tensor: point prediction
+        Parameters
```
fkiraly, line 95: good docstring!

Regarding the API: is this "as current" or "as reworked"?

* how do we ensure downwards compatibility with v1
* did we not decide on `(batch, variable, time, params)`?

This is as current, right?

fkiraly (changes requested):
In the PR description, can you please be more detailed of what you are doing, please?

* what is the state prior to the PR
* how are you changing it

Please be precise.

pytorch_forecasting/metrics/base_metrics.py
phoeenniixx: So this is a private method (`_to_prediction_3D`) - it contains the original logic of `to_prediction` that handles 2D and 3D inputs. If you look at the next function (`to_prediction`) , it contains the logic to handle the `(batch, time, num_target, params)`.
The new `to_prediction` can handle 2D, 3D inputs (like the original implementation) as well as the 4D input.

For 2D, 3D inputs, `to_prediction` uses this `_to_prediction_3D` and for 4D it passes the 3D slices of `[batch, time, params]` for each `target` in `num_target`  to `_to_prediction_3D` then stacks it up...

phoeenniixx:
> Also, a "refactor" means that public APIs remain the same - I guess this is not the case here? Or are you widening the contract?

So as discussed on discord, we decided to make the metrics polymorphic, i.e, it can handle any dimension. So if this is the case, then even the original model implementations (that has 3D outputs) are also "ideally" compatible with metrics - This is what I am trying to do. So if even the original model implementations are compatible with metrics, we are just adding the support for 4D model outputs, I thought word "refactor" would explain it well.  Am I messing things up here?

pytorch_forecasting/metrics/base_metrics.py
fkiraly: is this "as current" or "as reworked"?

fkiraly:
> Am I messing things up here?

Not sure - what I am missing is a clear understanding of the status quo design, and a clear description of the target design, both  including signatures of functions with expected input and output types, assumes and guarantees.

I would strongly recommend to write a design proposal or we might end up talking cross purpose.

For terminology:

* a "refactor" typically refers to changes in private layers that leave the logic and the public interfaces unchanged.
* a "rework" can include changes to the API, or changes to the logic.
* "widening the contract" means that a function or API accepts more inputs than previously, but inputs that were previously accepted are still accepted and lead to the same behaviour.

phoeenniixx:
Hi @agobbifbk @fkiraly @PranavBhatP, should we close this? As I think we decided not to make any changes to `metrics` and we dont need to change any contract wrt to `metrics`?

PR #1898, opened 2025-06-22
WIP 

Vishnu-Rangiah:
wip


jobs-git:
I have observed a memory leak in sktim/pytorch-forecasting too, it reached even more than 45Gb! But I am uncertain if its TimeSeriesDataset. But the model I use employs it though, so hopefully this could fix it.

fkiraly:
Why did you close this, @Vishnu-Rangiah?

Issue #1901, opened 2025-06-25
[BUG] EncoderNormalizer(method='standard', center=False) has wrong scale


cngmid:
**Describe the bug**
When setting `EncoderNormalizer` with the `standard` method and no centering, the transformation is not what is expected. The encoder correctly set its `.center_` to zero, but its `.scale_` is not set to the std of the fitted values. It is set to the mean of the fitted values and hence can even be negative.

**To Reproduce**
```python
from pytorch_forecasting.data.encoders import EncoderNormalizer
import torch

# generate a random sequence with mean -0.5 and unit std
x = torch.randn(60) - 0.5
print('x mean and std:', x.mean(), x.std())

# create a standard encoder with centering
encoder = EncoderNormalizer(
    method='standard',
    center=True,
    max_length=None,
    transformation=None,
    method_kwargs={}
)

# fit transform
y = encoder.fit_transform(x)

# center_ and scale_ match mean and std of x
print('center_ and scale_:', encoder.center_, encoder.scale_)

# the transformed is what is expected
print('error:', (y - (x - x.mean()) / x.std()).abs().max())


# create a standard encoder with *NO* centering
encoder = EncoderNormalizer(
    method='standard',
    center=False,
    max_length=None,
    transformation=None,
    method_kwargs={}
)

# fit transform
y = encoder.fit_transform(x)

# center_ is zero as expected, but scale_ *is not* the std of x, it is the *mean*
print('center_ and scale_:', encoder.center_, encoder.scale_)

# the transformed is *not* what is expected
print('error:', (y - x / x.std()).abs().max())
```

**Expected behavior**
In the `centering=False` case, `encoder.scale_` should be equal to `x.std()` (=0.9360) and `y` should be equal to ` x / x.std()`.

**Additional context**
In the code for the method `_set_parameters()` of the `class TorchNormalizer`, I don't understand the lines:
```
if not self.center and self.method != "identity":
            self.scale_ = self.center_
```

**Versions**
1.4.0

fkiraly:
This is indeed strange. It does look like a bug.

fnhirwa:
This is a bug going to open a PR to fix this.

PR #1902, opened 2025-06-25
[BUG] Fix issue with `EncodeNormalizer(method='standard', center=False)` for scale value

fnhirwa:
Fixes https://github.com/sktime/pytorch-forecasting/issues/1901
This PR fixes the issue with 
```python
encoder = EncoderNormalizer(
    method='standard',
    center=False,
    max_length=None,
    transformation=None,
    method_kwargs={}
)
```
which was returning `scale_` as `mean` of `x` instead its `std`

Temporarily backported lazywhere implementation from `scipy._lib._util.` to avoid import errors in the `statsmodels` with the latest scipy version.

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -191,7 +191,10 @@ def test_NegativeBinomialDistributionLoss(center, transformation):
         )
         samples = loss.sample(rescaled_parameters, 1)
         assert torch.isclose(target.mean(), samples.mean(), atol=0.1, rtol=0.5)
-        assert torch.isclose(target.std(), samples.std(), atol=0.1, rtol=0.5)
+        if transformation == "log1p" and not center:
+            assert torch.isclose(target.std(), samples.std(), atol=0.1, rtol=0.8)
+        else:
+            assert torch.isclose(target.std(), samples.std(), atol=0.1, rtol=0.5)
 
 
 @pytest.mark.parametrize(
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -664,7 +664,6 @@ class TorchNormalizer(
                 )
             self.scale_ = (q_75 - q_25) / 2.0 + eps
         if not self.center and self.method != "identity":
-            self.scale_ = self.center_
             if isinstance(y_center, torch.Tensor):
                 self.center_ = torch.zeros_like(self.center_)
             else:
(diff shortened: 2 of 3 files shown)


tests/test_metrics.py
```diff
@@ -191,7 +191,10 @@ def test_NegativeBinomialDistributionLoss(center, transformation):
         )
         samples = loss.sample(rescaled_parameters, 1)
         assert torch.isclose(target.mean(), samples.mean(), atol=0.1, rtol=0.5)
-        assert torch.isclose(target.std(), samples.std(), atol=0.1, rtol=0.5)
+        if transformation == "log1p" and not center:
+            assert torch.isclose(target.std(), samples.std(), atol=0.1, rtol=0.8)
```
fnhirwa, line 195: The variance distorts when the data is not centered, as the `log1p` transformation is `log(1+x)`

cngmid:
Linked to this issue, the method
```
    @property
    def min_length(self):
        if self.method == "identity":
            return 0  # no timeseries properties used
        elif self.center:
            return 1  # only calculation of mean required
        else:
            return 2  # requires std, i.e. at least 2 entries
```
of `class EncoderNormalizer` might also need to be reviewed.
If the method is not `"identity"`, one always needs at least 2 entries, independently whether `self.center` is `True` or `False`.

cngmid:
I had a closer look at the code of the method `TorchNormalizer._set_parameters()` in `encoders.py`. The block does still not make much sense:
```
if not self.center and self.method != "identity":
    self.scale_ = self.center_
    if isinstance(y_center, torch.Tensor):
        self.center_ = torch.zeros_like(self.center_)
    else:
        self.center_ = np.zeros_like(self.center_)
```
In the case `method= "robust", `self.center_` is set earlier in the method and *should not be modified* here.
I believe these lines should be moved into the block `elif self.method == "standard":`.
Would you agree?

fkiraly:
FYI, if you have time to look at the `__array_wrap__` issue, that would be appreciated - I tried here, but apparently I am not catching all cases that the function input may have: https://github.com/sktime/pytorch-forecasting/pull/1855

fnhirwa:
> FYI, if you have time to look at the `__array_wrap__` issue, that would be appreciated - I tried here, but apparently I am not catching all cases that the function input may have: #1855

sure will look into this🤞

PR #1903, opened 2025-06-26
[BUG] Fix issue with `plot_prediction_actual_by_variable`  unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'

fnhirwa:
fixes https://github.com/sktime/pytorch-forecasting/issues/1822

This PR fixes the issue with `plot_prediction_actual_by_variable` where the input `np.ndarray` was being multiplied to `torch.Tensor` in `TorchNormalizer` `__call__`

### DIFF
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -727,13 +727,13 @@ class TorchNormalizer(
         else:
             return y
 
-    def inverse_transform(self, y: torch.Tensor) -> torch.Tensor:
+    def inverse_transform(self, y: Union[torch.Tensor, np.ndarray]) -> torch.Tensor:
         """
         Inverse scale.
 
         Parameters
         ----------
-        y: torch.Tensor
+        y: Union[torch.Tensor, np.ndarray]
             scaled data
 
         Returns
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
we should add a test for this, if it fixes a bug

fkiraly:
Can you explain why we are changing tests, and not adding one? I was expecting a new tests.

In either case, I may need an explanation on your changes to the tests, that would be appreciated.

fnhirwa:
> Can you explain why we are changing tests, and not adding one? I was expecting a new tests.
> 
> In either case, I may need an explanation on your changes to the tests, that would be appreciated.

In the old test cases we were defaulting to `torch.Tensor` as a testing example https://github.com/sktime/pytorch-forecasting/blob/cc6a3d71d35c226cb47fc3f63b85aa136dc533e8/tests/test_data/test_encoders.py#L82
With this modification we can cover both `np.ndarray`, `pd.Series` and `torch.Tensor` as inputs to the Normalizer.

fkiraly (approved):
oh, you just moved it to the `mark.parametrize`, and added code to translate string to the object, is that correct?

Minor, non-blocking comment: my gut feeling says `pytest.fixture` in this case, for the str-to-object code, but it is clear enough what is happening (so, not blocking)

fnhirwa:
> oh, you just moved it to the `mark.parametrize`, and added code to translate string to the object, is that correct?
> 
> Minor, non-blocking comment: my gut feeling says `pytest.fixture` in this case, for the str-to-object code, but it is clear enough what is happening (so, not blocking)

I don't think making it a fixture would be a better design choice given that no other unit tests use this fixture.

fkiraly (approved):

PR #1908, opened 2025-07-01
[ENH] Improve test framework for v1 models

### LINKED ISSUE #1906: [ENH] Improve the test framework of v1 and make it more robust
The current test framework for model testing in v1 is inspired from the previous "hard-coded" tests for the models and needs some improvements to test which models are compatible with what kind of `loss` functions. 
Right now:
- The output format of the `forward` pass depends on the type of loss being used.
- There is no concrete contract defining the expected output format for each model-loss combination.
- This makes it difficult to know the model behavior and leads to fragile tests.

Moreover, right now, not all models in the package are tested through this test framework, so we need to add all the remaining models to the test framework as well.

We define  a tag `info:compatible_loss` (or any other better name?).
This tag is a list of strings, each representing a class of loss functions the model supports:
*  `"distr"` - Distribution Losses like `NormalDistributionLoss`
* `"quantile"` - `QuantileLoss`
* `"point"` - Point Prediction Losses like `MAE`

The tag for each model will list one or more of these strings, indicating compatible loss classes. For example:

```python
"info:compatible_loss" = ["point"]
```
means the model is compatible with point prediction losses, but not distribution losses and quantile losses.

Test Behavior:

- For each compatible loss class, the model will be tested with all applicable losses in that class to catch potential "edge cases".
- Each test will also verify that the output format (2D/3D tensor, correct shape) matches expectations.

Special case — `MultiLoss`:
Based on current manual testing, we assume that all models are compatible with `MultiLoss` (which combines multiple loss classes).
- All models will therefore be tested with `MultiLoss`.
- If any model fails when tested with `MultiLoss`, we will introduce an additional string `"multi"` to the tag list, indicating that the model explicitly supports `MultiLoss`.

Optionally, We should also think of a way to directly save the output format for each model-loss (or model-loss class?) pair, to see which models output formats are supported for each pair. This functionality later can be disabled. This could provide some insights about the output formats so that we can decide on eventual unified output format contract for all models.
Any thoughts @fkiraly , @agobbifbk, @PranavBhatP, @xandie985?

**Additional context**
This effort is mainly to understand more about `Metrics` and models and their interactions for the[ design document](https://hackmd.io/DL0_NjbERIy65laob22GlA) on concrete contracts model outputs and metrics.

phoeenniixx:
#### Reference Issues/PRs
Fixes #1906 

The PR makes the changes as suggested in #1906:

- [x]  Add tags `"info:y_type"` and `"info:pred_type"` to `model_pkg` classes to keep a track of losses a model is compatible with.
- [x] Add "loss classes" :
    -   `"distr"` - Distribution Losses like NormalDistributionLoss
    -   `"quantile"` - QuantileLoss
    -   `"point"` - Point Prediction Losses like MAE
- [x] Tests that loop over models and each loss that model is compatible with and test against common vignettes
- [ ] Add tests for `MQF2DistributionLoss` with the models it is compatible with
- [ ] Tests for `MultiLoss` containing only losses compatible with a given model.
- [x] Tests for output format contracts
- [x] Adds the remaining Models to the test framework


phoeenniixx:
For now only `DeepAR` tests are added (not the `MultiLoss` tests). Will add rest of the models in subsequent commits

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -164,9 +169,35 @@ def _generate_trainer_kwargs(self, test_name, **kwargs):
...
+                                and key in final_params
+                                and isinstance(final_params[key], dict)
+                            ):
+                                final_params[key].update(value)
+                            else:
+                                final_params[key] = value
+
+                        all_train_kwargs.append(final_params)
+                        train_kwargs_names.append(f"base_params_{i}_{loss_name}")
+        else:
```
phoeenniixx, line 235: This `if-else` block is just added so that the `model_pkg` classes that still not have the tag `info:compatible_loss` can also pass the tests, I will remove it once all the `model_pkg` classes have this tag

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -164,9 +169,35 @@ def _generate_trainer_kwargs(self, test_name, **kwargs):
...
+                for loss_instance in ALL_LOSSES_BY_TYPE.get(loss_type, []):
+                    loss_name = loss_instance.__class__.__name__
+                    loss_params = deepcopy(LOSS_SPECIFIC_PARAMS.get(loss_name, {}))
+                    loss_params["loss"] = loss_instance
+
+                    for i, base_params in enumerate(base_params_list):
+                        final_params = deepcopy(base_params)
+                        for key, value in loss_params.items():
+                            if (
+                                isinstance(value, dict)
```
fkiraly: 5 or more tabs is a code smell - can we avoid this?

phoeenniixx: sure, I moved the code to different functions

phoeenniixx:
This is strange, the tests are passing for macOS and ubuntu and failing for windows?
the error: 
```
self = LogNormal(), value = tensor([[1., 1., 1.],
        [1., 1., 0.]])

    def _validate_sample(self, value: Tensor) -> None:
        """
        Argument validation for distribution methods such as `log_prob`,
        `cdf` and `icdf`. The rightmost dimensions of a value to be
        scored via these methods must agree with the distribution's batch
        and event shapes.
    
        Args:
            value (Tensor): the tensor whose log probability is to be
                computed by the `log_prob` method.
        Raises
            ValueError: when the rightmost dimensions of `value` do not match the
                distribution's batch and event shapes.
        """
        if not isinstance(value, torch.Tensor):
            raise ValueError("The value argument to log_prob must be a Tensor")
    
        event_dim_start = len(value.size()) - len(self._event_shape)
        if value.size()[event_dim_start:] != self._event_shape:
            raise ValueError(
                f"The right-most size of value must match event_shape: {value.size()} vs {self._event_shape}."
            )
    
        actual_shape = value.size()
        expected_shape = self._batch_shape + self._event_shape
        for i, j in zip(reversed(actual_shape), reversed(expected_shape)):
            if i != 1 and j != 1 and i != j:
                raise ValueError(
                    f"Value is not broadcastable with batch_shape+event_shape: {actual_shape} vs {expected_shape}."
                )
        try:
            support = self.support
        except NotImplementedError:
            warnings.warn(
                f"{self.__class__} does not define `support` to enable "
                + "sample validation. Please initialize the distribution with "
                + "`validate_args=False` to turn off validation."
            )
            return
        assert support is not None
        valid = support.check(value)
        if not torch._is_all_true(valid):
>           raise ValueError(
                "Expected value argument "
                f"({type(value).__name__} of shape {tuple(value.shape)}) "
                f"to be within the support ({repr(support)}) "
                f"of the distribution {repr(self)}, "
                f"but found invalid values:\n{value}"
            )
E           ValueError: Expected value argument (Tensor of shape (2, 3)) to be within the support (GreaterThan(lower_bound=0.0)) of the distribution LogNormal(), but found invalid values:
E           tensor([[1., 1., 1.],
E                   [1., 1., 0.]])

C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\torch\distributions\distribution.py:317: ValueError
```

When I run locally this doesnot happen. Although I am on ubuntu so I cannot say much, maybe someone with windows test what the issue is?

fkiraly:
Is it always the same test, `test_integration[DecoderMLP-base_params-1-LogNormalDistributionLoss]`?

I am suspecting there is something with numerical accuracy happening, e.g., values that are close to zero but not quite zero.

phoeenniixx:
> Is it always the same test, `test_integration[DecoderMLP-base_params-1-LogNormalDistributionLoss]`?

Yes
> I am suspecting there is something with numerical accuracy happening, e.g., values that are close to zero but not quite zero.

I'll then shift to windows and will see locally what exactly is happening

fkiraly:
one trick is also to introduce prints or artificial raises in the code that trigger only when you expect the error - although in this specific case I fear that is not going to be easy, so trying to reproduce locally might be your best bet.

phoeenniixx:
From what I could see in `dwc["target"]` , there are values like 0.0010 that might be causing the issue?
See [here](https://gist.github.com/phoeenniixx/124213ae2d02bd2f1512d060a745cde0) the whole  `dwc["target"]`  
Whole `dwc` [here](https://gist.github.com/phoeenniixx/e267296df9dbb8bba23577bb983ab722)

fkiraly:
Strange.

* can you explain which object at which stage of the test this corresponds to?
* do you have ea hypothesis how the 0.010 values cause an issue?
* do the 0.010 values occur only on windows? If yes, why?
* if no, why are they a problem only on windows?

phoeenniixx:
Hi, so `dwc` is dataframe output of `data_with_covariates()` called inside `_get_test_dataloaders_from` of `DecoderMLP_pkg`. At first, I thought the issue is `dwc["target"]` has some values like 0.0010 and this is being taken as 0 somewhere. But I didn't find any concrete evidence regarding this.

Then maybe the issue is different float types as you said on discord, but as we know ( and I checked) `dwc["target"]` is `numpy.float64` in both ubuntu and windows. So this was also a dead end.

Then I found that in `LogNormal` of ` torch.distributions`, `scale` has the bound to be positive and its the same bound that is failing. So, I thought maybe this could be the issue? but the `rescale_parameters` of `LogNormalDistributionLoss` [[source]](https://github.com/sktime/pytorch-forecasting/blob/cc6a3d71d35c226cb47fc3f63b85aa136dc533e8/pytorch_forecasting/metrics/distributions.py#L225) already adds `F.softplus`, but still to be sure, I added  `F.softplus` in `map_x_to_distribution` (see [here](https://github.com/sktime/pytorch-forecasting/blob/cc6a3d71d35c226cb47fc3f63b85aa136dc533e8/pytorch_forecasting/metrics/distributions.py#L236-L237)) to the `scale`:
```python
def map_x_to_distribution(self, x: torch.Tensor) -> distributions.LogNormal:
        scale=F.softplus(x[..., 1])
        return self.distribution_class(loc=x[..., 0], scale=scale)
````
And obviously this also didn't work.

After sometime, I found that somehow `y_actual` (from [here](https://github.com/sktime/pytorch-forecasting/blob/cc6a3d71d35c226cb47fc3f63b85aa136dc533e8/pytorch_forecasting/metrics/base_metrics.py#L1046)) is getting value 0.0 in tensors. This is because `y` in `training_step` is getting this value 0 from the `batch`. So what my current hypothesis is that somehow the transformation to tensors of `y` is the actual issue. I still need some time to diagnose.

From this exprience I learned one thing atleast, that debugging should be "targetted" and not just random guesses, I think my random gusses also have made this process so complex.

Edit: just to make it clear and add more context how the `y` is having 0.
this is a value I copied from debugger for `y` from `training_step`:
```python
y = (tensor([[1., 1., 1.],
        [1., 0., 0.]]), None)
```
and in some cases (when I would rerun the debugger), the value is something like:
```python
y = (tensor([[1., 1., 1.],
        [1., 1., 0.]]), None)
```
or 
```python
y = (tensor([[1., 1., 1.],
        [1., 1., 1.]]), None)
```
In the second case, (when there is no 0), in some other `batch` `y` would get 0, so eventually it will fail again.
The one thing I noticed though is 0 appears only in the second list, and first list of tensors will always be 1.

phoeenniixx:
>Then maybe the issue is different float types as you said on discord, but as we know ( and I checked) dwc["target"] is numpy.float64 in both ubuntu and windows

To see if the float types Changes after transformation to tensors, I first need to find the place where this transformation is being done and there I can add `assert` and see the float types. But I would need some time as where exactly this transformation happens and how the data moves across `TimeSeriesDataset` is still a mystery to me in many ways :)

phoeenniixx:
The issue was `min_encoder_length` and `min_prediction_length` were so small that they were creating some problematic batches (got this bug just by mistake) and that it is failing in windows is just because while shuffling, this error gets masked by linux and macos mainly because we limit the number of batches (see [here](https://github.com/sktime/pytorch-forecasting/blob/cc6a3d71d35c226cb47fc3f63b85aa136dc533e8/pytorch_forecasting/tests/test_all_estimators.py#L202-L204)) to only 2.

fkiraly:
to quickly summarize our discussion in the tech meet today: the `LogNormal` issue looks like a cursed failure that might be hard to diagnose (and possibly even a problem in `torch` itself)

It is also proving that the test framework works well! Since it seems to be a genuine issue, rather than something to do with the framework - and the framework is intended to increase coverage so these combination cases get picked up.

The approach I would recommend is to park detected failures that are not immediately clear how to fix, and move them to issues - after all, the failures were there before, just not detected. Add some skipping mechanism to skip precisely the failure, open an issue with a description, and then we can deal with it later - and turn the test back on in the same PR as the fix (this pointer should be there in the issue)

This approach is a corollary of "test driven development" where you start with tests that fail, and then gradually change your code to make the tests pass - in separate PR and step-wise. If you write new code, then all tests fail initially; but we are improving coverage and consistency of an existing code base, so only some tests will fail in this case.

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -36,7 +73,7 @@ class PackageConfig:
     excluded_tests = EXCLUDED_TESTS
 
 
-class BaseFixtureGenerator(_BaseFixtureGenerator):
+class ModelBaseFixtureGenerator(BaseFixtureGenerator):
```
PranavBhatP: Minor change request, can this be named - `ModelFixtureGenerator` or `EstimatorFixtureGenerator`?

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -1,21 +1,58 @@
...
+            and isinstance(final_dict[key], dict)
+        ):
+            final_dict[key].update(value)
+        else:
+            final_dict[key] = value
+
+    return final_dict
+
+
 class PackageConfig:
```
PranavBhatP, line 56: Again, a minor change request - Can this be changed to `EstimatorPackageConfig`?

PranavBhatP: Again, a minor change request - Can this be changed to `EstimatorPackageConfig`?

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -1,21 +1,58 @@
...
+    LOSSES_BY_PRED_AND_Y_TYPE,
+    get_compatible_losses,
+)
 
 # whether to test only estimators from modules that are changed w.r.t. main
 # default is False, can be set to True by pytest --only_changed_modules True flag
 ONLY_CHANGED_MODULES = False
 
 
+def _merge_dict(dict_1, dict_2):
```
PranavBhatP: I did not go through the changes deeply, could you briefly tell me the purpose of this method? Where is the merge used?

PranavBhatP:
Initial glances, looks good to me in terms of design, but I haven't reviewed the code very deeply. I get the basic idea of what your PR is trying to do. Will review completely once the PR is ready 👍🏾 .

pytorch_forecasting/tests/test_all_estimators.py
phoeenniixx: This is mainly used in `_generate_final_param_list` to merge the `base_params` with losses to create new params with loss and `base_param`

fkiraly: I would call it `_nested_update`

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -67,6 +69,11 @@ def __init__(
             UserWarning,
         )
 
+    @classproperty
```
fkiraly, line 72: duplicate

pytorch_forecasting/models/nhits/_nhits_pkg.py
```diff
@@ -0,0 +1,103 @@
...
+            dl_default_kwargs = dict(
+                target="volume",
+                time_varying_unknown_reals=["volume"],
+                add_relative_time_idx=True,
+            )
+            dl_default_kwargs.update(data_loader_kwargs)
+            dataloaders_with_covariates = make_dataloaders(dwc, **dl_default_kwargs)
+            return dataloaders_with_covariates
+
+        if "backcast_loss_ratio" in params:
```
fkiraly: this looks exactly like the code above. Why is this duplicate, and why ar ethere separate cases in the first instance?

pytorch_forecasting/models/base/_base_model_v2.py
phoeenniixx: Actually this is `_base_model_v2` for `tesst_all_estimators_v2` which also inherits from the same fixture generator

pytorch_forecasting/models/nhits/_nhits_pkg.py
phoeenniixx: Sure, I will merge them using `or`
although I still need time with `nhits` as the tests are failing for it and I need to understand why

phoeenniixx:
> * in `BaseModel`, the `pkg` attribute duplicates the existing one

Please see the above comment, this `pkg` attribute is for the v2 test framework where right now models inherit from a different `BaseModel`

> it would be great if we can avoid conditioning directly on specific losses in the data generation. Can we not move to properties of losses? Or is this due to interface inconsistencies?

Yes, I still need to see how much of it can be moved to `_loss_mapping` as every model-loss pair introduces a different kind of inconsistency. For sure there are some cases (like for `NegativeBinomialDistributionLoss`) where we can move conditions to `_loss_mapping.py` but I think I should complete atleast one round across all the models first, so that I can collect all the common conditions and can see if any edge cases exist... I will add them in the next round :)

I also think `_loss_mapping` is not a good name as now it is not just mapping of losses but has logic to collect losses and special conditions as well, I will change its name to `_loss_prop` or something soon

fkiraly:
FYI, I think @PranavBhatP is facing a similar issue in #1907, namely that there are a few "common cases" and a few "outlier cases". Separating the two is a bit of an art, but what we need to do in the end.

phoeenniixx:
This is strange, locally the same tests fails but the issue is different - errors like
```RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn```
but here the errors are like:
```TypeError: lightning.pytorch.trainer.trainer.Trainer() got multiple values for keyword argument 'callbacks'```

Although this should not happen, I am not passing any new callback, and also the error msg on my local system is different... why!?

And locally 11 tests are failed, here only 8?
first param of each compatible model paired with `MQF2DistributionLoss` is passing here, but failing in my local system

**Comment by phoeenniixx** [comment]
Summing up the discussion on discord:
There is some issues with the usage of `MQF2DistributionLoss` and to debug it, will take a lot of time so we will skip it for now, to be taken on in future.
See the full discussion here: https://discord.com/channels/1075852648688930887/1299315056093429761/1397305858282749952

**Comment by phoeenniixx** [comment]
I have also logged model-loss output format here: https://gist.github.com/phoeenniixx/52ab40913fe4d0858c0beaa25e1b4e8c
It maybe useful in future

**Review by fkiraly — APPROVED** [review]
Very nice! Looks clean to me, extremely great!

File: pytorch_forecasting/tests/test_all_estimators.py [hunk 8]
```diff
@@ -114,42 +120,97 @@ def _all_objects(self):
...
 
-    def _generate_object_pkg(self, test_name, **kwargs):
-        """Return object package fixtures.
-
-        Fixtures parametrized
-        ---------------------
-        object_pkg: object package inheriting from BaseObject
-            ranges over all object packages not excluded by self.excluded_tests
+    @staticmethod
+    def _check_required_dependencies(object_pkg):
```
PranavBhatP, line 124: I see that you have used the implementation similar to the one used in #1907 . There was a bug in the code and I have changed it, no cause for alarm. I have attached the correct version of the function below

```python
@staticmethod
    def _check_required_dependencies(object_pkg):
        """
        Skip tests if the required dependencies for the metric are not installed
        in your environment.
        """

        required_deps = object_pkg.get_class_tag("python_dependencies")
        if required_deps:
            return _check_soft_dependencies(required_deps, severity="none")
        return True
```

You can make the change.

fkiraly: I think there is a function already in `skbase` that does this, `_check_estimator_deps`

fkiraly: (I would recommend to do this in a separate PR to keep individual PR small)

PR #1911, opened 2025-07-03
[ENH] Suppress `__array_wrap__` warning in `numpy 2` for `torch` and `pandas`

fnhirwa:
fixes https://github.com/sktime/pytorch-forecasting/issues/1764

Adding the comments from the Initial PR https://github.com/sktime/pytorch-forecasting/pull/1855 by @fkiraly 

The reason is that `numpy 2` added additional arguments to `__array_wrap__` which was interacting with `torch.Tensor.__array_wrap__` or `pandas.Series.__array_wrap__` that did not have said arguments on earlier versions, in `TorchNormalizer`.

The warning can be avoided by explicit coercion to `torch` prior to mixed usage - avoiding mixed types across packages entirely.

This also makes the coercions more explicit, enabling future refactors where we may want to move to more "clean" type assumptions instead of polymorphic union types.

### DIFF
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -719,7 +719,10 @@ class TorchNormalizer(
             elif isinstance(y, np.ndarray):
                 y_was = "numpy"
                 np_dtype = y.dtype
-                y = torch.from_numpy(y.astype(np.float32))
+                try:
+                    y = torch.from_numpy(y)
+                except TypeError:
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
BRILLIANT!

I failed fixing this in https://github.com/sktime/pytorch-forecasting/pull/1855. I assume it was unpleasant to track down...

What would be great if you could also update the docstring with a list of all types the inputs can be. That would be useful later for a rework of the data and encoder layer - but I do not see this as blocking for this PR.

Issue #1881, opened 2025-06-09
[DOC] I think the doc is incorrect in a misleading way - need to confirm


travis-leith:
#### Describe the issue linked to the documentation

In this section (https://pytorch-forecasting.readthedocs.io/en/stable/models.html#number-of-timeseries-and-their-relation-to-each-other) the line 

> If the timeseries denote different entities or exhibit very similar patterns accross the board, a model such as NBeats will work as well.

seems like it should really be

> If the timeseries denote different entities or exhibit very similar patterns accross the board, a model such as NBeats will **not** work as well.

fkiraly:
yes, probably? PR appreciated.

Pinaka07:
Hey @fkiraly can you please assign me this issue. thanks

PR #1914, opened 2025-07-09
[DOC] Correct documentation for N-BEATS

Pinaka07:
#### Reference Issues/PRs

fixes https://github.com/sktime/pytorch-forecasting/issues/1881
#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -71,7 +71,7 @@ a model that can learn relations between the timeseries can improve accuracy.
 Not that only :ref:`models that can process covariates <model-covariates>` can
 learn relationships between different timeseries.
 If the timeseries denote different entities or exhibit very similar patterns accross the board,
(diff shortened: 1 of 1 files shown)


Pinaka07:
"If the time series denote different entities or exhibit very similar patterns accross the board, a model such as N-BEATS will not work as well."

This might be slightly misleading — N-BEATS is actually designed to work well when time series are from different entities but share similar patterns, as it learns global structures across series.

It typically struggles when the time series are from different entities and exhibit highly diverse or uncorrelated patterns.

Consider rephrasing to:
 "If the timeseries denote different entities or exhibit very similar patterns across the board, a model such as N-Beats will not work as well."

fnhirwa (approved):
Looks Good to me😊

Pinaka07:
Thanks @fnhirwa 😊

Issue #1182, opened 2022-11-13
MQF2DistributionLoss - RuntimeError: Expected all tensors to be on the same device


andreareds:
- PyTorch-Forecasting version: 0.10.3
- PyTorch version: 1.12.1
- Python version: 3.10.6
- Operating System: 20.04.5 LTS

### Expected behaviour

I want to train a TFT model using the MQF2DistributionLoss for a set of bivariate time series.

### Actual behaviour

However, when I try to train, I get the following error. I tried to move the loss .to(torch.device('cuda')) as specified by the exception, and the model can be trained, but then I had an error in the inference step when I made the prediction.

`  
Traceback (most recent call last):
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/pytorch_forecasting/metrics/base_metrics.py", line 342, in forward
    res = metric(
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/torchmetrics/metric.py", line 245, in forward
    self._forward_cache = self._forward_reduce_state_update(*args, **kwargs)
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/torchmetrics/metric.py", line 309, in _forward_reduce_state_update
    self.update(*args, **kwargs)
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/torchmetrics/metric.py", line 395, in wrapped_func
    update(*args, **kwargs)
TypeError: MultiHorizonMetric.update() got an unexpected keyword argument 'encoder_target'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/torchmetrics/metric.py", line 395, in wrapped_func
    update(*args, **kwargs)
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/pytorch_forecasting/metrics/base_metrics.py", line 785, in update
    losses = self.loss(y_pred, target)
  File "/home/arossi/anaconda3/envs/tsapt/lib/python3.10/site-packages/pytorch_forecasting/metrics/distributions.py", line 368, in loss
    loss = distribution.energy_score(y_actual)
[...]

PR #1916, opened 2025-07-10
[BUG] Device inconstency in `MQF2DistributionLoss` raising: RuntimeError: Expected all tensors to be on the same device

fnhirwa:
Fixes https://github.com/sktime/pytorch-forecasting/issues/1182

In the current implementation, the `picnn` is initialized during class construction, and the device that it defaults to isn't being updated when the model is moved to another device.

* Added device movement method `to()` to ensure that `picnn` is moved along with the loss function.
* Added automatic device sync in `map_x_to_distribution` to ensure that `picnn` is on the same device as the input tensor.

### DIFF
--- a/pytorch_forecasting/metrics/distributions.py
+++ b/pytorch_forecasting/metrics/distributions.py
@@ -394,6 +394,7 @@ class MQF2DistributionLoss(DistributionLoss):
         self.prediction_length = prediction_length
         self.es_num_samples = es_num_samples
         self.beta = beta
+        self._transformation = None
 
         # define picnn
         convexnet = PICNN(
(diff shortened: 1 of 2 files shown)


fkiraly:
Requesting review from @PranavBhatP - is this loss currently tested via the framework?

PranavBhatP:
No, not in the present state on `main`, because it requires soft dependencies to run a CI test on this metric. This is the only reason I can think of for excluding this metric in the current `test_metrics.py`.

PranavBhatP:
I see that this PR handles the soft dependency, so I think it's a good start! Will review this soon.

fkiraly:
what is the status of this one? Given that we can handle soft deps in the test framework now.

fnhirwa:
> what is the status of this one? Given that we can handle soft deps in the test framework now.

There were some minor attributes mismatch in the device mocking function, working on the fix it will be ready soon.

fnhirwa:
This is ready for review now!

fkiraly (approved):

PR #1923, opened 2025-07-21
[ENH] Remove references to model from the `BaseFixtureGenerator`

phoeenniixx:
The current `BaseFixtureGenerator` is intended to serve as a common base class for generating fixtures in both model and metric test frameworks. However, it still contains model-specific references (e.g., `all_model_pkgs`), which is against its general-purpose design.

This PR refactors the class to remove model-specific terminology and behavior. It introduces more generic naming conventions (e.g., renaming `all_model_pkgs` to `all_pkgs`)

FYI @PranavBhatP

### DIFF
--- a/pytorch_forecasting/tests/_base/_fixture_generator.py
+++ b/pytorch_forecasting/tests/_base/_fixture_generator.py
@@ -115,7 +115,7 @@ class BaseFixtureGenerator(_BaseFixtureGenerator):
         if "object_pkg" in kwargs.keys():
             all_pkgs = [kwargs["object_pkg"]]
         else:
-            # call _generate_estimator_class to get all the classes
+            # call _generate_object_pkg to get all the packages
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
This is purely internal, so ok as a refactor.

May I suggest to also change `get_model_cls` to `get_cls`, and change accordingly downstream? Can be in a separate PR, or you can add it to this one.

phoeenniixx:
> May I suggest to also change `get_model_cls` to `get_cls`, and change accordingly downstream? Can be in a separate PR, or you can add it to this one.

I think this can be done in the respective PRs (#1908 and #1907) as that includes changes to the `obj_pkg` classes, which we are already changing (and adding) in those PRs

fkiraly:
ok, makes senes - I assume you realize that this will require changes in the respective common base class as well. If you do this identically in both PR, there will be no merge conflicts.

PR #1950, opened 2025-08-20
Fixing bug #1930 with proper syntax

### LINKED ISSUE #1930: [BUG] TimeSeriesDataSet wrong dtype when time_idx is included
**Describe the bug**
When `time_idx` or some other integer feature is included in the unknow reals, the dataset/dataloader elements are converted into the wrong dtype.

**To Reproduce**
The standard case, where scalers are defaulted to `sklearn.preprocessing.StandardScaler`:
```python
# Define the dataset
dataset = TimeSeriesDataSet(
    data,
    time_idx="time_idx",
    target="value",
    group_ids=["group"],
    static_categoricals=["group"],
    max_encoder_length=4,
    max_prediction_length=2,
    target_normalizer=None,
    # WATCH THIS
    scalers=None
)

dataloader = dataset.to_dataloader(batch_size=8)

x, y = next(iter(dataset))
print(f"dataset element {(x['x_cont'].dtype, y[0].dtype)} all features are set to int!")
display(x['x_cont'], y[0])

x, y = next(iter(dataloader))
print(f"dataloader element {(x['encoder_cont'].dtype, y[0][[0]].dtype)} all features are set to int!")
display(x['encoder_cont'][[0]], y[0][[0]])
```
This looks fine.

If no scaler is set however:
```
# Define the dataset
dataset = TimeSeriesDataSet(
    data.copy(),
    time_idx="time_idx",
    target="value",
    group_ids=["group"],
    static_categoricals=["group"],
    max_encoder_length=4,
    max_prediction_length=2,
    target_normalizer=None,
    # WATCH THIS
    scalers=dict(time_idx=None)
)

dataloader = dataset.to_dataloader(batch_size=8)

x, y = next(iter(dataset))
print(f"dataset element {(x['x_cont'].dtype, y[0].dtype)} all features are set to int!")
display(x['x_cont'], y[0])

x, y = next(iter(dataloader))
print(f"dataloader element {(x['encoder_cont'].dtype, y[0][[0]].dtype)} all features are set to int!")
display(x['encoder_cont'][[0]], y[0][[0]])
```
All the features have been cast to integer instead of floating.

**Source of the Problem**
The source of the problem lies in the function `TimeSeriesDataSet._data_to_tensors._to_tensor`:
```
elif isinstance(cols, list):  # and len(cols) > 0
    dtypekind = data.dtypes[cols[0]].kind
```
If there is more than one real, the `dtype` is selected **only based on the first real feature**. In our second example, the first feature is `time_idx`, which is an `np.int64`.

Notice that this is also true, if the first (or all! see [issue #958](https://github.com/sktime/pytorch-forecasting/issues/958)) real feature(s) turn(s) out to be represented in the source DataFrame by integers. In this case, one could fix the problem by enforcing a better `dytpe` on the relevant source DataFrame columns.

This is however not possible for `time_idx`, which has to be an integer.

**Possible Resolution**
A possible resolution is to modify the method as follows. An additional parameter `real=False` is added to force the continuous variables to be real.

I am not sure that the target should be forced to be real or shall be left as integers.

```
def _data_to_tensors(self, data: pd.DataFrame) -> Dict[str, torch.Tensor]:
    """Convert data to tensors for faster access with :py:meth:`~__getitem__`.

    Parameters
    ----------
    data : pd.DataFrame
        preprocessed data

    Returns
    -------
    Dict[str, torch.Tensor]
        dictionary of tensors for continous, categorical data, groups, target and
        time index
    """

    def _to_tensor(cols, long=True, real=False):  # <== **MODIFICATION**
        """Convert data[cols] to torch tensor.

        Converts sub-frames to numpy and then to torch tensor.
        Makes the following choices for types:

        * float columns are converted to torch.float
        * integer columns are converted to torch.int64 or torch.long,
          depending on the long argument
        """
        if not isinstance(cols, list) and cols not in data.columns:
            return None
        if isinstance(cols, list) and len(cols) == 0:
            dtypekind = "f"
        elif isinstance(cols, list):  # and len(cols) > 0
            # dtypekind = data.dtypes[cols[0]].kind       ## <== **MODIFICATION**
            dtypekind = data[cols].to_numpy().dtype.kind  ## <== **MODIFICATION**
        else:
            dtypekind = data.dtypes[cols].kind
        if real:                                                                     ## <== **MODIFICATION**
            return torch.tensor(data[cols].to_numpy(np.float64), dtype=torch.float)  ## <== **MODIFICATION**
        if not long:
            return torch.tensor(data[cols].to_numpy(np.int64), dtype=torch.int64)
        elif dtypekind in "bi":
            return torch.tensor(data[cols].to_numpy(np.int64), dtype=torch.long)
        else:
            return torch.tensor(data[cols].to_numpy(np.float64), dtype=torch.float)

    index = _to_tensor(self._group_ids, long=False)
    time = _to_tensor("__time_idx__", long=False)
    categorical = _to_tensor(self.flat_categoricals, long=False)

    weight = _to_tensor("__weight__")

    # get target
    if isinstance(self.target_normalizer, NaNLabelEncoder):
        target = [_to_tensor(f"__target__{self.target}")]
    else:
        ## [OPTIONAL] **MODIFICATION** (previous behaviour: set real to False)
        # if not isinstance(self.target, str):  # multi-target
            # target = [_to_tensor(f"__target__{name}") for name in self.target_names]
        # else:
            # target = [_to_tensor(f"__target__{self.target}", real=True)]
        if not isinstance(self.target, str):  # multi-target
            target = [
                # continuous target
                _to_tensor(f"__target__{name}", real=True)
                if name in self.time_varying_unknown_reals else
                _to_tensor(f"__target__{name}")
                for name in self.target_names]
        elif self.target in self.time_varying_unknown_reals:
            # continuous target
            target = [_to_tensor(f"__target__{self.target}", real=True)]
        else:
            target = [_to_tensor(f"__target__{self.target}")]

    # continuous covariates
    continuous = _to_tensor(self.reals, real=True)  ## <== **MODIFICATION**

    tensors = dict(
        reals=continuous,
        categoricals=categorical,
        groups=index,
        target=target,
        weight=weight,
        time=time,
    )
    return tensors
```

**Additional Comment**
As a final comment, this method assumes that the default `torch.dtype` is `torch.float32`. This might not be true. However other part of the code, in particular `BaseModel` would have to be modified too. See dicussion in [pull #1025](https://github.com/sktime/pytorch-forecasting/pull/1025).

**Versions**
v1.4.0

cngmid:
#### Reference Issues/PRs
Fixes #1930.

When `time_idx` or some other integer feature is included in the unknow reals, the dataset/dataloader elements are converted into the wrong dtype.

Modify the function the method `TimeSeriesDataSet._data_to_tensors._to_tensor()`

Added test function to `tests/test_data/test_timeseries.py` 

#### PR checklist

- [ x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [x ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries.py
@@ -1430,7 +1430,7 @@ class TimeSeriesDataSet(Dataset):
             time index
         """
 
-        def _to_tensor(cols, long=True) -> torch.Tensor:
+        def _to_tensor(cols, long=True, real=False) -> torch.Tensor:
             """Convert data[cols] to torch tensor.
 
             Converts sub-frames to numpy and then to torch tensor.
@@ -1439,16 +1439,21 @@ class TimeSeriesDataSet(Dataset):
             * float columns are converted to torch.float
             * integer columns are converted to torch.int64 or torch.long,
               depending on the long argument
+            * mixed columns are converted to their commonest type.
+            * if real argument is True, force the conversion to torch.float
             """
             if not isinstance(cols, list) and cols not in data.columns:
                 return None
             if isinstance(cols, list) and len(cols) == 0:
                 dtypekind = "f"
             elif isinstance(cols, list):  # and len(cols) > 0
-                dtypekind = data.dtypes[cols[0]].kind
+                # dtypekind = data.dtypes[cols[0]].kind
+                dtypekind = data[cols].to_numpy().dtype.kind
             else:
                 dtypekind = data.dtypes[cols].kind
-            if not long:
+            if real:
+                return torch.tensor(data[cols].to_numpy(np.float64), dtype=torch.float)
+            elif not long:
                 return torch.tensor(data[cols].to_numpy(np.int64), dtype=torch.int64)
             elif dtypekind in "bi":
                 return torch.tensor(data[cols].to_numpy(np.int64), dtype=torch.long)
--- a/tests/test_data/test_timeseries.py
+++ b/tests/test_data/test_timeseries.py
@@ -678,3 +678,41 @@ def test_graph_sampler(test_dataset):
         if idx > 100:
             break
     print(a)
+
+
+def test_correct_dtype_inference():
+    # Create a small dataset
+    data = pd.DataFrame(
+        {
+            "time_idx": np.arange(30),
+            "value": np.sin(np.arange(30) / 5) + np.random.normal(scale=1, size=30),
+            "group": ["A"] * 30,
+        }
+    )
+
+    # Define the dataset
+    dataset = TimeSeriesDataSet(
+        data.copy(),
+        time_idx="time_idx",
+        target="value",
+        group_ids=["group"],
+        static_categoricals=["group"],
+        max_encoder_length=4,
+        max_prediction_length=2,
+        time_varying_unknown_reals=["value"],
+        target_normalizer=None,
+        # WATCH THIS
+        time_varying_known_reals=["time_idx"],
+        scalers=dict(time_idx=None),
+    )
+
+    # and the dataloader
+    dataloader = dataset.to_dataloader(batch_size=8)
+
+    x, y = next(iter(dataset))
+    # real features must be real
+    assert x["x_cont"].dtype is torch.float32
+
+    x, y = next(iter(dataloader))
+    # real features must be real
+    assert x["encoder_cont"].dtype is torch.float32


pytorch_forecasting/data/timeseries/_timeseries.py
```diff
@@ -1439,16 +1439,21 @@ def _to_tensor(cols, long=True) -> torch.Tensor:
...
+            * if real argument is True, force the conversion to torch.float
             """
             if not isinstance(cols, list) and cols not in data.columns:
                 return None
             if isinstance(cols, list) and len(cols) == 0:
                 dtypekind = "f"
             elif isinstance(cols, list):  # and len(cols) > 0
-                dtypekind = data.dtypes[cols[0]].kind
+                # dtypekind = data.dtypes[cols[0]].kind
+                dtypekind = data[cols].to_numpy().dtype.kind
```
fkiraly: why are we doing this? It feels wasteful to do the conversion for the entire data container, instead of looking up the dtype.

fkiraly (changes requested):
Looks good, I have a question about `to_numpy` - should we not avoid conversion of the entire data to `numpy`?

cngmid:
> Looks good, I have a question about `to_numpy` - should we not avoid conversion of the entire data to `numpy`?

Good point.

Would this be better?
```
dtypekind = np.result_type(*data[cols].dtypes.to_list()).kind
```

fkiraly:
> Would this be better?

Yes, I think this would avoid the memory leak, since we never read the values

cngmid:
> dtypekind = np.result_type(*data[cols].dtypes.to_list()).kind

I just pushed this change.

pytorch_forecasting/data/timeseries/_timeseries.py
```diff
@@ -1439,16 +1439,21 @@ def _to_tensor(cols, long=True) -> torch.Tensor:
             * float columns are converted to torch.float
             * integer columns are converted to torch.int64 or torch.long,
               depending on the long argument
+            * mixed columns are converted to their commonest type.
```
fkiraly: can you update this as well? I think the type is completely determined by the arguments now?

cngmid:
```
"""Convert data[cols] to torch tensor.

    Converts sub-frames to numpy and then to torch tensor.
    Makes the following choices for types:

    - real is True:
        * the sub-frame is converted to a torch.float tensor
    - long is True (and real is False):
        * the sub-frame is converted to a torch.long tensor
    - real is False and long is False:
        * if all columns are integer or boolean, the sub-frame is
          converted to a torch.int64 tensor
        * if one column is a float, the sub-frame is converted to
          a torch.float tensor
"""
```
Is it clearer like this?

fkiraly: yes, much clearer thanks! Is it true though?

cngmid: just pushed it through

cngmid:
> great, could you update the docstring?

done and pushed through

fkiraly (approved):
Great, thanks!

PR #1954, opened 2025-08-27
[ENH] `check_estimator` utility for checking new estimators against unified API contract

fkiraly:
This adds a `check_estimator` utility for checking new estimators against unified API contract, and test registry utilities.

Mostly based on the `sktime` utilities of the same names, changes are predominantly import path changes.

### DIFF
--- a/pytorch_forecasting/tests/test_class_register.py
+++ b/pytorch_forecasting/tests/test_class_register.py
@@ -1,4 +1,5 @@
 # copyright: pytorch-forecasting developers, BSD-3-Clause License (see LICENSE file)
+# mostly based on the sktime utilities of the same name (BSD-3 Clause)
 """Registry and dispatcher for test classes.
 
 Module does not contain tests, only test utilities.
--- a/pytorch_forecasting/tests/test_class_register.py
+++ b/pytorch_forecasting/tests/test_class_register.py
@@ -1,0 +1,64 @@
+# copyright: pytorch-forecasting developers, BSD-3-Clause License (see LICENSE file)
+"""Registry and dispatcher for test classes.
+
+Module does not contain tests, only test utilities.
+"""
+
+__author__ = ["fkiraly"]
+
+from inspect import isclass
+
+
+def get_test_class_registry():
+    """Return test class registry.
+
+    Wrapped in a function to avoid circular imports.
+
+    Returns
+    -------
+    testclass_dict : dict
+        test class registry
+        keys are scitypes, values are test classes TestAll[Scitype]
+    """
+    from pytorch_forecasting.tests.test_all_estimators import TestAllPtForecasters
+    from pytorch_forecasting.tests.test_all_estimators_v2 import TestAllPtForecastersV2
+
+    testclass_dict = dict()
+    testclass_dict["forecaster_pytorch_v1"] = TestAllPtForecasters
+    testclass_dict["forecaster_pytorch_v2"] = TestAllPtForecastersV2
+
+    return testclass_dict
+
+
+def get_test_classes_for_obj(obj):
+    """Get all test classes relevant for an object or estimator.
+
+    Parameters
+    ----------
+    obj : object or estimator, descendant of sktime BaseObject or BaseEstimator
+        object or estimator for which to get test classes
+
+    Returns
+    -------
+    test_classes : list of test classes
+        list of test classes relevant for obj
+        these are references to the actual classes, not strings
+        if obj was not a descendant of BaseObject or BaseEstimator, returns empty list
+    """
+    if hasattr(obj, "_pkg"):
+        obj = obj._pkg
+
+    testclass_dict = get_test_class_registry()
+
+    try:
+        obj_scitypes = obj.get_tag("object_type")
+        if not isinstance(obj_scitypes, (list, tuple, set)):
+            obj_scitypes = [obj_scitypes]
+    except Exception:
+        obj_scitypes = []
+
+    for obj_scitype in obj_scitypes:
+        if obj_scitype in testclass_dict:
+            test_clss += [testclass_dict[obj_scitype]]
+
+    return test_clss
--- a/pytorch_forecasting/utils/_estimator_checks.py
+++ b/pytorch_forecasting/utils/_estimator_checks.py
@@ -1,0 +1,230 @@
+"""Estimator checker for extension."""
+
+__author__ = ["fkiraly"]
+__all__ = ["check_estimator"]
+
+
+def check_estimator(
+    estimator,
+    raise_exceptions=False,
+    tests_to_run=None,
+    fixtures_to_run=None,
+    verbose=True,
+    tests_to_exclude=None,
+    fixtures_to_exclude=None,
+):
+    """Run all tests on one single estimator.
+
+    Tests that are run on estimator:
+        all tests in test_all_estimators
+        all interface compatibility tests from the module of estimator's scitype
+            for example, test_all_forecasters if estimator is a forecaster
+
+    Parameters
+    ----------
+    estimator : estimator class or estimator instance
+    raise_exceptions : bool, optional, default=False
+        whether to return exceptions/failures in the results dict, or raise them
+
+        * if False: returns exceptions in returned ``results`` dict
+        * if True: raises exceptions as they occur
+
+    tests_to_run : str or list of str, optional. Default = run all tests.
+        Names (test/function name string) of tests to run.
+        sub-sets tests that are run to the tests given here.
+    fixtures_to_run : str or list of str, optional. Default = run all tests.
+        pytest test-fixture combination codes, which test-fixture combinations to run.
+        sub-sets tests and fixtures to run to the list given here.
+        If both tests_to_run and fixtures_to_run are provided, runs the *union*,
+        i.e., all test-fixture combinations for tests in tests_to_run,
+            plus all test-fixture combinations in fixtures_to_run.
+    verbose : str, optional, default=True.
+        whether to print out informative summary of tests run.
+    tests_to_exclude : str or list of str, names of tests to exclude. default = None
+        removes tests that should not be run, after subsetting via tests_to_run.
+    fixtures_to_exclude : str or list of str, fixtures to exclude. default = None
+        removes test-fixture combinations that should not be run.
+        This is done after subsetting via fixtures_to_run.
+
+    Returns
+    -------
+    results : dict of results of the tests in self
+        keys are test/fixture strings, identical as in pytest, e.g., test[fixture]
+        entries are the string "PASSED" if the test passed,
+        or the exception raised if the test did not pass
+        returned only if all tests pass, or raise_exceptions=False
+
+    Raises
+    ------
+    if raise_exceptions=True,
+    raises any exception produced by the tests directly
+
+    Examples
+    --------
+    >>> from pytorch_forecasting.models import NBeats
+    >>> from pytorch_forecasting.utilss import check_estimator
+
+    Running all tests for NBeats class,
+    this uses all instances from get_test_params and compatible scenarios
+
+    >>> results = check_estimator(NBeats)
+    All tests PASSED!
+
+
+    Running specific test (all fixtures) for NBeats
+
+    >>> results = check_estimator(NBeats, tests_to_run="test_clone")
+    All tests PASSED!
+
+    {'test_clone[NBeats-0]': 'PASSED',
+    'test_clone[NBeatsr-1]': 'PASSED'}
+
+    Running one specific test-fixture-combination for NBeats
+
+    >>> check_estimator(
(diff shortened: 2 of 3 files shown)


fkiraly:
@phoeenniixx, the templated `check_estimator` is unable to retrieve any tests, and I do not understand why.

It would be appreciated if you could review, perhaps you overrode something?

pytorch_forecasting/tests/test_class_register.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Returns
+    -------
+    test_classes : list of test classes
+        list of test classes relevant for obj
+        these are references to the actual classes, not strings
+        if obj was not a descendant of BaseObject or BaseEstimator, returns empty list
+    """
+    if hasattr(obj, "_pkg"):
+        obj = obj._pkg
```
phoeenniixx: `obj._pkg` is a method, so it should be `obj._pkg()`

pytorch_forecasting/tests/test_class_register.py
```diff
@@ -0,0 +1,64 @@
...
+
+    testclass_dict = get_test_class_registry()
+
+    try:
+        obj_scitypes = obj.get_tag("object_type")
```
phoeenniixx: its `obj.get_class_tag()` in `ptf`

pytorch_forecasting/utils/_estimator_checks.py
```diff
@@ -0,0 +1,233 @@
...
+
+    for test_cls in test_clss_for_est:
+        test_cls_results = test_cls().run_tests(
+            estimator=estimator,
+            raise_exceptions=raise_exceptions,
+            tests_to_run=tests_to_run,
+            fixtures_to_run=fixtures_to_run,
+            tests_to_exclude=tests_to_exclude,
+            fixtures_to_exclude=fixtures_to_exclude,
+            verbose=verbose and raise_exceptions,
```
phoeenniixx: I was also looking at `QuickTester. run_tests()`, it has following `__init__`: 
```python
obj,
raise_exceptions=False,
tests_to_run=None,
fixtures_to_run=None,
tests_to_exclude=None,
fixtures_to_exclude=None,
```
are we overriding it somehow?

phoeenniixx:
I made the above changes to a branch, after pulling this PR locally, I see the `check_estimator` is able to collect tests, but it is collecting ALL the tests, for all the models, maybe because `test_clss_for_est` in `check_estimator` gets `[<class 'pytorch_forecasting.tests.test_all_estimators.TestAllPtForecasters'>]` and not the actual model class?

pytorch_forecasting/tests/test_class_register.py
fkiraly: `get_tag` also collects class tags, but good spot

fkiraly: this is probably it!

I keep saying, never use try/except for condition checking because it is bad style, instead check for the condition explicitly. Otherwise it is hard to diagnose if it fails to behave as expected - e.g., if it fails due to a genuine exception rather than the condition being false.

(try/except is fine for error handling, this is about abusing it for condition checking)

And now I make the same mistake. I should have listened to myself :-)

pytorch_forecasting/tests/test_class_register.py
fkiraly: did you change it to be a `classmethod`?

I originally designed it as `classproperty`, so this call would be valid.

pytorch_forecasting/tests/test_class_register.py
phoeenniixx: I actually did the same to find this out :) 
just removed the `try-except` block... Thanks to your teachings

pytorch_forecasting/tests/test_class_register.py
fkiraly: ah, this is actually another mistake. It is a `classattribute`, but `MyClass.pkg` should be called, not `_pkg`

phoeenniixx: No i didnt, i just looked at the `DeepAR` implementation, there it was already `classmethod`?
https://github.com/sktime/pytorch-forecasting/blob/3821c0bbca6d35efd00c1bcf51b42da1971b24a4/pytorch_forecasting/models/deepar/_deepar.py#L40-L45

phoeenniixx: You made it `classmethod` in #1888, ig

fkiraly: Yes, I know, I understood what is going on - `pkg` is a `classattribute` that is the public interface, and it calls the private extension locus `_pkg` which is a `classmethod`, like `fit` and `_fit`.

pytorch_forecasting/tests/test_class_register.py
fkiraly: well, all I am saying, it is worth listening to them oneself too.

pytorch_forecasting/utils/_estimator_checks.py
fkiraly: yes, this is also a bug!

`scikit-base` renames every "estimator" thing to "object" or "obj".

This is clearer for users of `scikit-base` imo, but causes a discrepancy that is prone to bugs like this.
(so I was originally against, even if "estimator" is a less clear term)

We had long discussions with other core devs on whether we should rename this way, in the end it was decided that we do.

fkiraly:
ok, I am now at the point that you had also arrived at - all tests are collected, instead of only those for the estimator

fkiraly:
ok, I think the issue is with the additional involvement of pkg.

The generator driving the collection is that of the pkg classes, which was not overwritten in `QuickTester`.

I have written an experimental override of `QuickTester` which now limits the pkg classes based on the argument of `run_tests`.

If this works, we should also refactor the `scikit-base` `QuickTester` to enable plugins like this - most of the code is duplicate.

fkiraly:
it seems to be collecting the right tests now, but they are failing when run through `check_estimator`?

FYI @phoeenniixx

fkiraly:
@phoeenniixx, I think it works now

phoeenniixx:
I just pulled the branch locally, when I try :
```python
check_estimator(my_model, raise_exceptions=True)
```
it prints output of all the tests, but I think in `sktime`, we only get the raised  exception? Can we do something similar here? I mean `check_estimator` just returns the raise exception and `"ALL TESTS PASSED!"`, if none is raised? This will  keep the console clean and the user will get only the problematic fixtures, rather than all the fixtures?

Because if there are like 50 test-fixtures to be tested, then it would be hard for the user to actually find which one is failing as the console also shows the passing ones and he'd have to move across the whole console to actually find the failing param...

What do you think?

fkiraly:
Anything the tests print gets printed by default, if `verbose=True`. If you put `verbose=False`, then nothing gets printed.

Or would you prefer that only the `check_estimator` printout gets printed, and nothing from within the tests?

fkiraly:
I have now changed the `verbose` argument to be an integer - at default, nothing from within the tests is printed, but it can be configured to print all by changing it to the integer 2.

phoeenniixx:
> I have now changed the `verbose` argument to be an integer - at default, nothing from within the tests is printed, but it can be configured to print all by changing it to the integer 2.

Yes that makes more sense! Thanks!
Though I have few questions/suggestions:
- I have a doubt about `verbose` - what is the difference between different values of `verbose`? When I try `verbose=0` I see something like this:

   image

   But I think `verbose=0` should not print all this? It should be at `verbose=1` or `verbose=2`. I think the "amount" of info printed should increase as value increase
 
     - Only error (if any) for `verbose=0` and the `dict` that is printed at last.
     - The info printed in above image (for each fixture) for `verbose=1`.
     - The whole output of the tests (along with the training `verbose` output) for each fixture in `verbose=2`.

- Should we add one more param `supress_warnings` which is `False` by default. Eg, right now, I was also testing it for v2 and as we know we having warnings at every layer - D1, D2, model etc, and this would be noise for someone who is debugging, so it would be good if we could supress them. Or does it already exist in `skbase` ?

fkiraly:
that is strange, I am suppressing stdout - line 246, via `StdoutMute` - why is it still being printed? Is it not coming from stdout?

fkiraly:
I have now added code that suppresses stderr as well. Do you still see it?

fkiraly:
I just tried, it is still printing - this is pretty aggressive. Where is it printing to, and why does it get displayed?

phoeenniixx:
I think I found the issue - its `lightning` logging. Actually `Trainer`'s logging is not supressed by `StdoutMute`, you'll have to specifically add these agrs to the `Trainer`:
```python
trainer = pl.Trainer(
        ...,
        enable_progress_bar=False,  # Disables the progress bar
        enable_model_summary=False,  # Disables the model summary printout
```
and it would still print some logs like: 
```
Running in `fast_dev_run` mode: will run the requested loop using 1 batch(es). Logging and checkpointing is suppressed.
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
HPU available: False, using: 0 HPUs
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
```
And to supress this, you'll have to use this statement before `trainer`:
```python
    import logging
    logging.getLogger("lightning.pytorch").setLevel(logging.WARNING)
```

phoeenniixx:
I think we'll have to edit `_integration` here for this to actually work :)
This I think makes the things a bit more complex then they are meant to be?

fkiraly:
it would be better if we could catch the printout directly at where it is printed to - otherwise the machinery is too coupled to specifics of `lightning`

phoeenniixx:
So, maybe we should do something like this:
- We add the params `enable_progress_bar` and `enable_model_summary` to `false` by default in `_integration`, to be made `True` only if the user says so.
- We add the `logging.getLogger("lightning.pytorch").setLevel(logging.WARNING)` at all times as idts we need in tests the info what we are using - its always cpu..?

fkiraly:
That will prevent the tests from printing, yes - I think that is a good solution for now.

The bigger problem is not resolved though, how to programmatically suppress lightning output.

phoeenniixx:
I asked AI the same question and I think it solved it for us :)
Create a special class:
```python
import logging
from contextlib import ContextDecorator

class LoggerMute(ContextDecorator):
    """Context manager to temporarily mute specified loggers."""
    def __init__(self, logger_names, active=True):
        self.logger_names = logger_names
        self.original_levels = {}
        self.active = active

    def __enter__(self):
        if self.active:
            for name in self.logger_names:
                logger = logging.getLogger(name)
                self.original_levels[name] = logger.level
                logger.setLevel(logging.CRITICAL + 1) # Mute all levels
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        if self.active:
            for name, level in self.original_levels.items():
                logger = logging.getLogger(name)
                logger.setLevel(level)
```
and use it in `run_tests`
```python
                # inside loop B
                loggers_to_mute = ["lightning.pytorch", "pytorch_lightning"]
                try:
                    # Add the LoggerMute to your with statement!
                    with StderrMute(active=verbose < 2), \
                            StdoutMute(active=verbose < 2), \
                            LoggerMute(loggers_to_mute, active=verbose < 2):

                        test_fun(**deepcopy(args))

                    results[key] = "PASSED"
                    print_if_verbose("PASSED")
```

phoeenniixx:
This is because `trainer` doesnot print, it "logs". I think that is why `stdoutmute` was not working..?

fkiraly:
but logging is not the same as printing. Printing happens on the console - so where does the printout come from? The python `logging` logger, for instance, does *not* automatically print.

phoeenniixx:
I tried reading the `Trainer` from `lightning`, the code flow is quite confusing but from what I am able to understand is this:
There is a class call `ModelSummary`, which gets the model summary and logs the model summary. See [here](https://github.com/Lightning-AI/pytorch-lightning/blob/44e57d06f39f8f0e071855e398bc3e4641719eaf/src/lightning/pytorch/callbacks/model_summary.py#L110).

And I tried jsut shutting the logging, and then this is not printed on the console. So, my suspicion is if you just suppress the logging, the info is never collected and it is not printed. From what I have been able to understand, sometimes people write to different streams then what `stdoutmute` can supress and that is why it is printed always. And here, the `log` is used to save the info (see [here](https://github.com/Lightning-AI/pytorch-lightning/blob/44e57d06f39f8f0e071855e398bc3e4641719eaf/src/lightning/pytorch/callbacks/model_summary.py#L110)) and then this info is somehow used to get the final model summary and is written to some different stream than what `std` can handle. So, I think we could try just supressing the `logging` here?
```python
    import logging
    logging.getLogger("lightning.pytorch").setLevel(logging.WARNING)
```

fkiraly:
how exactly would a context manager look like that shuts down the `lightning` printouts then? Is it simply setting the logger to warning?

PR #1973, opened 2025-09-20
[ENH] Added metadata Testing for models and data modules for ptf-v2

### LINKED ISSUE #1887: [ENH] `metadata` Testing for models and data modules for `ptf-v2`
Right now, the test framework (#1841) only checks if the `metadata` is a `dict`, but we should add the `metadata` testing to to test  if the `metadata` of datamodule is calculated correctly and passed correctly to dataloaders and model for initialisation.

The blocker is different `data_module`s might have different `metadata` configurations so we can't directly add these tests to the integration test.

One idea proposed by @fkiraly:
Create an intermediate model class (between `_BasePtForecasterV2` and `ModelMetadata` class) where we add a method `_check_metadata` that can take the specific test for metadata of datamodule based on which data module the model is using , i.e, if the model is using `EncoderDecoderDataModule` , `_check_metadata` will test the `metadata` of the `EncoderDecoderDataModule`. As this will be a parent of `ModelMetadata`, there will no duplication of code

So the idea is:
```python
class _BaseObject(_SkbaseBaseObject):
  pass

class _BasePtForecaster(_BaseObject):
    pass

class _BasePtForecasterV2(_BasePtForecaster):

    _tags = {
    }

class _EncoderDecoderConfigBase(_BasePtForecasterV2):
  def _check_metadata():
      # metadata tests

class TFTMetadata(_EncoderDecoderConfigBase):
  # testing logic

```
if this is a `TSLibDataModule` then, the `ConfigBase` class will change and will have tests specific for `TSLibDataModule`

ParamThakkar123:

This Pull request adds metadata testing for models and data modules for `ptf-v2`. I have added `ConfigBase` for the `EncoderDecoderTimeSeriesDataModule` and the `TslibDataModule`. Added new classes specific to each of data modules like for models using `EncoderDecoderTimeSeriesDataModule` I added `EncoderDecoderConfigBase` and for `TslibDataModule` I added `TslibConfigBase`

Yes, All tests passes locally

#### PR checklist

- [X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [X] Added/modified tests
- [X] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.

### DIFF
--- a/pytorch_forecasting/models/base/_base_object.py
+++ b/pytorch_forecasting/models/base/_base_object.py
@@ -118,3 +118,56 @@ class _BasePtForecasterV2(_BasePtForecaster_Common):
     _tags = {
         "object_type": "forecaster_pytorch_v2",
     }
+
+
+class _EncoderDecoderConfigBase(_BasePtForecasterV2):
+    def _check_metadata(self, metadata):
+        assert isinstance(metadata, dict)
+        required_keys = [
+            "encoder_cat",
+            "encoder_cont",
+            "decoder_cat",
+            "decoder_cont",
+            "target",
+            "max_encoder_length",
+            "min_encoder_length",
+            "max_prediction_length",
+            "min_prediction_length",
+            "static_categorical_features",
+            "static_continuous_features",
+        ]
+
+        for key in required_keys:
+            assert key in metadata, f"Key {key} missing in metadata"
+
+        assert metadata["encoder_cat"] >= 0
+        assert metadata["encoder_cont"] >= 0
+        assert metadata["decoder_cat"] >= 0
+        assert metadata["decoder_cont"] >= 0
+        assert metadata["target"] > 0
+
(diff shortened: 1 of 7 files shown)


fkiraly (changes requested):
Thanks, interesting idea!

I am a bit confused about the logic here. I assume that you intend the new test `test_d2_metadata` to retrieve the new classes that you added, but this will not happen? Since the `object_pkg` variable will be, for instance, `TimeXer_pkg_v2`, but that class will never have the check method?

Design-wise, I think the best place to locate the checker method is imo not the model package class, but the D2 layer data class that produces the metadata?

Issue #1975, opened 2025-09-29
[ENH] standardise output format of `NBeats` and `NBeatsKAN` estimators for point predictions


PranavBhatP:
**Is your feature request related to a problem? Please describe.**
As per the findings of #1971, it has been observed that `NBeats` and `NBeatsKAN` are non-conformant when it comes to point prediction outputs. Currently, they output a 2d tensor `(batch_size, prediction_length)`, in this case. This needs to be a 3d tensor `(batch_size,prediction_length, 1)`.

**Describe the solution you'd like**
A simple fix which adds a 3rd dimension to the existing 2D tensor return of the model. If the described fix is not enough, other approaches are welcome. Refer #1966 for some context on the exact output format which is also being tested and currently skipped for `NBeats`, `NBeatsKAN`, `xLSTMTime` and `TimeXer` (this is being addressed in #1936).

**Additional Context**
As a simple starting point, here is the code to be referenced for the changes at the link - https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nbeats/_nbeats_adapter.py#L69C9-L112C10. Same goes for NBeatsKAN

sanskarmodi8:
Hey @PranavBhatP, Can i work on this issue?

PranavBhatP:
sure, go ahead!

PR #1977, opened 2025-10-01
Standardize output format of NBeats and NBeatsKAN estimators

sanskarmodi8:
This PR updates the NBeatsAdapter to ensure that both NBeats and NBeatsKAN models return 3D tensors (batch_size, prediction_length, 1) for point predictions, aligning with the expected output format #1975.

### DIFF
--- a/pytorch_forecasting/models/nbeats/_nbeats_adapter.py
+++ b/pytorch_forecasting/models/nbeats/_nbeats_adapter.py
@@ -93,20 +93,20 @@ class NBeatsAdapter(BaseModel):
             forecast = forecast + forecast_block
 
         return self.to_network_output(
-            prediction=self.transform_output(forecast, target_scale=x["target_scale"]),
+            prediction=self.transform_output(forecast.unsqueeze(-1), target_scale=x["target_scale"]),
             backcast=self.transform_output(
-                prediction=target - backcast, target_scale=x["target_scale"]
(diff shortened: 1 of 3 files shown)


sanskarmodi8:
I'm not sure why this particular test is failing in CI. When I run `pytest -k "test_integration and NHiTS" -v` locally, it passes without any errors.

PranavBhatP:
The tests are passing because it is currently skipped for `NBeats` and `NBeatsKAN`. Go to their respective package files namely `NBeats_pkg` and `NBeatsKAN_pkg` and remove the test skip tag at the bottom of `_tags`.

PranavBhatP:
> When I run pytest -k "test_integration and NHiTS" -v

The model we are concerned about is NBeats not NHiTS right...

sanskarmodi8:
> > When I run pytest -k "test_integration and NHiTS" -v
> 
> The model we are concerned about is NBeats not NHiTS right...

Ah yes, you’re right—my bad! The test I ran locally was for NHiTS, but the model in my PR is NBeats. I just wanted to check that the changes I made didn’t break the workflow for a similar model.

PranavBhatP:
Thanks for the changes!

sanskarmodi8:
Hey, just wanted to check if there’s anything else needed from my side before it can be merged. Thanks!

PR #1978, opened 2025-10-01
standardize output format of xLSTMTime estimator for point predictions

sanskarmodi8:
This PR standardizes the output format of the xLSTMTime estimator for point predictions.


sanskarmodi8:
Hey, just wanted to check if there’s anything else needed from my side before it can be merged. Thanks!

PR #1980, opened 2025-10-05
[MNT] python 3.9 end-of-life

phoeenniixx:
Removes 3.9 from supported versions.

Should be released with 1.5.0


fkiraly:
I think the version in the notebook tests still needs to be updated.

PR #1988, opened 2025-10-30
[MNT] Fix typos

szepeviktor:

Correct misspellings with https://github.com/crate-ci/typos

Please help me find false positives.

_Inspired by @fkiraly_

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries_v2.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries_v2.py
@@ -70,7 +70,7 @@ class TimeSeries(Dataset):
     unknown : list of str, optional, default = no variables
         list of variables that are not known in the future,
         list may also contain list of str, which are then grouped together
-        (e.g. useful for whether categories).
+        (e.g. useful for weather categories).
     static : list of str, optional, default = all variables not in known, unknown
         list of variables that do not change over time,
         list may also contain list of str, which are then grouped together.


fkiraly (changes requested):
This seems to be causing some formatting failures?

szepeviktor:
> This seems to be causing some formatting failures?

Yes, I need to learn to run Python tools ...

szepeviktor:
I think running tools only on changed files (pre-commit) is a **very bad** idea.

fkiraly:
> I think running tools only on changed files (pre-commit) is a **very bad** idea.

why that?

There used to be a lot of unformatted legacy code, so for a while there would have been a large amount of failures in every PR.

I think it was a good idea and still is?

fkiraly (approved):
Went through all, and looks good! Thanks

szepeviktor:
Glad to contribute to Franzprojects.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -70,7 +70,7 @@ class TimeSeries(Dataset):
     unknown : list of str, optional, default = no variables
         list of variables that are not known in the future,
         list may also contain list of str, which are then grouped together
-        (e.g. useful for weather categories).
+        (e.g. useful for whether categories).
```
szepeviktor: Could this be really "weather"?

fkiraly: it probably is really weather

szepeviktor: Reverted.

fkiraly:
> Glad to contribute to Franzprojects.

Strictly speaking a JanBeitnerProject?

PR #1989, opened 2025-10-31
[BUG] Torch doesn't support the conversion of non-writeable numpyFix non-writeable encoder issue and update tests

### LINKED ISSUE #1987: [BUG] Torch doesn't support the conversion of non-writeable numpy array
**Describe the bug**
When converting pandas objects to numpy, the resulting arrays are non-writeable if the option `pd.options.mode.copy_on_write` is set to `True`, which will be the default in future pandas versions.
PyTorch does not support non-writable tensors. It warns that "writing to this tensor will result in undefined behavior".

This seems to affect principally the normalizers defined in `data\encoders.py`

I know that there is an effort to rewrite the class `TimeSeriesDataSet`. In case you would like to fix this within the current version, I could produce a Pull request.

**To Reproduce**
```python
import numpy as np
import pandas as pd

pd.options.mode.copy_on_write =True

# Create a small dataset
data = pd.DataFrame({
    "time_idx": np.arange(30),
    "value": np.sin(np.arange(30) / 5) + np.random.normal(scale=0.1, size=30),
    "feature": np.cos(np.arange(30) / 5) + np.random.normal(scale=0.1, size=30),
    "group": ["A"] * 30
})

# data["feature"].values.flags["WRITEABLE"] = False
data["feature"].values.flags
```
yields
```
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
```

Then
```python
from sklearn.preprocessing import StandardScaler
from pytorch_forecasting import TimeSeriesDataSet

# Define the dataset
dataset = TimeSeriesDataSet(
    data,
    time_idx="time_idx",
    target="value",
    group_ids=["group"],
    static_categoricals=["group"],
    max_encoder_length=4,
    max_prediction_length=2,
    time_varying_unknown_reals=["value", "feature"],
    target_normalizer=None,
    scalers = {"feature": StandardScaler()},
)
```
throws the warning
```
C:\Users\dcang\GitHub\pytorch-forecasting\pytorch_forecasting\data\encoders.py:718: UserWarning: The given  NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor    will result in undefined behavior. You may want to copy the array to protect its data or make it writable before  converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally   at C:\actions-runner\_work\pytorch\pytorch\pytorch\torch\csrc\utils\tensor_numpy.cpp:209.)
  y = torch.as_tensor(y)
```

**Additional context**
In the file `data\encoders.ps` there are two instances where a pandas object is converted into an array:
```
230                y = y.to_numpy()
716                y = y.values
```
both case should be replaced with `y = y.to_numpy(deep=True)`.

**Versions**
<details>
System:
    python: 3.10.16 | packaged by Anaconda, Inc. | (main, Dec 11 2024, 16:19:12) [MSC v.1929 64 bit (AMD64)]
executable: C:\ProgramData\miniconda3\envs\sq3\python.exe
   machine: Windows-10-10.0.26200-SP0

Python dependencies:
          pip: 25.0
pytorch-forecasting: 1.5.0
        torch: 2.7.1+cu126
    lightning: 2.5.5
        numpy: 1.26.4
        scipy: 1.15.2
       pandas: 2.3.3
      cpflows: 0.1.2
   matplotlib: 3.10.6
       optuna: 4.3.0
optuna-integration: 4.5.0
pytorch_optimizer: 2.12.0
 scikit-learn: 1.7.2
  scikit-base: None
  statsmodels: 0.14.5
</details>

cngmid:
#### Reference Issues/PRs
Fixes #1987

When converting pandas objects to numpy, the resulting arrays are non-writeable if the option `pd.options.mode.copy_on_write` is set to `True`, which will be the default in future pandas versions.
PyTorch does not support non-writable tensors. It warns that "writing to this tensor will result in undefined behavior".

Added test function to `tests/test_data/test_timeseries.py`
But the real way to test it is to set globally `pd.options.mode.copy_on_write = True` and see whther a warning appears somewhere.

#### Any other comments?
I think pandas v3.0 will enforce this flag.

#### PR checklist

- [ ] Added/modified tests


fnhirwa (approved):
hey @cngmid 

thanks for the PR, the changes looks good to me.

PR #1994, opened 2025-11-28
[ENH] Implementing the `iTransformer` model in PTFv2.

### LINKED ISSUE #1899: [ENH] implementing `iTransformer` from `tslib`.
Implementing the `iTransformer` model in PTFv2. 

Basis: https://github.com/thuml/iTransformer/blob/main/model/iTransformer.py
Paper: https://arxiv.org/pdf/2310.06625

JATAYU000:
#### Reference Issues/PRs

Fixes #1899 

Have started interfacing `iTransformer` in PTFv2, from the TSlib repository [thuml/iTransformer](https://github.com/thuml/iTransformer/blob/main/model/iTransformer.py)
Work in progress, would like to have suggestions on it.

- Current Implementation compliance with PTFv2

Not yet

#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,6 +114,40 @@ jobs:
         shell: bash
         run: python -m pytest
 
+  test-deps-2025:
+    needs: pytest-nosoftdeps
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.12"]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Get full Python version
+        id: full-python-version
+        shell: bash
+        run: echo version=$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))") >> $GITHUB_OUTPUT
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          pip install ".[dev, dependencies_2025, github-actions]"
+
+      - name: Show dependencies
+        run: python -m pip list
+
+      - name: Run pytest
+        shell: bash
+        run: python -m pytest
+
   pytest:
     name: Run pytest
     needs: pytest-nosoftdeps
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Release Notes
 
+## v1.6.1
+Patch release focusing on:
+* Bug fix  to solve the persisting bug of passing `weights_only` in `load_from_checkpoint` for `lightning <2.6`.
+* Bug fix to non-writeable encoder issue caused by pandas copy-on-write behavior.
+
+### Fixes
+
+* [BUG] Torch doesn't support the conversion of non-writeable numpyFix non-writeable encoder issue and update tests (#1989) @cngmid
+* [BUG] Solve the persisting bug of passing `weights_only` in `load_from_checkpoint` (#2027) @phoeenniixx
+
+### Maintenance
+
+* [MNT] Add CI step with pinned dependencies as of Nov 2025 (#2029) @phoeenniixx
+
+### All Contributors
+@cngmid, @phoeenniixx
+
 ## v1.6.0
 Release focusing on:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# Contributing guide
+
+You can find our contributing guide on our [website](https://www.sktime.net/en/latest/get_involved/contributing.html).
--- a/examples/ar.py
+++ b/examples/ar.py
@@ -3,7 +3,7 @@ import warnings
 import lightning.pytorch as pl
 from lightning.pytorch.callbacks import EarlyStopping, LearningRateMonitor
 import pandas as pd
-from pandas.core.common import SettingWithCopyWarning
+from pandas.errors import SettingWithCopyWarning
 import torch
 
 from pytorch_forecasting import GroupNormalizer, TimeSeriesDataSet
--- a/examples/stallion.py
+++ b/examples/stallion.py
@@ -5,7 +5,7 @@ import lightning.pytorch as pl
 from lightning.pytorch.callbacks import EarlyStopping, LearningRateMonitor
 from lightning.pytorch.loggers import TensorBoardLogger
 import numpy as np
-from pandas.core.common import SettingWithCopyWarning
+from pandas.errors import SettingWithCopyWarning
 
 from pytorch_forecasting import (
     GroupNormalizer,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "pytorch-forecasting"
 readme = "README.md"         # Markdown files are supported
-version = "1.6.0"            # is being replaced automatically
+version = "1.6.1"            # is being replaced automatically
 
 authors = [
   {name = "Jan Beitner"},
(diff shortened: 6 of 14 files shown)


pytorch_forecasting/models/itransformer/submodules.py
PranavBhatP, line 1: Conventionally for v2, all the layers of the models' architecture are present in the [layers](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/layers) directory. Many of the layers you are using here can be directly imported from this directory, I see a lot of commonality. I would suggest only adding new layers  (if not already present in `layers`) as a subdirectory - `layers/_<layer-type>`. `sub_modules.py` is a v1 convention

pytorch_forecasting/models/itransformer/_itransformer_v2.py
```diff
@@ -0,0 +1,201 @@
...
+        # B N E -> B N E
+        # the dimensions of embedded time series has been inverted
+        enc_out, attns = self.encoder(enc_out, attn_mask=None)
+
+        # B N E -> B N S -> B S N
+        dec_out = self.projector(enc_out).permute(0, 2, 1)[
+            :, :, :N
+        ]  # filter covariates
+
+        if self.use_norm:
```
PranavBhatP: `self.use_norm` is not required in the model code since it will be handled by the D1/D2 layers. Normalization and denomalization need not be handled here. Simply return `dec_out`.

pytorch_forecasting/models/itransformer/_itransformer_pkg_v2.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return {
+            "train": train_dataloader,
+            "val": val_dataloader,
+            "test": test_dataloader,
+            "data_module": train_datamodule,
+        }
+
+    @classmethod
+    def get_test_train_params(cls):
```
PranavBhatP, line 36: Can you add a few more test cases here?

PranavBhatP (changes requested):
Thanks for the PR @JATAYU000 . I've dropped some comments on the PR.

pytorch_forecasting/models/itransformer/_itransformer_v2.py
```diff
@@ -0,0 +1,201 @@
...
+from torch.optim import Optimizer
+
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class iTransformer(TslibBaseModel):
+    """
+    An implementation of iTransformer model for v2 of pytorch-forecasting.
+
+    Parameters
```
PranavBhatP, line 24: Docstring for model hyperparameters is missing.

pytorch_forecasting/models/itransformer/submodules.py
JATAYU000: I found some changes in some layers and let it be in `submodules` for now in draft pr, will fix this .

pytorch_forecasting/models/itransformer/_itransformer_v2.py
JATAYU000: Oh Thank you for pointing that out.

JATAYU000:
> I would suggest only adding new layers (if not already present in layers) as a subdirectory - layers/_<layer-type>

@PranavBhatP The `EncoderLayer` in `layers/_encoders/` requires `cross_attention`  , but `iTransformer` only needs `self_attention`.
Should I make `cross_attention` optional in the existing layer? or create a separate `EncoderLayer` for `iTransformer`?

fkiraly:
re layers, I would do as follows:

* if exact same layer is available in `layers`, reuse it
* add new layers in `layers`
* if layer with modification is needed, add it as a separate layer
* optionally - this PR but also can be later PR - check if multiple similar layers can be "unified" in a single layer with more parameters

PranavBhatP:
> optionally - this PR but also can be later PR - check if multiple similar layers can be "unified" in a single layer with more parameters

@fkiraly seems like a nice good first issue?

PranavBhatP:
@JATAYU000 any hurdles with fixing issues in the PR?

JATAYU000:
@PranavBhatP There are'nt any hurdles, I was busy with some other projects, I needed a review on the model and layer implementation since it updates the `TimeXer` layer to be a common TSLib layer, Have to add tests will do that.

pytorch_forecasting/layers/_encoders/_encoder.py
```diff
@@ -2,39 +2,50 @@
...
-import torch
 import torch.nn as nn
-import torch.nn.functional as F
 
 
 class Encoder(nn.Module):
     """
-    Encoder module for the TimeXer model.
+    Encoder module for Tslib models.
     Args:
```
phoeenniixx, line 11: please change this to numpydoc style docstring

pytorch_forecasting/layers/_encoders/_encoder_layer.py
```diff
@@ -2,36 +2,34 @@
...
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
 
 class EncoderLayer(nn.Module):
     """
-    Encoder layer for the TimeXer model.
+    Encoder layer for TsLib models.
     Args:
```
phoeenniixx, line 13: please change this to numpydoc style docstring

pytorch_forecasting/models/itransformer/_itransformer_pkg_v2.py
```diff
@@ -0,0 +1,76 @@
...
+        """Get test train params."""
+        from pytorch_forecasting.metrics import QuantileLoss
+
+        params = [
+            {},
+            dict(d_model=16, n_heads=2, e_layers=2, d_ff=64),
+            dict(
+                d_model=32,
```
phoeenniixx, line 44: should we try one param with `output_attetion=True` as well? to cover all possibilities?

agobbifbk:
This is misleading, timestamp are not processed by this layer :-) 
```python
enc_out = self.enc_embedding(x_enc, x_mark_enc)  # covariates (e.g timestamp) 
```
Moreover here I don't see any reference to the crossattention
```python
        self.encoder = Encoder(
            [
                EncoderLayer(
                    self_attention=AttentionLayer(
                        FullAttention(
                            False,
                            self.factor,
                            attention_dropout=self.dropout,
                            output_attention=True,
                        ),
                        self.d_model,
                        self.n_heads,
                    ),
                    d_model=self.d_model,
                    d_ff=self.d_ff,
                    dropout=self.dropout,
                    activation=self.activation,
                    output_attention=True,
                )
                for _ in range(self.e_layers)
            ],
            norm_layer=torch.nn.LayerNorm(self.d_model),
            output_attention=True,
        )
        if self.n_quantiles is not None:
            self.projector = nn.Linear(
                self.d_model, self.prediction_length * self.n_quantiles, bias=True
            )
        else:
            self.projector = nn.Linear(self.d_model, self.prediction_length, bias=True)
```

the Encoder layer has crossattention None by default
```
class EncoderLayer(nn.Module):
    def __init__(
        self,
        self_attention,
        cross_attention=None,
        d_model=512,
        d_ff=None,
        dropout=0.1,
        activation="relu",
        output_attention=False,
    ):
```
How the user can use this? Or I'm missing something?

JATAYU000:
> This is misleading, timestamp are not processed by this layer :-)

The comment was cut, it was supposed to be `# covariates (e.g timestamp) are embedded as tokens`

> Moreover here I don't see any reference to the cross_attention

`iTransformer` does not require to pass `cross_attention` since it defaults to None.
when it is passed it is included in the block, defaulting it to None was to make the layer flexible to the layers which doesn't need `cross_atention` as in `iTransformer` or any other inverse models implemented from `TsLib` in the future

If there is any other reason to not to default to None? I did not understand the exact problem could you please  elaborate a bit more @agobbifbk

PR #1997, opened 2025-11-30
[ENH] Efficient Attention Backend for TimeXer

### LINKED ISSUE #1990: [ENH] Sub-Optimal `FullAttention` Implementations Taken from TSLib
I've been observing the work being done for the `v2` release of `pytorch-forecasting`, including the implementation of the `TimeXer` model, taken from https://github.com/thuml/Time-Series-Library. 

Obviously, we all know that the TSLib is great for when it comes to releasing SOTA models to the public, but imo a lot of code in that original library is more or less outdated and un-optimized given what's possible in newer versions of torch. Why I'm highlighting this is that in TSLib torch is pinned at `1.7.1` (see [here](https://github.com/thuml/Time-Series-Library/blob/3846bfb81c8f3a86f154965819409d5a12a121c9/requirements.txt#L12C1-L12C13)), while in `pytorch-forecasting` minimum allowed version is at `torch==2.0.0`. 

Given this, imo there's no need to actually carry over the TSLib-s `torch.einsum()`-based attention computation to this project (see l66 of `layers/_attention/_full_attention.py` [here](https://github.com/sktime/pytorch-forecasting/blob/400b00f7c676ed869bcc026544743861112c1af7/pytorch_forecasting/layers/_attention/_full_attention.py#L66) and l67 of `/models/timexer/sub_modules.py` [here](https://github.com/sktime/pytorch-forecasting/blob/400b00f7c676ed869bcc026544743861112c1af7/pytorch_forecasting/models/timexer/sub_modules.py#L67) ).

The `.einsum()` computation is much more memory hungry and slower than for the `torch.nn.functional.scaled_dot_product_attention` now natively available in torch + the latter can use flash attn kernel if the user's GPU supports it. So I don't really see a reason to be using the slower implementation especially as *you don't need it*, because it seems you're not exposing the attention scores (see `output_attention=False` [here](https://github.com/sktime/pytorch-forecasting/blob/400b00f7c676ed869bcc026544743861112c1af7/pytorch_forecasting/models/timexer/_timexer.py#L275) in the `TimeXer` class) to the users anyways.

Besides the original TSLib modules have a lot of unused args kept in there for some specific models (e.g. the `tau` and `delta` in the `FullAttention` module are meant for the Non-Stationary Transformer specifically, and none of the more recent TSLib models use it). There's also a lot of other memory-related optimizations possible to the TSLib implementations anyways, such as using a single fused layer for Q,K,V projection and chunking them instead of separate projections and so on.

All of these things as of now seem to be carried over into your implementations of the models, so I think a lot could be improved ahead of the `v2` release so that this package can offer users *a major perf improvement over what's originally possible with TSLib models*.

I'm not really requesting a solution from your side. If this is something that *you'd allow me to pick up* as an issue to refactor the TimeXer class accordingly, I'd be very happy. I've worked with TimeXer for my own use cases extensively and found major speedup and memory savings after refactoring the original outdated TSLib attention implementations.

Another alternative possible: If the reasoning for keeping `torch.einsum()` based attn implementation is that you want to have the attention scores accessible, we could also refactor the Attention classes to have two backends, one could be `.einsum()`-based, able to return the attention scores (as you have now) and another would be the `torch.nn.functional.scaled_dot_product_attention`-based, offering optimized perf but w/o the ability to return the attention scores.

**Additional context**
I've done some benchmarking tests between the two attention methods discussed here for my purposes and can share the results with you as well. If I can take this issue on, would include some test modules (+ maybe some kind of benchmarking notebook) to demonstrate the speedup to you as well.

Big fan of your work in general, and I'd be happy if I can contribute 🫡

anasashb:
#### Reference Issues/PRs

Fixes #1990.

This Pull Request adds a `use_efficient_attention` boolean argument to the TimeXer model (v1 and v2 versions, both), which, if set to `=True`, switches to a more memory-efficient and faster attention implementation using `torch.nn.functional_scaled_dot_product_attention()` instead of the `torch.einsum()` solution inside the `FullAttention` class (v1 and v2 versions both). 

The newly introduced argument is currently set to False to keep the new feature completely backwards compatible.

Additionally, there's a very minor bugfix in the `PositionalEmbedding` class (v1 and v2 versions both), where a bug carried over from `tslib` used to define:

```python
pe.require_grad = False
```

In torch, the correct attribute for whether a tensor requires grad is called `.requires_grad`. This bug has also been fixed.

Reviewers should focus on the implementation of `_einsum_attention()` and `_efficient attention()` which are new private methods that `def forward()` of the `FullAttention` class calls to handle attention implementation. 

I did not make any other changes to the code, but if it works for you I could also:
- Remove some unused args such as `tau`, `delta`, `factor` scattered across the tslib code carried over here
- Add type annotations to the PyTorch code from tslib

Or if you'd be OK with those changes too, I can also open a separate PR.

Yes, in both: `tests/test_models/test_timxer.py` and `tests/test_models/test_timexer_v2.py`. These include new assertions in initialization tests, as well as parameterization of the `use_efficient_attention` for integration tests.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/timexer/_timexer_pkg.py
+++ b/pytorch_forecasting/models/timexer/_timexer_pkg.py
@@ -79,6 +79,15 @@ class TimeXer_pkg(_BasePtForecaster):
                     ),
                 ),
             },
+            {
+                "hidden_size": 32,
+                "patch_length": 1,
+                "n_heads": 4,
+                "e_layers": 1,
+                "d_ff": 32,
+                "dropout": 0.1,
+                "use_efficient_attention": True,
+            },
         ]
 
     @classmethod
--- a/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py
+++ b/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py
@@ -162,4 +162,13 @@ class TimeXer_pkg_v2(_BasePtForecasterV2):
                 ),
                 loss=QuantileLoss(quantiles=[0.1, 0.5, 0.9]),
             ),
+            dict(
+                hidden_size=32,
+                patch_length=1,
+                n_heads=4,
+                e_layers=1,
+                d_ff=32,
+                dropout=0.1,
+                use_efficient_attention=True,
+            ),
         ]


anasashb:
***Some more notes:***

1. The CI/CD pipe seems to partially fail but it looks like all of the errors are due to weight loading code for Temporal Fusion Transformer, not related to the feature in this PR 
2. As I said in the PR description:

> I did not make any other changes to the code, but if it works for you I could also:
> - Remove some unused args such as tau, delta, factor scattered across the tslib code carried over here
> - Add type annotations to the PyTorch code from tslib

If your team's ok with me addressing these, I could address these in this PR or open a new one, *whatever works for your team's code review process*

*p.s.* if anyone wants to benchmark the speed and memory consumption of the old and new attention backends, I can give you a script for that. I was not exactly sure if script such as that made sense to be included as part of the package itself

phoeenniixx:
> The CI/CD pipe seems to partially fail but it looks like all of the errors are due to weight loading code for Temporal Fusion Transformer, not related to the feature in this PR

Yes the issue is known although we still doesn't know the exact source (see #1998, and the discussion from the discord thread [here](https://discord.com/channels/1075852648688930887/1443988904054095932))

> If your team's ok with me addressing these, I could address these in this PR or open a new one, whatever works for your team's code review process

I'd prefer a new PR (stacked on this PR, or maybe after this PR is merged) to keep the "responsibilities" separate for both the PRs.
FYI @PranavBhatP @fkiraly @agobbifbk

fkiraly:
> if anyone wants to benchmark the speed and memory consumption of the old and new attention backends, I can give you a script for that. I was not exactly sure if script such as that made sense to be included as part of the package itself

hm, that feels extremely useful! Could you put that into `utils`, in a separate PR? We could use that for performance monitoring in the CI.

pytorch_forecasting/layers/_attention/_full_attention.py
```diff
@@ -35,7 +35,10 @@ class FullAttention(nn.Module):
         factor (int): Factor for scaling the attention scores.
         scale (float): Scaling factor for attention scores.
         attention_dropout (float): Dropout rate for attention scores.
-        output_attention (bool): Whether to output attention weights."""
+        output_attention (bool): Whether to output attention weights.
+        efficient_attention (bool): Whether to use torch's native efficient
```
phoeenniixx: Should it be `use_efficient_attention` here?
Also please explain what "efficient attention" means here and how is it different from `einsum_attention`.

Also please use numpydoc style docstrings. I think this part is from the older part of the code, that's why it's still not updated. Updating the style to numpydoc style would be greatly appreciated!

pytorch_forecasting/models/timexer/sub_modules.py
```diff
@@ -36,7 +36,10 @@ class FullAttention(nn.Module):
         factor (int): Factor for scaling the attention scores.
         scale (float): Scaling factor for attention scores.
         attention_dropout (float): Dropout rate for attention scores.
-        output_attention (bool): Whether to output attention weights."""
+        output_attention (bool): Whether to output attention weights.
+        efficient_attention (bool): Whether to use torch's native efficient
```
phoeenniixx: same comment as above!

phoeenniixx (changes requested):
Thanks a lot for this PR @anasashb ! This is really great
I have added some comments:
- related to the docstrings - I think this is from the older part of the code that's why it is still not using numpydoc style. I would really appreciate if you could use numpydoc style docstrings here. We still need to update the docstrings from the whole codebase :)
- Can you also add some fixtures in [_timexer_pkg](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg.py) and [_timexer_pkg_v2](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py). We are moving from the standalone tests to a unified test framework and now only just adding test fixtures and some configs would work to test the whole models. You can see some examples in the fixtures already present in the above files and you can also look at any other model, all models have this `pkg` class now which is used to test these models. All you need to do is just update the `get_base_test_params` (in case of v1`_timexer_pkg`) and `get_test_train_params` (in case of v2 `_timexer_pkg_v2`) to have the fixtures to test the new attention mechanism.

anasashb:
Thanks for the comments @phoeenniixx!

- - -

> Should it be use_efficient_attention here?

Great catch! Addressed for both `sub_modules.py` and `_full_attention.py` in https://github.com/sktime/pytorch-forecasting/pull/1997/commits/4009aa5ef918e550eae175a7d98d1444b8f83286

- - -

> Also please explain what "efficient attention" means here

Good point also, I added a better-detailed explanation of what to expect from the `use_efficient_attention` arg in both the `FullAttention` classes, as well as the `TimXer` classes (v1 and v2 both) in https://github.com/sktime/pytorch-forecasting/pull/1997/commits/c5bf10c833ef016084dfbcacbac9f326bea64d4c

- - -

> Can you also add some fixtures in [_timexer_pkg](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg.py) and [_timexer_pkg_v2](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py)

Addressed in: https://github.com/sktime/pytorch-forecasting/pull/1997/commits/c774ac53486fd3b641acbc463b6936dd340b0cc2

- - -

> really appreciate if you could use numpydoc style docstrings here

I haven't done this yet. I *can add them as the next commit* if you prefer so, but what I could also do is to begin working on a separate PR for this. I could also place an issue about it, or just directly start working on it, but the scope would be sort of  like cleanup for TimeXer code:

- Add docstrings across all used modules / classes used by TimeXer
- Switch the existing docstrings (including the one you pointed out) used by TimeXer to numpydocstyle
- Get rid of unused / carry-over args from `tslib`
- Add type annotations across all modules used by TimeXer

What do you think about this? I think this could be helpful for the developers on your side also, because much of the modules used by TimeXers are re-usable by a lot of other models in `tslib`, for example PatchTST and iTransformer  (btw I'd be very happy to try adding either PatchTST or iTransformer to this package once we're done with this work on TimeXer).

phoeenniixx:
> what I could also do is to begin working on a separate PR for this. I could also place an issue about it, or just directly start working on it, but the scope would be sort of like cleanup for TimeXer code

Thanks @anasashb! I really appreciate this! If you could do all this in a separate PR, it would really be helpful. And yes, please raise an issue as well so that tracking of the rework is easier.

Also pinging @PranavBhatP @agobbifbk for a review, since you have more experience with `tslib` models

agobbifbk:
Hello there, just a question:
```
self.endogenous_vars = endogenous_vars
self.exogenous_vars = exogenous_vars
```
These are something that the user needs to specify or are inherited from the d1-d2 layer definition? What is the difference?  

```
history_time_idx
```
What represents? the indexes between 0 an N where N is the context length? @phoeenniixx can you check this out? It seems that `DataEmbedding_inverted` can handle `x_mark` variable, usually in TSLIB this is related to categorical data, even if they are just apply a linear layer (no embeddings). In case the `EnEmbedding` already process the position information using the `EnEmbedding` layer no?

PranavBhatP:
> Hello there, just a question:
self.endogenous_vars = endogenous_vars
self.exogenous_vars = exogenous_vars
These are something that the user needs to specify or are inherited from the d1-d2 layer definition? What is the difference?

Hi @agobbifbk, I can answer this...I think these two attributes were initially added as overrides to the variables specified by D2 because D2 wasn't fully working when the model was ready. These attributes are redundant now and its removal is being tracked in issue - #2003.

phoeenniixx:
> It seems that `DataEmbedding_inverted` can handle `x_mark` variable, usually in TSLIB this is related to categorical data, even if they are just apply a linear layer (no embeddings). In case the `EnEmbedding` already process the position information using the `EnEmbedding` layer no?

Based on my understanding (@PranavBhatP please correct me if I am wrong)
- `DataEmbedding_inverted` handles exogenous data (see [here](https://github.com/sktime/pytorch-forecasting/blob/aae13bae82f99f9bf646770668c947ca456c10a7/pytorch_forecasting/models/timexer/_timexer_v2.py#L302)) and `EnEmbedding` handles endogenous data ( see [here](https://github.com/sktime/pytorch-forecasting/blob/aae13bae82f99f9bf646770668c947ca456c10a7/pytorch_forecasting/models/timexer/_timexer_v2.py#L301)). 
- `DataEmbedding_inverted` handles the whole series at once while `EnEmbedding` handles by creating patches (see [here](https://github.com/sktime/pytorch-forecasting/blob/aae13bae82f99f9bf646770668c947ca456c10a7/pytorch_forecasting/layers/_embeddings/_en_embedding.py#L45)).
- `DataEmbedding_inverted` concatenates `x_mark` with `x` (see [here](https://github.com/sktime/pytorch-forecasting/blob/aae13bae82f99f9bf646770668c947ca456c10a7/pytorch_forecasting/layers/_embeddings/_data_embedding.py#L36)) to tell the model the temporal feats of the exogenous variables while `EnEmbedding` uses position info (see [here](https://github.com/sktime/pytorch-forecasting/blob/aae13bae82f99f9bf646770668c947ca456c10a7/pytorch_forecasting/layers/_embeddings/_en_embedding.py#L48)) to tell the model the Order of the patches. So, I think they both have a different function?

I also looked at the implementation from `tslib`, they seem to follow the [same way](https://github.com/thuml/Time-Series-Library/blob/d403917f4a271e23ca17c277cce0d77035a33c7d/models/TimeXer.py#L167-L168)

Either way, I dont think that is related to the changes from this PR? 
We could discuss about this, and maybe raise an issue if you are sceptical about it?

phoeenniixx (approved):
Thanks @anasashb for the PR!
Looking forward to your contributions in future!

Issue #1840, opened 2025-05-20
[ENH] avoid "load from repo" with stallion data, replace in testing


fkiraly:
The `get_stallion` utility downloads the stallion data from the repository, each time it is called. This seems wasteful of download traffic, and ceases to work without an internet connection.

At least in testing, this should be replaced - perhaps by a synthetic dummy dataset of comparable format.

Further, for the user-facing case, we should consider moving it into the package - it is 1MB and 0.8MB after compression, so not too large - but also not minimal size.

RihaanBH-1810:
@fkiraly I would like to work on this issue, can I take this up ?

fkiraly:
@RihaanBH-1810, yes, much appreciated!

PR #1999, opened 2025-11-30
[ENH] Generate synthetic stallion data for testing.

RihaanBH-1810:
This PR solves #1840 by replacing the get_stallion_data which downloads the dataset with a function which generates a similar dataset for testing.

#### Reference Issues/PRs
#1840

Replace get_stallion_data usage in tests with get_stallion_dummy_data to avoid repeated network downloads and support offline CI.

Is this the right approach for generating synthetic data for tests ? 

No

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
   "torch >=2.0.0,!=2.0.1,<3.0.0",
   "lightning >=2.0.0,<2.7.0",
   "scipy >=1.8,<2.0",
-  "pandas >=1.3.0,<3.0.0",
+  "pandas >=1.3.0,<3.1.0",
   "scikit-learn >=1.2,<2.0",
   "scikit-base <0.14.0",
 ]
--- a/pytorch_forecasting/data/__init__.py
+++ b/pytorch_forecasting/data/__init__.py
@@ -1,8 +1,9 @@
 """
-Datasets, etc. for timeseries data.
+Utilities for time series dataset construction and preprocessing.
 
-Handling timeseries data is not trivial. It requires special treatment.
-This sub-package provides the necessary tools to abstracts the necessary work.
+This subpackage provides dataset classes, normalization and encoding
+utilities, and batching tools required to transform raw time series data
+into model-ready PyTorch datasets.
 """
 
 from pytorch_forecasting.data.encoders import (
--- a/pytorch_forecasting/data/encoders.py
+++ b/pytorch_forecasting/data/encoders.py
@@ -1293,7 +1293,14 @@ class GroupNormalizer(TorchNormalizer):
             params = norm
         else:
             try:
-                params = self.norm_.loc[groups].to_numpy()
+                indexer = groups
+                if (
+                    isinstance(groups, tuple)
+                    and len(groups) == 1
+                    and not isinstance(self.norm_.index, pd.MultiIndex)
+                ):
+                    indexer = groups[0]
+                params = self.norm_.loc[indexer].to_numpy()
             except (KeyError, TypeError):
                 params = np.asarray([self.missing_[name] for name in self.names])
         return params
--- a/pytorch_forecasting/layers/_embeddings/_data_embedding.py
+++ b/pytorch_forecasting/layers/_embeddings/_data_embedding.py
@@ -14,12 +14,19 @@ import torch.nn.functional as F
 class DataEmbedding_inverted(nn.Module):
     """
     Data embedding module for time series data.
-    Args:
-        c_in (int): Number of input features.
-        d_model (int): Dimension of the model.
-        embed_type (str): Type of embedding to use. Defaults to "fixed".
-        freq (str): Frequency of the time series data. Defaults to "h".
-        dropout (float): Dropout rate. Defaults to 0.1.
+
+    Parameters
+    ----------
+    c_in : int
+         Number of input features.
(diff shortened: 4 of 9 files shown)


phoeenniixx (changes requested):
Welcome to `pytorch-forecasting` @RihaanBH-1810!
Thanks for the PR, this is good!
Although, there are some genuine errors, can you please have a look at them? Please let us know if you need any help!

RihaanBH-1810:
> Welcome to `pytorch-forecasting` @RihaanBH-1810! Thanks for the PR, this is good! Although, there are some genuine errors/failures, can you please have a look at them? Please let us know if you need any help!

Hello, sorry I got caught up with another project that I was part of, now that it's over I'll get this done asap.

RihaanBH-1810:
@phoeenniixx I have fixed all the errors/failures can you please take a look at it.

phoeenniixx:
Why are we changing `encoders ` and `distribution`? I think they are out of the scope of the PR? The main focus was just to add synthetic data right? I dont understand the need of making changes to the above files? Can you please clarify this?

RihaanBH-1810:
@phoeenniixx I had to make changes to the `encoders` and `distribution` because they were running into some edge cases while running the tests with the new dummy data, most of the changes are just covering those edge cases.

RihaanBH-1810:
@phoeenniixx I've reverted the changes in the `encoders` and `distribution` the only changes made to are in the function generating the dataset.
Although all tests are passing, I still have a doubt

Initially I had changed this to Agency_02  because of the way the data was being generated, at least one batch would have all of the entries as Agency_01 so the overwrite would not work.

in `/pytorch_forecasting/tests/test_data/test_timeseries.py`
```
@pytest.mark.parametrize(
    "value,variable,target",
    [
        (1.0, "price_regular", "encoder"),
        (1.0, "price_regular", "all"),
        (1.0, "price_regular", "decoder"),
        ("Agency_01", "agency", "all"),
        ("Agency_01", "agency", "decoder"),
    ],
)
def test_overwrite_values(test_dataset, value, variable, target):
    dataset = deepcopy(test_dataset)
```

so I just changed it to Agency_02  in the test and it worked but then since I was asked to not change anything in the tests 

The only way I could think of fixing this was just to limit the number of the Agency_01 pairs in the dataset and it worked, but I don't know if this is the right approach.

in `pytorch_forecasting/data/examples.py`
```
    agency_list = [f"Agency_{i:02d}" for i in range(1, 61)]
    sku_list = [f"SKU_{j:02d}" for j in range(1, 36)]
    selected_pairs = [(a, s) for a in agency_list for s in sku_list]
    rng.shuffle(selected_pairs)

    first_pair_batch = [p for p in selected_pairs if p[0] == "Agency_01"]
    other_pairs = [p for p in selected_pairs if p[0] != "Agency_01"]
    selected_pairs = first_pair_batch[:2] + other_pairs
    rng.shuffle(selected_pairs)

    agencies = np.array([p[0] for p in selected_pairs])
    skus = np.array([p[1] for p in selected_pairs])
 ```

could you please clarify.

**Comment by phoeenniixx** [comment]
I am sorry but I am not able to understand your issue, can you please elaborate on this? Please clearly mention the error messages and the issues you are facing...

**Comment by RihaanBH-1810** [comment]
@phoeenniixx 
The `test_overwrite_values[Agency_01-agency-all]` and `test_overwrite_values[Agency_01-agency-decoder]` tests were failing:
```
>               assert not changed, f"Output {name} should change"
E               AssertionError: Output encoder_cat should change
E               assert not tensor(True)
```

Now this is because in the `examples.py` file in the `get_stallion_dummy_data` function I was creating agency sku pairs like this 
```
    agency_list = [f"Agency_{i:02d}" for i in range(1, 61)]
    sku_list = [f"SKU_{j:02d}" for j in range(1, 36)]
    selected_pairs = [(a, s) for a in agency_list for s in sku_list]
    rng.shuffle(selected_pairs)

    agencies = np.array([p[0] for p in selected_pairs])
    skus = np.array([p[1] for p in selected_pairs])
```

`TimeSeriesDataSet` internally sorts time series alphabetically by `group_ids (["agency", "sku"])`, so `Agency_01` series always come first. With 60 agencies x 35 SKUs, `Agency_01` had 35 time series. Each time series produces roughly 13 sliding window samples, giving Agency_01 about 455 samples total. Since the sequential dataloader (used when train=False) serves samples in order and the batch size is 64, the entire first batch ends up containing only Agency_01 samples. When the test calls `set_overwrite_values("Agency_01", variable="agency", ...)`, every sample in that batch already has agency = Agency_01, so nothing actually changes. The test then compares the overwritten batch against the control batch, finds them identical, and fails.

So to fix this I did something like this :
```
first_pair_batch = [p for p in selected_pairs if p[0] == "Agency_01"]
other_pairs = [p for p in selected_pairs if p[0] != "Agency_01"]
selected_pairs = first_pair_batch[:2] + other_pairs
rng.shuffle(selected_pairs)
```

Now Agency_01 will only pair with 2 SKUs so the first batch will not have all `Agency_01` values so now the overwrite test passes, but then this did not seem like a proper solution to me since I am only limiting `Agency_01` pairs to pass the test, If I were to do the same for all other agencies then it would fail the `test_raise_short_encoder_length` test.

ps: I did use ai to format a few bits of the message so that I could communicate better, please excuse me.

PR #2000, opened 2025-12-01
[BUG] Solve the unpickling error in weight Loading

### LINKED ISSUE #1998: [BUG] `Unpickling Error` in weight loading
**Describe the bug**

There is an `Unpickling Error` in weight loading of `pytorch-forecasting` because the weights are not loaded correctly. 
The issue seems to be something related to `torch` which is leading to fail in CI/CD pipeline. 
The exact source of the bug is still unknown.

This is a traceback copied from the tests:
```
TestAllPtForecasters.test_integration[TemporalFusionTransformer-base_params-2-QuantileLoss] _

self = <pytorch_forecasting.tests.test_all_estimators.TestAllPtForecasters object at 0x11fdfeb50>
object_pkg = <class 'pytorch_forecasting.models.temporal_fusion_transformer._tft_pkg.TemporalFusionTransformer_pkg'>
trainer_kwargs = {'causal_attention': False, 'loss': QuantileLoss(quantiles=[0.02, 0.1, 0.25, 0.5, 0.75, 0.9, 0.98])}
tmp_path = PosixPath('/private/var/folders/p6/nlmq3k8146990kpkxl73mq340000gn/T/pytest-of-runner/pytest-0/test_integration_TemporalFusio0')

    def test_integration(
        self,
        object_pkg,
        trainer_kwargs,
        tmp_path,
    ):
        """Test end-to-end workflow training and predicting using model."""
    
        object_class = object_pkg.get_cls()
        dataloaders = object_pkg._get_test_dataloaders_from(trainer_kwargs)
    
>       _integration(object_class, dataloaders, tmp_path, **trainer_kwargs)

pytorch_forecasting/tests/test_all_estimators.py:386: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pytorch_forecasting/tests/test_all_estimators.py:320: in _integration
    net = estimator_cls.load_from_checkpoint(
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/lightning/pytorch/utilities/model_helpers.py:130: in wrapper
    return self.method(cls_type, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/lightning/pytorch/core/module.py:1781: in load_from_checkpoint
    loaded = _load_from_checkpoint(
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/lightning/pytorch/core/saving.py:65: in _load_from_checkpoint
    checkpoint = pl_load(checkpoint_path, map_location=map_location, weights_only=weights_only)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/lightning/fabric/utilities/cloud_io.py:73: in _load
    return torch.load(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

f = <fsspec.implementations.local.LocalFileOpener object at 0x1483299f0>
map_location = <function _default_map_location at 0x11eb3ef20>
pickle_module = None, weights_only = True, mmap = False
pickle_load_args = {'encoding': 'utf-8'}
_get_wo_message = <function load.<locals>._get_wo_message at 0x12b1089a0>
weights_only_not_set = True, true_values = ['1', 'y', 'yes', 'true']
force_weights_only_load = False, force_no_weights_only_load = False

    def load(
        f: FileLike,
        map_location: MAP_LOCATION = None,
        pickle_module: Any = None,
        *,
        weights_only: Optional[bool] = None,
        mmap: Optional[bool] = None,
        **pickle_load_args: Any,
    ) -> Any:
        # Reference: https://github.com/pytorch/pytorch/issues/54354
        # The first line of this docstring overrides the one Sphinx generates for the
        # documentation. We need it so that Sphinx doesn't leak `pickle`s path from
        # the build environment (e.g. `<module 'pickle' from '/leaked/path').
    
        """load(f, map_location=None, pickle_module=pickle, *, weights_only=True, mmap=None, **pickle_load_args)
    
        Loads an object saved with :func:`torch.save` from a file.
    
        :func:`torch.load` uses Python's unpickling facilities but treats storages,
        which underlie tensors, specially. They are first deserialized on the
        CPU and are then moved to the device they were saved from. If this fails
        (e.g. because the run time system doesn't have certain devices), an exception
        is raised. However, storages can be dynamically remapped to an alternative
        set of devices using the :attr:`map_location` argument.
    
        If :attr:`map_location` is a callable, it will be called once for each serialized
        storage with two arguments: storage and location. The storage argument
        will be the initial deserialization of the storage, residing on the CPU.
        Each serialized storage has a location tag associated with it which
        identifies the device it was saved from, and this tag is the second
        argument passed to :attr:`map_location`. The builtin location tags are ``'cpu'``
        for CPU tensors and ``'cuda:device_id'`` (e.g. ``'cuda:2'``) for CUDA tensors.
        :attr:`map_location` should return either ``None`` or a storage. If
        :attr:`map_location` returns a storage, it will be used as the final deserialized
        object, already moved to the right device. Otherwise, :func:`torch.load` will
        fall back to the default behavior, as if :attr:`map_location` wasn't specified.
    
        If :attr:`map_location` is a :class:`torch.device` object or a string containing
        a device tag, it indicates the location where all tensors should be loaded.
    
        Otherwise, if :attr:`map_location` is a dict, it will be used to remap location tags
        appearing in the file (keys), to ones that specify where to put the
        storages (values).
    
        User extensions can register their own location tags and tagging and
        deserialization methods using :func:`torch.serialization.register_package`.
    
        See :ref:`layout-control` for more advanced tools to manipulate a checkpoint.
    
        Args:
            f: a file-like object (has to implement :meth:`read`, :meth:`readline`, :meth:`tell`, and :meth:`seek`),
                or a string or os.PathLike object containing a file name
            map_location: a function, :class:`torch.device`, string or a dict specifying how to remap storage
                locations
            pickle_module: module used for unpickling metadata and objects (has to
                match the :attr:`pickle_module` used to serialize file)
            weights_only: Indicates whether unpickler should be restricted to
                loading only tensors, primitive types, dictionaries
                and any types added via :func:`torch.serialization.add_safe_globals`.
                See :ref:`weights-only` for more details.
            mmap: Indicates whether the file should be mapped rather than loading all the storages into memory.
                Typically, tensor storages in the file will first be moved from disk to CPU memory, after which they
                are moved to the location that they were tagged with when saving, or specified by ``map_location``. This
                second step is a no-op if the final location is CPU. When the ``mmap`` flag is set, instead of copying the
                tensor storages from disk to CPU memory in the first step, ``f`` is mapped, which means tensor storages
                will be lazily loaded when their data is accessed.
            pickle_load_args: (Python 3 only) optional keyword arguments passed over to
                :func:`pickle_module.load` and :func:`pickle_module.Unpickler`, e.g.,
                :attr:`errors=...`.
    
        .. warning::
            :func:`torch.load()` unless `weights_only` parameter is set to `True`,
            uses ``pickle`` module implicitly, which is known to be insecure.
            It is possible to construct malicious pickle data which will execute arbitrary code
            during unpickling. Never load data that could have come from an untrusted
            source in an unsafe mode, or that could have been tampered with. **Only load data you trust**.
    
        .. note::
            When you call :func:`torch.load()` on a file which contains GPU tensors, those tensors
            will be loaded to GPU by default. You can call ``torch.load(.., map_location='cpu')``
            and then :meth:`load_state_dict` to avoid GPU RAM surge when loading a model checkpoint.
    
        .. note::
            By default, we decode byte strings as ``utf-8``.  This is to avoid a common error
            case ``UnicodeDecodeError: 'ascii' codec can't decode byte 0x...``
            when loading files saved by Python 2 in Python 3.  If this default
            is incorrect, you may use an extra :attr:`encoding` keyword argument to specify how
            these objects should be loaded, e.g., :attr:`encoding='latin1'` decodes them
            to strings using ``latin1`` encoding, and :attr:`encoding='bytes'` keeps them
            as byte arrays which can be decoded later with ``byte_array.decode(...)``.
    
        Example:
            >>> # xdoctest: +SKIP("undefined filepaths")
            >>> torch.load("tensors.pt", weights_only=True)
            # Load all tensors onto the CPU
            >>> torch.load(
            ...     "tensors.pt",
            ...     map_location=torch.device("cpu"),
            ...     weights_only=True,
            ... )
            # Load all tensors onto the CPU, using a function
            >>> torch.load(
            ...     "tensors.pt",
            ...     map_location=lambda storage, loc: storage,
            ...     weights_only=True,
            ... )
            # Load all tensors onto GPU 1
            >>> torch.load(
            ...     "tensors.pt",
            ...     map_location=lambda storage, loc: storage.cuda(1),
            ...     weights_only=True,
            ... )  # type: ignore[attr-defined]
            # Map tensors from GPU 1 to GPU 0
            >>> torch.load(
            ...     "tensors.pt",
            ...     map_location={"cuda:1": "cuda:0"},
            ...     weights_only=True,
            ... )
            # Load tensor from io.BytesIO object
            # Loading from a buffer setting weights_only=False, warning this can be unsafe
            >>> with open("tensor.pt", "rb") as f:
            ...     buffer = io.BytesIO(f.read())
            >>> torch.load(buffer, weights_only=False)
            # Load a module with 'ascii' encoding for unpickling
            # Loading from a module setting weights_only=False, warning this can be unsafe
            >>> torch.load("module.pt", encoding="ascii", weights_only=False)
        """
        torch._C._log_api_usage_once("torch.load")
        DOCS_MESSAGE = (
            "\n\nCheck the documentation of torch.load to learn more about types accepted by default with "
            "weights_only https://pytorch.org/docs/stable/generated/torch.load.html."
        )
    
        def _get_wo_message(message: str) -> str:
            unsafe_global_pattern = r"GLOBAL (\S+) was not an allowed global by default."
            has_unsafe_global = re.search(unsafe_global_pattern, message) is not None
            blocklist_pattern = r"whose module (\S+) is blocked"
            has_blocklist = re.search(blocklist_pattern, message) is not None
            import_pattern = r"(\S+) must be (\S+) to load"
            has_import = re.search(import_pattern, message) is not None
            if has_unsafe_global:
                updated_message = (
                    "Weights only load failed. This file can still be loaded, to do so you have two options, "
                    "\033[1mdo those steps only if you trust the source of the checkpoint\033[0m. "
                    f"\n\t(1) {UNSAFE_MESSAGE}\n\t(2) Alternatively, to load with `weights_only=True` please check "
                    "the recommended steps in the following error message.\n\tWeightsUnpickler error: "
                    + message
                )
            else:
                if has_import:
                    return f"Weights only load failed. {message}\n {UNSAFE_MESSAGE}\n"
                else:
                    updated_message = f"Weights only load failed. {UNSAFE_MESSAGE}\n"
                    if not has_blocklist:
                        updated_message += (
                            "Please file an issue with the following so that we can make "
                            "`weights_only=True` compatible with your use case: WeightsUnpickler error: "
                        )
                updated_message += "\n\n" + message
            return updated_message + DOCS_MESSAGE
    
        weights_only_not_set = weights_only is None
    
        if weights_only_not_set:
            weights_only = _default_to_weights_only(pickle_module)
    
        true_values = ["1", "y", "yes", "true"]
        # Add ability to force safe only or non-safe weight loads via environment variables
        force_weights_only_load = (
            os.getenv("TORCH_FORCE_WEIGHTS_ONLY_LOAD", "0") in true_values
        )
        force_no_weights_only_load = (
            os.getenv("TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD", "0") in true_values
        )
    
        if force_weights_only_load and force_no_weights_only_load:
            raise RuntimeError(
                "Only one of `TORCH_FORCE_WEIGHTS_ONLY_LOAD` or `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD` "
                "should be set, but both were set."
            )
        elif force_weights_only_load:
            weights_only = True
        elif force_no_weights_only_load:
            # TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD can only override if callsite did not explicitly set weights_only
            if weights_only_not_set:
                warnings.warn(
                    "Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the"
                    "`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.",
                    UserWarning,
                    stacklevel=2,
                )
                weights_only = False
    
        if weights_only:
            if pickle_module is not None:
                raise RuntimeError(
                    "Can not safely load weights when explicit pickle_module is specified"
                )
        else:
            if pickle_module is None:
                pickle_module = pickle
    
        # make flipping default BC-compatible
        if mmap is None:
            from torch.utils.serialization import config
    
            mmap = config.load.mmap
    
        _check_dill_version(pickle_module)
    
        if "encoding" not in pickle_load_args.keys():
            pickle_load_args["encoding"] = "utf-8"
    
        with _open_file_like(f, "rb") as opened_file:
            if _is_zipfile(opened_file):
                # The zipfile reader is going to advance the current file position.
                # If we want to actually tail call to torch.jit.load, we need to
                # reset back to the original position.
                orig_position = opened_file.tell()
                overall_storage = None
                with _open_zipfile_reader(opened_file) as opened_zipfile:
                    if _is_torchscript_zip(opened_zipfile):
                        warnings.warn(
                            "'torch.load' received a zip file that looks like a TorchScript archive"
                            " dispatching to 'torch.jit.load' (call 'torch.jit.load' directly to"
                            " silence this warning)",
                            UserWarning,
                        )
                        if weights_only:
                            raise RuntimeError(
                                "Cannot use ``weights_only=True`` with TorchScript archives passed to "
                                "``torch.load``. " + UNSAFE_MESSAGE
                            )
                        opened_file.seek(orig_position)
                        return torch.jit.load(opened_file, map_location=map_location)
                    if mmap:
                        if not _is_path(f):
                            raise ValueError(
                                "f must be a file path in order to use the mmap argument"
                            )
                        size = os.path.getsize(f)
                        if not IS_WINDOWS:
                            shared = get_default_mmap_options() == MAP_SHARED
                        else:
                            shared = False
                        overall_storage = torch.UntypedStorage.from_file(
                            os.fspath(f), shared, size
                        )
                    if weights_only:
                        try:
                            return _load(
                                opened_zipfile,
                                map_location,
                                _weights_only_unpickler,
                                overall_storage=overall_storage,
                                **pickle_load_args,
                            )
                        except pickle.UnpicklingError as e:
>                           raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
E                           _pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
E                           	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
E                           	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
E                           	WeightsUnpickler error: Unsupported global: GLOBAL pytorch_forecasting.data.encoders.GroupNormalizer was not an allowed global by default. Please use `torch.serialization.add_safe_globals([pytorch_forecasting.data.encoders.GroupNormalizer])` or the `torch.serialization.safe_globals([pytorch_forecasting.data.encoders.GroupNormalizer])` context manager to allowlist this global if you trust this class/function.
E                           
E                           Check the documentation of torch.load to learn more about types accepted by default with weights_only https://pytorch.org/docs/stable/generated/torch.load.html.

/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/torch/serialization.py:1529: UnpicklingError
```

**To Reproduce**

Please look at the test logs here: https://github.com/sktime/pytorch-forecasting/actions/runs/19799950889/job/56725769642?pr=1997

phoeenniixx:
There is an Unpickling Error in weight loading of `pytorch-forecasting-v1` because the weights are not loaded correctly.
The issue seems to be something related to torch which is leading to fail in CI/CD pipeline.

The issue surfaced after the latest `lightning` update where they exposed the `weights_only` param and let `torch` handle it itself.

Fixes #1998

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,4 @@
+# Code of Conduct
+
+All contributors, maintainers, and participants are expected to follow the
+`sktime` Code of Conduct found on our [website](https://www.sktime.net/en/latest/get_involved/code_of_conduct.html).
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,5 @@
+# Governance
+
+`pytorch-forecasting` is governed by the `sktime` community.
+
+You can find our governance guidelines on our [website](https://www.sktime.net/en/latest/get_involved/governance.html).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,7 +106,7 @@ dev = [
 
 # docs - dependencies for building the documentation
 docs = [
-  "sphinx>3.2,<8.2.4",
+  "sphinx>3.2,<9.1.1",
   "pydata-sphinx-theme",
   "nbsphinx",
   "pandoc",


fkiraly:
hm, not yet, right?

phoeenniixx:
Yes, this PR is incomplete until we find a way to pass `weights_only` to `tuner`.

fkiraly:
would overriding `tuner` work? Feels dangerous, but if we are forced to?

pytorch_forecasting/models/temporal_fusion_transformer/tuning.py
```diff
@@ -23,6 +25,24 @@
 optuna_logger = logging.getLogger("optuna")
 
 
+# todo: Remove this class once lightning allows the pass of weights_only to tuner
+class NewTuner(Tuner):
```
fkiraly: I would make this private

pytorch_forecasting/models/temporal_fusion_transformer/tuning.py
```diff
@@ -23,6 +25,24 @@
...
+    def lr_find(self, *args, **kwargs):
+        original_load = torch.load
+
+        @functools.wraps(original_load)
+        def new_load(*load_args, **load_kwargs):
+            load_kwargs["weights_only"] = False
+            return original_load(*load_args, **load_kwargs)
+
+        torch.load = new_load
```
fkiraly: can we replace this with an explicit if/else check? based on `_check_soft_dependencis` perhaps

pytorch_forecasting/models/temporal_fusion_transformer/tuning.py
```diff
@@ -23,6 +25,24 @@
...
+
+        try:
+            return super().lr_find(*args, **kwargs)
+        finally:
+            torch.load = original_load
```
fkiraly: doing this feels dangerous, because this has side effects on everything!

I would not do this. Is there a way we can override only in a call, or override a method in a class?

phoeenniixx: I have wrapped `trainer.fit` here then.

phoeenniixx: Wrapping around `trainer.fit` didn't work as `lr_find` calls `trainer._checkpoint_connector.restore`  after `fit` that overrode the `weights_only` param from `fit`, so now I have wrapped `load_checkpoint` instead (its deeper in the traceback call - close to `torch.load`, but still in the `lightning` layer). I think this should solve the issue?

phoeenniixx:
Hi @fkiraly, can you please review this PR?

pytorch_forecasting/models/temporal_fusion_transformer/tuning.py
```diff
@@ -23,6 +27,26 @@
 optuna_logger = logging.getLogger("optuna")
 
 
+# todo: Remove this class once lightning allows the pass of weights_only to tuner
```
fkiraly: can you add a todo witih version number, so it gets checked with every minor release? E.g., todo next.minor.version.
Release managers typically search for version numbers of the release to be done.

phoeenniixx: But the problem is we are not sure about the version number, this class would have to stay as is until `lightning` supports `weights_only` in `tuner`, and based on current response, I am not if they plan to do so in recent future

phoeenniixx: See https://github.com/Lightning-AI/pytorch-lightning/issues/21401

fkiraly: same for other todos

tests/test_models/test_deepar.py
```diff
@@ -90,7 +90,9 @@ def _integration(
         test_outputs = trainer.test(net, dataloaders=test_dataloader)
         assert len(test_outputs) > 0
         # check loading
-        net = DeepAR.load_from_checkpoint(trainer.checkpoint_callback.best_model_path)
+        net = DeepAR.load_from_checkpoint(
```
fkiraly: I think this might still be a breaking change? Why does the current loader patter no longer work.

I think current behaviour should be maintained, so currently passing tests should still pass.

fkiraly (changes requested):
Requests regarding release management preparation, and potentially breaking change in default of `load_from_checkpoint`.

tests/test_models/test_deepar.py
phoeenniixx: So,should I add `_check_soft_dep` at all points? Sure... I will do so

pytorch_forecasting/models/temporal_fusion_transformer/tuning.py
fkiraly: > But the problem is we are not sure about the version number, this class would have to stay as is until `lightning` supports `weights_only` in `tuner`, and based on current response, I am not if they plan to do so in recent future

The todo should be to check.
That is, at 1.6.0, check whether the issue is resolved (with description how). If yes, action the instructions. If no, increment the version number to 1.7.0.

This ensures we check at every release whether the change condition is satisfied (rather than forgetting about it)

tests/test_models/test_deepar.py
fkiraly: no, I think the code in the tests should still work. This is v1 code that should not experience a breaking change.

Meaning, all v1 tests should pass after the update.

The version checking distinction should happen under the hood, only.

phoeenniixx: Hi @fkiraly !
I implemented a wrapper, I think this should work? what do you think?

docs/source/tutorials/deepar.ipynb
```diff
@@ -348,7 +348,12 @@
    ],
    "source": [
     "# find optimal learning rate\n",
-    "from lightning.pytorch.tuner import Tuner\n",
+    "# from lightning.pytorch.tuner import Tuner\n",
```
fkiraly: can we avoid this change?

phoeenniixx: Sorry, I am not sure we can avoid this. The issue is this change is at `lightning` level and not at `ptf` level. I had to override `Tuner` of `lightning` to overcome this issue. Unless there are upstream changes, I don’t see a clean way to avoid this.

Do you see any alternative approach here? I may be missing something.😅

Happy new year btw!!! :)

phoeenniixx:
Hi @fkiraly, I am not sure we can avoid this breaking change as it has propagated from the `lightning` upstream. Do you see any way to cleanly avoid this?

I think we need to clearly mention it in the release that for tuning in v1, we need to use `_NewTuner` in place of `lightning.Tuner` until the issues are handled upstream?

phoeenniixx:
> * can we remove the try/finally from the `Tuner`? I think we know the exact conditions. Generally, try/except should always be avoided if instead a condition can be used.

I didnt know if this `self._trainer` is called to some other places as well internally (eg by some other methods of `Tuner` which a user may call). So, I thought it would be better to take it to the earlier state - without this "patched and forced" `weights_only` arg. I mean our code should work fine without this block, but it will **_fix_** that this exact `trainer` instance would always have `weights_only` as `False` for `load_checkpoint` and it may not what a user want? I could not exactly think of any specific case, but still to be on a safer side, I thought of this as a fail safe as we are messing with an instance which actually doesnot lie in a layer we have full control over (its not `ptf` but `lightning` layer).

So, Even if `lr_find` were to fail due whatever reason (like Out of Memory, or any bugs) , it would not break the whole `Tuner` instance of the user.

Does this makes sense or am I just overthinking here?😅
what do you think should I remove this?

fkiraly:
>  So, I thought it would be better to take it to the earlier state - without this "patched and forced" weights_only arg

If that is your thinking, should there not be another `super` call after `finally` restores the original method?

phoeenniixx:
> > So, I thought it would be better to take it to the earlier state - without this "patched and forced" weights_only arg
> 
> If that is your thinking, should there not be another `super` call after `finally` restores the original method?

I am not sure about that. I mean we are patching `self._trainer`, so we should restore it's functionality, and we don't need to call `super()` for that. It is already taken from the parent class. So, that's what I am trying to restore in `finally` block.
```python
strategy.load_checkpoint=original_load_checkpoint # where strategy = self._trainer.strategy
```

phoeenniixx:
> But that could lead to execution without a failure where something actually has failed under the hood?

That's why we are not using `except` block here. If the code inside `try` fails, it would still throw an error but only after executing the `finally` block (i.e, after restoring `load_checkpoint`)...

fkiraly:
ah yes, you are right!

PR #2012, opened 2025-12-20
[ENH] Refactor N-BEATS blocks to separate KAN logic

### LINKED ISSUE #2011: [ENH] separate `NBeats` blocks from kan
Currently, the NBeats blocks are "fused" with the KAN design, it is simply an if/else in most cases, based on the boolean parameter `use_kan`.

This is not good practice, so we should split the NBeats layers, `NBEATSBlock`, `NBEATSSeasonalBlock`, and `NBEATSTrendBlock` into two - an `NBEATSBlock` and an `NBEATSBlockKAN`, etc.

The `kan_params` should become explicit kwargs of `__init__` rather than being passed as a `dict`.

Current use of `use_kan=False` should map to `NBEATSBlock`, and use of `use_kan=True` to a new `NBEATSBlockKAN`, with parameters passed explicitly.

khenm:
#### Reference Issues/PRs
Fixes #2011

---
This PR refactors the N-BEATS architecture to explicitly separate standard MLP blocks from KAN-based blocks, removing the previous `if/else` dependency on the `use_kan` flag.

* **Split Classes:** Created distinct classes for all variants to ensure clean separation of concerns:
    * `NBEATSBlock` / `NBEATSBlockKAN`
    * `NBEATSSeasonalBlock` / `NBEATSSeasonalBlockKAN`
    * `NBEATSTrendBlock` / `NBEATSTrendBlockKAN`
* **Mixin Pattern:** Implemented `SeasonalMixin` and `TrendMixin` to handle the shared mathematical logic (Fourier series and Polynomial generation) between standard and KAN blocks. 
* Unpacked the `kan_params` dictionary into explicit `__init__` arguments (e.g., `num`, `k`, `noise_scale`) in the KAN classes.
* Updated `NBeatsKAN` to correctly initialize these new specific block classes based on the `stack_types` argument.
---
* Please verify that `SeasonalMixin` and `TrendMixin` correctly capture the shared logic without unwanted side effects.
* Check if the use of `**kan_kwargs` in the subclasses correctly passes parameters up to `NBEATSBlockKAN`.
* I restored `F.relu` in `NBEATSGenericBlockKAN` to ensure consistency with the standard generic block. 
---
Yes, I added `test_nbeats_kan_integration` which covers the following:
1. Verifies Init -> Train -> Save -> Load -> Predict flows working together.
2. Asserts that KAN-specific hyperparameters (e.g., `num=10`) are correctly saved to `hparams` and restored upon checkpoint loading.
3. Asserts that the model actually initializes the correct class types (e.g., `NBEATSTrendBlockKAN`) when `stack_types` are requested.
---
#### Any other comments?
No comment.

---
#### PR checklist
- [x] Added/modified tests
- [x] Used pre-commit hooks.

### DIFF
--- a/pytorch_forecasting/layers/_nbeats/_blocks.py
+++ b/pytorch_forecasting/layers/_nbeats/_blocks.py
@@ -31,9 +31,7 @@ class SeasonalMixin:
         Forecast side of the seasonality basis matrix.
     """
 
-    def _init_seasonal(
-        self, backcast_length, forecast_length, thetas_dim, min_period
-    ):
+    def _init_seasonal(self, backcast_length, forecast_length, thetas_dim, min_period):
         """
         Initialize seasonal backcast and forecast coefficients.
         """
@@ -322,9 +320,7 @@ class NBEATSBlockKAN(nn.Module):
         for _ in range(num_block_layers - 1):
             if self.dropout > 0:
                 layers.append(nn.Dropout(p=self.dropout))
-            layers.append(
-                KANLayer(in_dim=units, out_dim=units, **self.kan_params)
-            )
+            layers.append(KANLayer(in_dim=units, out_dim=units, **self.kan_params))
         self.fc = nn.Sequential(*layers)
 
         # theta layers used by subclasses
--- a/tests/test_models/test_nbeats_kan.py
+++ b/tests/test_models/test_nbeats_kan.py
@@ -1,12 +1,13 @@
-
-import pytest
 import shutil
+
 import lightning.pytorch as pl
 from lightning.pytorch.callbacks import EarlyStopping
 from lightning.pytorch.loggers import TensorBoardLogger
+import pytest
 
 from pytorch_forecasting.models.nbeats._nbeatskan import NBeatsKAN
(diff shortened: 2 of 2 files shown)


khenm:
Hi! I want to follow up on my comment in #2011.

I went ahead and implemented the refactor to separate the KAN logic from the standard N-BEATS. 

Please review this PR for me and let me know if you'd like any adjustments!

docs/source/tutorials/ar.ipynb
```diff
@@ -187,10 +187,16 @@
     "    max_prediction_length=prediction_length,\n",
     ")\n",
     "\n",
-    "validation = TimeSeriesDataSet.from_dataset(training, data, min_prediction_idx=training_cutoff + 1)\n",
```
fkiraly, line 190: why are there changes in this notebook? It seems unrelated. Please revert.

khenm: Thanks for pointing it out for me. And yes, these changes are unintended. Im quite new to using pre-commit hooks in this workflow, and I didnt realize the hooks had formatted the notebook when I commited. I have reverted those changes and push the clean version. It should be correct now.

PR #2028, opened 2026-01-22
[DOC] Clarify RMSE implementation and reduction logic #1541

### LINKED ISSUE #1541: RMSE defined as MSE
https://github.com/jdb78/pytorch-forecasting/blob/68a0eb5f1701801142ce976fa50305b29507845a/pytorch_forecasting/metrics/point.py#L137C1-L149C20

Ds0uz4:
#### Reference Issues/PRs

Fixes #1541.

This PR clears up the ambiguity in the RMSE docstring identified in #1541 by explaining how the computation is actually carried out. The documentation now makes it clear that the loss method only computes the per-element squared error, and that the averaging and square root are applied later during the reduction step using the sqrt-mean strategy.

Verify that the updated docstring clearly explains the delegation of the square root operation to the reduction phase.

No. This is a documentation-only change ([DOC]) and does not modify the functional logic of the code.
#### Any other comments?

This clarification will help prevent future confusion for users who examine the source code and expect to see an explicit torch.sqrt call within the loss method.

#### PR checklist

- [ ] Added/modified tests
- [x] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with pre-commit install.
  To run hooks independent of commit, execute pre-commit run --all-files

### DIFF
--- a/pytorch_forecasting/metrics/point.py
+++ b/pytorch_forecasting/metrics/point.py
@@ -158,8 +158,8 @@ class RMSE(MultiHorizonMetric):
     the squared error.
     """
 
-    def _init_(self, reduction="sqrt-mean", **kwargs):
-        super()._init_(reduction=reduction, **kwargs)
+    def __init__(self, reduction="sqrt-mean", **kwargs):
+        super().__init__(reduction=reduction, **kwargs)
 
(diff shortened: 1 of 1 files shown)


phoeenniixx (approved):
Thanks a lot @Ds0uz4 !

PR #2029, opened 2026-01-22
[MNT] Add CI step with pinned dependencies as of Nov 2025

phoeenniixx:
This PR  adds CI step with pinned dependencies as of Nov 2025 just before the `lightning 2.6` release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,13 +115,12 @@ jobs:
         run: python -m pytest
 
   test-deps-2025:
-    name: no-softdeps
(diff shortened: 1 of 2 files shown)


fkiraly:
Great! We expect this to fail, since https://github.com/sktime/pytorch-forecasting/pull/2027 is not merged, right? So the failure validates this works as expected.

fkiraly (approved):

phoeenniixx:
Yes exactly, I merged this branch into #2027, and it seems to pass all the tests. This means the fix is working now.

PR #2041, opened 2026-01-28
[ENH] Add LSTM with native multi-target forecasting support

### LINKED ISSUE #2025: [ENH][DESIGN] Native multivariate support for LSTMModel in pytorch_forecasting
## Context

`TimeSeriesDataSet` supports multi-target forecasting, but the current LSTMModel example (and implementation) is effectively univariate, which can lead to confusion and runtime shape mismatches when multiple targets are provided.

An older PR (#1449) attempted to address this, but it is now stale relative to `main`.

## Goal

Clarify and/or extend LSTMModel support such that behavior is explicit and consistent:
- Either clearly documented as univariate only, or
- Provide native multi-target (multivariate) support

## Proposed Approach

1. Review #1449 against current `main` to identify:
   - What logic is still applicable
   - What no longer applies due to recent refactors
2. Define a minimal, backward-compatible API for multivariate targets, including:
   - Decoder output dimensionality
   - Loss handling
   - Target normalization
3. Decide whether documentation clarification should be done in parallel

## Questions for Maintainers

- Is native multivariate support for LSTMModel desired in principle?
- If yes, should this be a single model with `n_targets > 1` or separate heads?
- Should documentation clarification be merged independently?

Happy to proceed with implementation once aligned.

Varshith-Yadav:
#### Reference Issues/PRs
- fixes #2025 
- Related to discussion on multivariate LSTM support and stale PR #1449
- Addresses confusion around univariate-only LSTM examples discussed in the community thread

This PR adds a **minimal LSTM model** with **native multi-target (multivariate) forecasting support** to the core codebase.

Key points:
- Implements an encoder–decoder LSTM compatible with `AutoRegressiveBaseModel`
- Supports `n_targets >= 1`, with `n_targets = 1` as a special case for backward compatibility
- Preserves structured outputs during training to ensure correct interaction with `MultiLoss`
- Ensures full decoder-horizon predictions during training to avoid shape mismatches
- Adds comprehensive tests covering forward pass, loss computation, backward pass, and prediction length

The model is intentionally simple and distinct from `xLSTM`, serving as a clear baseline forecaster aligned with existing dataset and loss abstractions.

- [x] API design and naming of the baseline LSTM model
- [x] Correct handling of multi-target outputs with `MultiLoss`
- [x] Decoder behavior across the full prediction horizon
- [x] Test coverage and backward compatibility for the univariate case

Yes. New tests were added to cover:
- Univariate forward pass (regression test)
- Multi-target forward pass
- Multi-target loss computation
- Backward pass smoke test
- Prediction length correctness

These tests ensure both correctness and protection against future regressions.

#### Any other comments?

- All tests pass locally.
- Pre-commit hooks were run with `pre-commit run --all-files`.
- `nbqa-black` may fail locally on Python 3.12.5 due to a known Black AST issue; CI should cover formatting on a supported Python version.

Happy to iterate on naming, scope, or documentation based on reviewer feedback.

### DIFF
--- a/pytorch_forecasting/__init__.py
+++ b/pytorch_forecasting/__init__.py
@@ -41,6 +41,7 @@ from pytorch_forecasting.models import (
     BaseModelWithCovariates,
     DecoderMLP,
     DeepAR,
+    LSTMModel,
     MultiEmbedding,
     NBeats,
     NBeatsKAN,
--- a/pytorch_forecasting/models/lstm/__init__.py
+++ b/pytorch_forecasting/models/lstm/__init__.py
@@ -3,6 +3,4 @@
 from pytorch_forecasting.models.lstm._lstm import LSTMModel
 from pytorch_forecasting.models.lstm._lstm_pkg import LSTMModel_pkg
 
-LSTMModel._pkg = staticmethod(lambda: LSTMModel_pkg)
-
 __all__ = ["LSTMModel", "LSTMModel_pkg"]
--- a/pytorch_forecasting/models/lstm/_lstm.py
+++ b/pytorch_forecasting/models/lstm/_lstm.py
@@ -32,6 +32,13 @@ class LSTMModel(AutoRegressiveBaseModel):
         **kwargs: Additional arguments passed to BaseModel.
     """
 
+    @classmethod
+    def _pkg(cls):
+        """Package containing the model."""
+        from pytorch_forecasting.models.lstm._lstm_pkg import LSTMModel_pkg
+
+        return LSTMModel_pkg
+
     def __init__(
         self,
         target: str | list[str],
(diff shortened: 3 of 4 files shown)


pytorch_forecasting/models/lstm/_lstm_pkg.py
```diff
@@ -0,0 +1,3 @@
+from pytorch_forecasting.models.lstm._lstm import LSTMModel
+
+LSTMModel_pkg = LSTMModel
```
phoeenniixx: I think you should add test fixtures here, please look at other models' pkgs for examples!
Some examples: [DeeAR_pkg](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/deepar/_deepar_pkg.py), [NBEATS_pkg](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nbeats/_nbeats_pkg.py)

Varshith-Yadav: Thanks for the pointer!

I’ll align _lstm_pkg.py with the existing patterns used in ``DeepAR_pkg`` and ``NBEATS_pkg`` and add the appropriate test fixtures there. I’ll push an update shortly.

Please let me know if there’s anything specific you’d like the fixtures to cover beyond the standard instantiation and smoke checks

PR #2043, opened 2026-01-31
[ENH] Add N-BEATS v2 model and tests

### LINKED ISSUE #2034: [ENH] Add v2 interface support for N-BEATS
Referencing #1992 

This issue tracks adding v2 interface support for the N-BEATS model.

The model currently does not have a v2 package. The plan is to interface the existing implementation following the v2 pattern used by DLinear, SAMformer, and TFT.

CC: @fkiraly

PalakB09:
#### Reference Issues/PRs

Fixes #2034 

Adds a v2 implementation of the N-BEATS model built on the BaseModel v2 / tslib API. Previously, N-BEATS did not have a v2 implementation. This change is additive and includes tests.

A test module `test_nbeats_v2.py` was added covering:
- Forward pass output shapes
- Presence and shape of decomposition outputs (trend and seasonality)
- Training step and backward pass with MASE
- Package container registry methods

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/nbeats/_nbeats_v2.py
+++ b/pytorch_forecasting/models/nbeats/_nbeats_v2.py
@@ -15,7 +15,7 @@ from pytorch_forecasting.metrics import MASE, Metric
 from pytorch_forecasting.models.base._base_model_v2 import BaseModel
 
 
-class NBEATS(BaseModel):
+class NBEATS_v2(BaseModel):
     """
     Initialize NBEATS Model.
 
--- a/pytorch_forecasting/models/nbeats/_nbeats_v2_pkg.py
+++ b/pytorch_forecasting/models/nbeats/_nbeats_v2_pkg.py
@@ -5,7 +5,7 @@ class NBEATS_pkg_v2(Base_pkg):
     """NBEATS v2 package container."""
 
     _tags = {
-        "info:name": "NBEATS",
+        "info:name": "NBEATS_v2",
         "authors": ["PalakB09"],
         "capability:exogenous": False,
         "capability:multivariate": False,
@@ -17,9 +17,9 @@ class NBEATS_pkg_v2(Base_pkg):
     @classmethod
     def get_cls(cls):
         """Get model class."""
-        from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS
+        from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS_v2
 
-        return NBEATS
+        return NBEATS_v2
 
     @classmethod
     def get_datamodule_cls(cls):
--- a/pytorch_forecasting/models/nbeats/__init__.py
+++ b/pytorch_forecasting/models/nbeats/__init__.py
@@ -14,7 +14,7 @@ from pytorch_forecasting.models.nbeats._grid_callback import GridUpdateCallback
 from pytorch_forecasting.models.nbeats._nbeats import NBeats
 from pytorch_forecasting.models.nbeats._nbeats_adapter import NBeatsAdapter
 from pytorch_forecasting.models.nbeats._nbeats_pkg import NBeats_pkg
-from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS
+from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS_v2
 from pytorch_forecasting.models.nbeats._nbeats_v2_pkg import NBEATS_pkg_v2
 from pytorch_forecasting.models.nbeats._nbeatskan import NBeatsKAN
 from pytorch_forecasting.models.nbeats._nbeatskan_pkg import NBeatsKAN_pkg
@@ -30,6 +30,6 @@ __all__ = [
     "NBeatsAdapter",
     "GridUpdateCallback",
     # v2 exports
-    "NBEATS",
+    "NBEATS_v2",
     "NBEATS_pkg_v2",
 ]
--- a/tests/test_models/test_nbeats_v2.py
+++ b/tests/test_models/test_nbeats_v2.py
@@ -1,7 +1,7 @@
 import torch
 
 from pytorch_forecasting.metrics import MAE
-from pytorch_forecasting.models.nbeats import NBEATS
+from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS_v2
 from pytorch_forecasting.models.nbeats._nbeats_v2_pkg import NBEATS_pkg_v2
 
 
@@ -17,7 +17,7 @@ def test_nbeats_v2_forward_shapes():
     context_length = 10
     prediction_length = 5
 
-    model = NBEATS(metadata=_make_metadata(context_length, prediction_length))
+    model = NBEATS_v2(metadata=_make_metadata(context_length, prediction_length))
     model.eval()
 
     x = {
@@ -43,7 +43,7 @@ def test_nbeats_v2_forward_shapes():
 
 
 def test_nbeats_v2_decomposition_outputs_exist():
-    model = NBEATS(metadata=_make_metadata())
+    model = NBEATS_v2(metadata=_make_metadata())
     model.eval()
 
     batch_size = 2
@@ -64,7 +64,7 @@ def test_nbeats_v2_training_step_and_backward():
     context_length = 8
     prediction_length = 4
 
-    model = NBEATS(
+    model = NBEATS_v2(
         loss=MAE(),
         metadata=_make_metadata(context_length, prediction_length),
     )
@@ -88,9 +88,10 @@ def test_nbeats_v2_training_step_and_backward():
 
 def test_nbeats_v2_pkg_get_cls():
     cls = NBEATS_pkg_v2.get_cls()
-    assert cls is NBEATS
+    assert cls is NBEATS_v2
 
 
 def test_nbeats_v2_pkg_test_params():
-    params = NBEATS_pkg_v2.get_test_params()
-    assert isinstance(params, dict)
+    params = NBEATS_pkg_v2.get_test_train_params()
+    assert isinstance(params, list)
+    assert isinstance(params[0], dict)


pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+class NBEATS_v2(TslibBaseModel):
+    @classmethod
+    def _pkg(cls):
+        from pytorch_forecasting.models.nbeats._nbeats_v2_pkg import (
+            NBEATS_v2_pkg_v2,
+        )
+
+        return NBEATS_v2_pkg_v2
```
phoeenniixx: You can simply name the `pkg` class as `NBEATS_pkg_v2`, no need of adding v2 twice!

pytorch_forecasting/models/nbeats/_nbeats_v2_pkg.py
```diff
@@ -0,0 +1,46 @@
...
+    def get_cls(cls):
+        from pytorch_forecasting.models.nbeats._nbeats_v2 import NBEATS_v2
+
+        return NBEATS_v2
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
```
phoeenniixx: I think you should use `EncoderDecoderDataModule` given this is from the original v1 model pool?
`TslibDataModule` is mainly for the models brought from `tslib` package.

phoeenniixx:
Thanks @PalakB09 for the PR and welcome to `pytorch-forecasting`!
Although, I know the PR is WiP, I just wanted to ask one question: did you write the source code from scratch or took from v1? As we were trying to understand the gaps that are still present between v1 and v2 and if you used the v1 source  code, can you please let us know the issues you faced while migrating the model?

PalakB09:
Thanks for the update @phoeenniixx, I’m working on the changes.

The v2 implementation wasn’t written from scratch, I used the v1 source code. 
I mainly faced two issues. The first was around MASE handling, which required overriding parts of the training and validation flow. The second was restructuring how losses and metrics are defined, since v1 allowed more implicit behaviour, while v2 enforces more stricter. This part was taken mostly from v1, but the core model logic remains unchanged.

Currently, CI is failing due to a doctest issue in check_estimator(NBEATS). However, when I run
`pytest pytorch_forecasting/tests/test_doctest.py -v` locally, all doctests pass successfully. All the rest of the tests are passing.
Could you please guide me on how to approach or debug this failure?

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,339 @@
...
+    MAPE,
+    MASE,
+    RMSE,
+    SMAPE,
+    Metric,
+)
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class NBEATS_v2(TslibBaseModel):
```
PranavBhatP: Please use the v2 `BaseModel` here as well.

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,339 @@
...
+        self.sharing = sharing or [True, True]
+        self.expansion_coefficient_lengths = expansion_coefficient_lengths or [3, 7]
+
+        self.dropout = dropout
+        self.backcast_loss_ratio = backcast_loss_ratio
+
+        self._init_network()
+
+    @classmethod
+    def from_dataset(cls, dataset: TimeSeriesDataSet, **kwargs):
```
PranavBhatP: I see that you are using the v1 `TimeSeriesDataset` for a proposed v2 model. `TimeSeriesDataset` is the dataset abstraction for v1 and will be no longer used going into v2 - use the `TimeSeries` object.

PranavBhatP: Just a generic comment - you can try to look at existing implementations of v2 models to understand how the model layer API works (we have models like TimeXer which have been migrated sucessfully from v1 to v2), try to see the code and then apply the same for `NBeats`. This implementation still has a lot of changes to be made.

PalakB09:
Thanks for the inputs, understood my mistake, working on it

PalakB09:
Hi @phoeenniixx @PranavBhatP 
Thanks for reviewing!
I have addressed the changes. Earlier, I was importing the wrong base class, which I have now fixed, and the CI is green.
The only issue I ran into was with quantile loss since the model outputs a point forecast, so I added a small helper to expand the prediction for quantile loss. Please let me know if this approach looks correct.
Thank you

phoeenniixx:
If the model performs only performs point predictions, how are you sure it is working on `QuantileLoss` correctly? I think even the v1 model just worked on point prediction losses. 
Can you please clarify this?

PalakB09:
Yes, you are right, v1 is using point forecasting only and not quantile loss.
I initially added it while exploring probabilistic variants in some libraries, but yes, that would also require changing the metrics.
I have removed the quantile loss and am now following the v1 behavior.

PalakB09:
Hi @phoeenniixx 
Can you please take a look and suggest changes
Thank you!

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,176 @@
...
+    def _pkg(cls):
+        from pytorch_forecasting.models.nbeats._nbeats_v2_pkg import (
+            NBEATS_pkg_v2,
+        )
+
+        return NBEATS_pkg_v2
+
+    def __init__(
+        self,
+        *,
```
phoeenniixx, line 84: why are you adding this?

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,176 @@
...
+        dropout: float = 0.0,
+        loss: Metric | None = None,
+        logging_metrics: list[nn.Module] | None = None,
+        optimizer: Optimizer | str | None = "adam",
+        optimizer_params: dict | None = None,
+        lr_scheduler: str | None = None,
+        lr_scheduler_params: dict | None = None,
+        metadata: dict | None = None,
+        **kwargs,
+    ):
```
phoeenniixx, line 100: please add the docstrings in numpydocstyle explaining the model and the params

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,176 @@
...
+        for stack_id, stack_type in enumerate(self.stack_types):
+            if self.sharing[stack_id]:
+                block = self._make_block(stack_id, stack_type)
+                for _ in range(self.num_blocks[stack_id]):
+                    self.net_blocks.append(block)
+            else:
+                for _ in range(self.num_blocks[stack_id]):
+                    self.net_blocks.append(self._make_block(stack_id, stack_type))
+
+    def forward(self, x: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
```
phoeenniixx, line 174: Add the docstrings please

pytorch_forecasting/models/nbeats/_nbeats_v2.py
```diff
@@ -0,0 +1,176 @@
...
+from pytorch_forecasting.layers._nbeats._blocks import (
+    NBEATSGenericBlock,
+    NBEATSSeasonalBlock,
+    NBEATSTrendBlock,
+)
+from pytorch_forecasting.metrics import MASE, Metric
+from pytorch_forecasting.models.base._base_model_v2 import BaseModel
+
+
+class NBEATS(BaseModel):
```
phoeenniixx: Rename it as `NBEATS_v2`

PalakB09: I am facing this error if I name it as NBEATS_v2 and the pkg as nbeats_v2_pkg. It requires the package name to be NBEATS_v2_pkg_v2.
Also I checked tft implementation, they aren't importing the model in init.py. Should I do it the same way? But the name is still not including v2 in it 


```text
def test_pkg_linkage(self, object_pkg, object_class):
    """Test that the package is linked correctly."""
    # check name method
    msg = (
        f"Package {object_pkg}.name() does not match class "
        f"name {object_class.__name__}. "
        "The expected package name is "
        f"{object_class.__name__}_pkg."
    )
    assert object_pkg.name() == object_class.__name__, msg

    # check naming convention
    msg = (
        f"Package {object_pkg.__name__} does not match class "
        f"name {object_class.__name__}. "
        "The expected package name is "
        f"{object_class.__name__}_pkg."
    )
>   assert object_pkg.__name__ == object_class.__name__ + "_pkg_v2", msg
E   AssertionError: Package NBEATS_pkg_v2 does not match class name NBEATS_v2. The expected package name is NBEATS_v2_pkg.
E   assert 'NBEATS_pkg_v2' == 'NBEATS_v2_pkg_v2'
E
E     - NBEATS_v2_pkg_v2
E     ?        ---
E     + NBEATS_pkg_v2

File: pytorch_forecasting/models/nbeats/_nbeats_v2.py (hunk 5 above)
Comment (PalakB09) [reply pytorch_forecasting/models/nbeats/_nbeats_v2.py:84]: Updated

**Comment by PalakB09** [comment]
Hi @phoeenniixx 
I have added the docstrings. Could you please confirm regarding the naming part?
Thank you!

**Comment by PalakB09** [comment]
Hi @phoeenniixx and @PranavBhatP
I have applied the naming changes now that the naming convention PR has been merged.
Could you please take a look?

Thank you!

File: pytorch_forecasting/models/nbeats/_nbeats_v2.py (hunk 5 above)
Comment (phoeenniixx) [reply pytorch_forecasting/models/nbeats/_nbeats_v2.py:84]: I meant this `*`?

File: pytorch_forecasting/models/nbeats/_nbeats_v2.py [hunk 9]
```diff
@@ -0,0 +1,245 @@
...
+        optimizer: Optimizer | str | None = "adam",
+        optimizer_params: dict | None = None,
+        lr_scheduler: str | None = None,
+        lr_scheduler_params: dict | None = None,
+        metadata: dict | None = None,
+        **kwargs,
+    ):
+        """Initialize the NBEATS_v2 model."""
+        if loss is None:
+            loss = MASE()
```
phoeenniixx, line 103: What kind of losses this model can handle? Like Point prediction losses, distribution, quantile (or all)?
I think we should add a check about this?

phoeenniixx (changes requested):
Thanks! 
I think  you should first collect all the params sent from `__init__` into `self`.
```python
self.param=param
```
( i could see params are `loss` is not collected at all)
Then, if the param needs some changes, then you should define a new `self._param` and then use that.
```python
self._param = op(self.param) # any operation on self.param
```

PR #2044, opened 2026-02-01
[MNT] Bump pandas version to `pandas 3.0`

phoeenniixx:
Bumps `pandas` bound to allow `pandas 3.0`.

Draft to see failures and passes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
   "torch >=2.0.0,!=2.0.1,<3.0.0",
   "lightning >=2.0.0,<2.7.0",
   "scipy >=1.8,<2.0",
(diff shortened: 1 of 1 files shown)


fkiraly:
also failing at freq M? Is this ever used explicitly, or only in example data generation?

fkiraly:
I think it is only in test data generation and not core code - would you agree?

If so, then https://github.com/sktime/pytorch-forecasting/pull/2045 should fix the failures.

phoeenniixx:
> I think it is only in test data generation and not core code - would you agree?
> 
> If so, then #2045 should fix the failures.

I think so. As much as i remember its only `data_scenarios`

fkiraly:
there are a few more, see #2045

PR #2045, opened 2026-02-02
[MNT] `pandas 3` compatibility - test data

fkiraly:
Fixes `pandas 3` incompatibility with tests. Core code does not seem to be affected by `pandas 3` changes, only an instance in the data generated for tests that used the deprecated `"M"` frequency.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# Contributing guide
+
+You can find our contributing guide on our [website](https://www.sktime.net/en/latest/get_involved/contributing.html).


phoeenniixx:
I think we need to make some more changes - to supress these warnings:
```
Warning: For backward compatibility, 'str' dtypes are included by select_dtypes when 'object' dtype is specified. This behavior is deprecated and will be removed in a future version. Explicitly pass 'str' to `include` to select them, or to `exclude` to remove them and silence this warning.
```

fkiraly:
well, but at this point in time they are just warnings, no? Of course a fix is preferred above no fix, but it would not hit users as long as we safety-bound at MINOR versions.

phoeenniixx:
Yeah no problem, I will find these issues, and raise a PR afterwards :)

fkiraly:
hmm, less failures but now some remaining related to `GroupNormalizer`

phoeenniixx:
The error is originating from this line:
https://github.com/sktime/pytorch-forecasting/blob/71252890405f3b15b7edf5b9645d87d1d980faef/pytorch_forecasting/data/encoders.py#L1296

I was looking at `pandas` repo, there I can see they enforced `strict=True` in internal `zip()` calls in many parts of the package recently. (see [here](https://github.com/pandas-dev/pandas/pulls?page=1&q=is%3Apr+is%3Aclosed+zip) and the [issue](https://github.com/pandas-dev/pandas/issues/62434)). Can this be the reason?

fkiraly:
possibly? Anway to make it compatible with multiple `pandas` versions?

phoeenniixx:
I am not sure. I am a little confused, I am not able to understand why this is failing exactly. I need to dig deeper and understand what is happening. If you want we can merge this fix (just not bump the `pandas` version). I think the `zip` issue would need some time from my side to find out what exactly is happening

fkiraly:
ok, reverted the version bump and will merge this as a partial fix then.

PR #2049, opened 2026-02-03
[ENH] Add v1 network extension template

### LINKED ISSUE #1953: [ENH] extension templates for neural network classes
As suggested in a recent tech meeting, we should add extension templates for the v1 networks and network packages API, to enable users to add new networks and get them tested immediately in the CI.

This could be modelled upon the extension template in `sktime`
https://github.com/sktime/sktime/tree/main/extension_templates

although, unlike in `sktime`, adding a neural network will mean adding at least two files - a `lightning` model, and a package class with tags. Thus, the extension template should probably be not a single file, but a folder, which mimicks the structure of the folder within the `models` module.

Saptami191:
Fixes #1953

#### What does this implement/fix?

This PR adds a minimal **v1 model extension template** at:
`pytorch_forecasting/models/extension_templates/v1_network_template/`.

I created this because existing v1 models follow two different patterns (single-file like `baseline.py`  vs folder-based like `rnn/`), which is confusing for new contributors.

The template includes:
* `model.py`: minimal `BaseModel` subclass with required methods.
* `package.py`: minimal package container with `_tags` and `get_cls()` for CI discovery.
* `README.md`: short instructions for contributors.

This only adds a template for v1 and does not modify existing models.
Did you add any tests?
No ,this PR only adds a contributor template.

#### PR checklist

* [x] PR title starts with [ENH]
* [ ] Added/modified tests
* [x] Used pre-commit hooks


Saptami191:
happy to fix id any changes required

pytorch_forecasting/models/extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,30 @@
...
+    Minimal template model for contributors.
+    """
+
+    def forward(self, x: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
+        """
+        Minimal placeholder forward pass.
+        Users should replace this with their own implementation.
+        """
+        last_values = x["encoder_target"][:, -1]
+        prediction = last_values[:, None]
```
phoeenniixx: I think you should just add the pointers (comments) how the `forward` and other functions should work. We should not place code here. Look at `sktime` [extension templates](https://github.com/sktime/sktime/tree/main/extension_templates) for reference

pytorch_forecasting/models/extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,30 @@
...
+import torch
+
+from pytorch_forecasting.models import BaseModel
+
+
+class ExampleNetwork(BaseModel):
+    """
+    Minimal template model for contributors.
+    """
+
```
phoeenniixx: You should also add `__init__()`, `_pkg()` and `from_dataset()`  here. Look at v1 models like [deepar](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/deepar/_deepar.py) for reference

pytorch_forecasting/models/extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,43 @@
...
+
+    This is REQUIRED for CI discovery and testing.
+    """
+
+    _tags = {
+        "info:name": "CustomNetwork",
+        # adjust these tags for your actual model
+        "info:pred_type": ["point"],
+        "info:y_type": ["numeric"],
+        "capability:exogenous": True,
```
phoeenniixx: Here you should add the comments stating what each tag mean and what are the values each tag can take here.
Also, the name of `_pkg` class should match the model name. If you are using `ExampleNetwork`, then it should be `ExampleNetwork_pkg` and the `"info:name"` tag should be `ExampleNetwork`.

Also, I think there are some tags missing here, you should add them as well.

pytorch_forecasting/models/extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def get_cls(cls):
+        """REQUIRED: return the actual model class."""
+        from .model import ExampleNetwork
+
+        return ExampleNetwork
+
+    @classmethod
+    def get_base_test_params(cls):
```
phoeenniixx: This is not optional, these are test fixtures, so they are required. Also, add the `_get_test_dataloaders_from()` here as well.

pytorch_forecasting/models/extension_templates/v1_network_template/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. Copy this folder to `pytorch_forecasting/models/` and rename it for your model.
+2. Implement your model logic in `model.py`.
+3. Update `_tags` in `package.py` to reflect your model’s actual capabilities.
+4. Ensure `get_cls()` returns your model class.
+
+## Notes
+
+- This template targets **v1 only** (it does not cover the v2 API).
+- `get_base_test_params()` is **optional** — most models can simply return `[{}]`
```
phoeenniixx: It is not optional and it cannot return just `[{}]`. It should have the fixtures to check all the possible edge cases etc

phoeenniixx (changes requested):
Thanks @Saptami191 for the PR! 
I have few more suggestions in addition to the below comments:
- Please revert the changes to `stallion.ipynb`.
- Please move the templates outside of the `pytorch_forecasting/models` to the main repo level. `pytorch_forecasting` folder should contains only the implementations that will use the extension templates you are working on, the collection of extension templates should live at the main repo level.

Please have a look at `sktime` [extension templates](https://github.com/sktime/sktime/tree/main/extension_templates). These templates should work as a boiler plate for the contributors, so please add comments to specify how the functions etc should be written here and how the tags should be set etc.

The `sktime` extension templates donot have a folder struct like `ptf`, but I think they should give you a pretty good idea how the templates should look like!

Saptami191:
@phoeenniixx  i have moved the v1 network template to the top level extension_templates/ directory as u requested and updated the README file to clarify it's  usage  and required test fixtures, and _get_test_dataloaders_from().
happy to make further changes if needed.

Saptami191:
@phoeenniixx the changes request is solved , i think it is now mergable

docs/source/tutorials/stallion.ipynb
phoeenniixx, line 1: Please revert this change

extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,150 @@
...
+
+        REQUIRED:
+        - Call super().__init__(**kwargs)
+
+        TYPICAL PATTERN:
+        - Save hyperparameters with self.save_hyperparameters()
+        - Initialize layers (e.g., RNN, Transformer, MLP, etc.)
+        """
+
+        self.save_hyperparameters()
```
phoeenniixx, line 49: explain why this is called and what are the args of this function. and add pointers like the naming conventions and that `self.params` are not to be overwritten. (see sktime [templates](https://github.com/sktime/sktime/blob/2ae9c6f3e76392a80b266f05998593ba8deadaa5/extension_templates/forecasting_simple.py#L134-L147) to what i mean)

extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @classmethod
+    def _pkg(cls):
+        """
+        REQUIRED for v1 models.
+
+        Returns the package container class that defines metadata (_tags)
+        and test fixtures.
+        """
+        from .package import ExampleNetwork_pkg
```
phoeenniixx, line 64: Add the pointer that it should be in the same folder.
also the name of `py` file should be same as the `pkg` class and should be private. Something like this:
```python
from pytorch_forecasting.models.ExampleNetwork._ExampleNetwork_pkg import ExampleNetwork_pkg
````

File: extension_templates/v1_network_template/model.py [hunk 8]
```diff
@@ -0,0 +1,150 @@
...
+            return self.to_network_output(prediction=your_prediction_tensor)
+
+        The shape of prediction should typically be:
+        (batch_size, decoder_length, target_dim)
+        """
+
+        # TODO: replace with your actual model logic
+        raise NotImplementedError("Implement forward() in your custom model")
+
+    def to_prediction(
```
phoeenniixx, line 130: it might not be required every time. YOu could add pointers when you need to override it. I dont think it is always required. (Same for `to_quantiles`)

extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,78 @@
...
+    This is required for CI discovery, registration, and testing of v1 models.
+    """
+
+    _tags = {
+        # Human-readable model name (MUST match the model class)
+        "info:name": "ExampleNetwork",
+        # What type of predictions this model produces:
+        # - ["point"]        → deterministic forecasts
+        # - ["quantile"]     → probabilistic quantiles
+        # - ["distribution"] → full predictive distribution
```
phoeenniixx, line 24: The naming is wrong. its `["distr"]`

extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,78 @@
...
+        "info:pred_type": ["point"],
+        # What type of target the model supports:
+        # Typically: ["numeric"]
```
phoeenniixx, line 27: its not always just `["numeric"]` it can be `["category"]` as well for categorcial losses. please add that as well

extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,78 @@
...
+        # 1 = lightweight, 5 = very heavy
+        "info:compute": 2,
+        # GitHub usernames of contributors
+        "authors": ["your-github-handle"],
+    }
+
+    @classmethod
+    def get_cls(cls):
+        """Return the actual Lightning model class."""
+        from .model import ExampleNetwork
```
phoeenniixx, line 53: dont use `.model` please use the complete name `from pytorch-forecasting.models.ExampleNetwork`

extension_templates/v1_network_template/package.py
```diff
@@ -0,0 +1,78 @@
...
+        Return minimal test parameters for CI.
+
+        This must return parameters that construct valid model instances.
+        Multiple parameter sets are recommended to cover edge cases.
+        """
+        return [
+            {"hidden_size": 8},
+            {"hidden_size": 16},
+        ]
+
```
phoeenniixx, line 69: Please look at `sktime` [extension templates](https://github.com/sktime/sktime/blob/2ae9c6f3e76392a80b266f05998593ba8deadaa5/extension_templates/forecasting_simple.py#L227-L275) and add exaclty how the params should be written

phoeenniixx (changes requested):
Thanks! I have given some suggestions above.
The extension templates not only act as boiler plate but also as a guide how the estimators should be implemented, in that case i feel this still lacks enough info and pointers. Please look at some sktime examples to understand how to add those pointers. You might need to look at some implementations of `v1` in `pytorch-forecasting` to understand the trends and how to implements these models. Then it would be easy to add these pointers :)

PR #2052, opened 2026-02-13
[ENH] Deepar v2 final fixes

### LINKED ISSUE #2007: [ENH] Add v2 interface support for DeepAR

Soham-47:
Fixes #2007 

Implemented the V2 architecture for the DeepAR model,improved scaling & distribution handling and fixed the index and value errors.


Soham-47:
hii @phoeenniixx , could you please take a look at this PR when convenient

PR #2053, opened 2026-02-14
[MNT] Ensure full pandas 3 compatibility

phoeenniixx:
This PR adds the complete compatibility for pandas 3. Followed by #2045


phoeenniixx:
The issue was `groups` was sent as `tuple` and `pandas 3` uses tuples for multi-index only. Previously even standard-index was sent as `tuple` and this fails in `pandas 3`.

fkiraly (changes requested):
Why `pandas` why.

Could you also attempt to remove the try/except? The construction looks too brittle now. We should try to replace it by an exact condition if it does not cause too much overhead in terms of compute (we should check that).

phoeenniixx:
Hmm, that would mean changing the legacy code, I would look into it....

PR #2055, opened 2026-02-16
[MNT] migrate test-deps-2025 to use uv date cutoffs

### LINKED ISSUE #2054: [MNT] migrate `test-deps-2025` to use `uv` date cutoffs
Instead of defining an explicit dependency set for 2025 timestamp regression testing, we should use `uv`'s dedicated functionality for this use case, namely the `UV_EXCLUDE_NEWER` environment variable.

Also see https://github.com/sktime/sktime/pull/9383

RUPESH-KUMAR01:
#### Reference Issues/PRs
Fixes #2054 
Related: https://github.com/sktime/sktime/pull/9383

Replaces the explicit dependency set with uv’s UV_EXCLUDE_NEWER to install dependencies from a time-frozen snapshot.

Correct usage of UV_EXCLUDE_NEWER for timestamp regression testing

Cutoff date choice (2025-01-01) and whether it matches project expectations

No new tests added, as this change only affects CI configuration.

#### PR checklist

- [ ] Added/modified tests


RUPESH-KUMAR01:
The cutoff date was set to 2025-10-10 to maintain compatibility with skbase. Earlier dates resolved to versions that did not include _safe_import, which led to test collection errors. The cutoff date was therefore chosen based on dependency compatibility.

phoeenniixx:
FYI @fkiraly, please have a look at this!

fkiraly (changes requested):
I see - `scikit-base` has no dependencies itself.

Is there a way to make 2024 and 2025 timestamps apply for all other packages except `scikit-base`, but keeping `scikit-base` at the newest possible version?

RUPESH-KUMAR01:
Hi @fkiraly,
The tests will now run with scikit-base at the latest version, while the rest of the packages will use a cutoff date of 2025-01-01. Please let me know if we should choose a different cutoff date.

.github/workflows/test.yml
```diff
@@ -139,7 +144,16 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          pip install ".[dev, dependencies_2025, github-actions]"
+          uv pip install ".[dev, github-actions]"
+        env:
+          UV_SYSTEM_PYTHON: 1
+          UV_EXCLUDE_NEWER: "2025-01-01"
+
+      - name: Install latest scikit-base
```
phoeenniixx: in place of this we could also do something like [this](https://github.com/sktime/sktime/blob/c76a06575d48764310e03f85e2b4e39b45d03a61/.github/workflows/test.yml#L429-L432)?
I think that would be cleaner that re-installing `skbase`? Also please also keep `setuptools` to at its latest version

.github/workflows/test.yml
```diff
@@ -139,7 +144,16 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          pip install ".[dev, dependencies_2025, github-actions]"
+          uv pip install ".[dev, github-actions]"
+        env:
+          UV_SYSTEM_PYTHON: 1
+          UV_EXCLUDE_NEWER: "2025-01-01"
```
phoeenniixx: Can you please move the date to Aug 2025?

docs/source/tutorials/stallion.ipynb
phoeenniixx, line 1: PLease revert this change, idts it is related to the PR?

phoeenniixx (changes requested):
Thanks a lot!
I have added few comments please have a look at them.
Also, please remove this block from `pyproject.toml`, it is redundant and not useful after your updates!
https://github.com/sktime/pytorch-forecasting/blob/a03b98494a34bbed9a6457a4e1d753822e474d53/pyproject.toml#L120-L128


FYI @fkiraly

phoeenniixx:
FYI @fkiraly

phoeenniixx (approved):

fkiraly (approved):
Great, looks like it works!

PR #2073, opened 2026-02-20
[ENH] Add support for nn losses to ptf-v2

### LINKED ISSUE #1970: [ENH] Add support for `nn` losses to `ptf-v2`
Currently, `ptf-v2` only supports `metrics` as loss functions. As suggested by @fkiraly, we should also provide users some more flexibility by allowing the user to use `nn` losses as well.
But this poses some problems:
- `nn` losses cannot handle `list` and currently for multi-target, `ptf-v2` uses `list` of tensors
-  Some `nn` losses need the `target` (ground truth) and `y_pred` to be of same shape , while in `ptf-v2` they are of different shapes (`target` - 2D, `y_pred` - 3D). While other `nn` losses require them to be of different shapes. The choice of shapes clearly depends on what kind of loss we are using.

We could:
- Create an adapter class for `nn` losses that handle the shapes internally, so that the output contract of model layer remains consistent. The shape handling specific to the loss we are using, will happen inside these adapters.
- Create a wrapper for `MultiLoss` inside the `BaseModel` for `nn` losses (or for all losses?), which directly pass a list of same `nn` loss (or any other loss) if only one loss is passed in case of multi-target. This list will have same `len` as `num_targets`. This wrapper would send only tensors to `nn` losses and not `list`.

Nidhicodes:
#### Reference Issues/PRs

Fixes #1970 

This PR adds first-class support for standard PyTorch nn.Module losses (e.g. nn.MSELoss, nn.L1Loss) in the pytorch-forecasting v2 framework.

Key changes:
-Introduces NNLossAdapter, a lightweight wrapper that adapts nn.Module losses to the ptf-v2 loss/metric API.
-Enables users to pass native PyTorch losses directly to models (e.g. DLinear(loss=nn.MSELoss())) without manual wrapping.
-Ensures compatibility with BaseModel.predict() by implementing to_prediction and to_quantiles.

The adapter:
-Handles (target, weight) inputs by applying weights internally while respecting the loss’s original reduction.
-Supports multi-target predictions by splitting [B, T, N] tensors and summing losses across targets.
-Enforces point-prediction-only usage (H=1) for losses that are not horizon-aware, with clear error messages otherwise.

Please focus on:

> Correctness of the NNLossAdapter shape handling and weighting logic
> Auto-wrapping logic in BaseModel for distinguishing ptf metrics vs raw nn.Module losses
> Error handling and clarity of failure modes for unsupported multi-horizon predictions
> API consistency with existing ptf-v2 loss abstractions

Yes.
A new test module pytorch_forecasting/tests/test_nn_loss_adapter.py was added, covering:
-Basic loss wrapping behavior
-Weighted losses
-Multi-target predictions
-Proper error handling for unsupported multi-horizon inputs

#### Any other comments?

This PR is intentionally scoped to point-prediction losses only, keeping the behavior explicit and predictable.
Future extensions (e.g., horizon-aware nn.Module losses) can be built on top of this adapter if desired.

Happy to iterate based on reviewer feedback.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/docs/source/tutorials/stallion.ipynb
+++ b/docs/source/tutorials/stallion.ipynb
@@ -41,7 +41,6 @@
    "source": [
     "import warnings\n",
     "\n",
-    "\n",
     "warnings.filterwarnings(\"ignore\")  # avoid printing out absolute paths"
    ]
   },
--- a/pytorch_forecasting/metrics/__init__.py
+++ b/pytorch_forecasting/metrics/__init__.py
@@ -37,6 +37,7 @@ from pytorch_forecasting.metrics.distributions import (
     NegativeBinomialDistributionLoss,
     NormalDistributionLoss,
 )
+from pytorch_forecasting.metrics.nn_loss_adapter import NNLossAdapter
 from pytorch_forecasting.metrics.point import (
     MAE,
     MAPE,
@@ -50,6 +51,7 @@ from pytorch_forecasting.metrics.point import (
 from pytorch_forecasting.metrics.quantile import QuantileLoss
 
 __all__ = [
+    "NNLossAdapter",
     "MultiHorizonMetric",
     "DistributionLoss",
     "MultivariateDistributionLoss",
--- a/pytorch_forecasting/metrics/nn_loss_adapter.py
+++ b/pytorch_forecasting/metrics/nn_loss_adapter.py
@@ -1,5 +1,3 @@
-from typing import List, Optional, Tuple, Union
-
 import torch
 import torch.nn as nn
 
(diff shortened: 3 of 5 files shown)


fkiraly (changes requested):
Looks great!

I would actually write the modified loss to `self._loss` rather than modifying `self.loss` - and then have everything in the `BaseModel` reference `self._loss`.

Opinions about this request, @agobbifbk, @phoeenniixx, @PranavBhatP ?

PranavBhatP:
> I would actually write the modified loss to self._loss rather than modifying self.loss - and then have everything in the BaseModel reference self._loss.

Makes sense to me!

Faakhir30:
Hi @Nidhicodes, are you still working on this? this PR looks stale, though the adapter class LGTM apart from mentioned changes by maintainers, I'd like to open a followup PR taking your commits (if this is stale).

Issue #1974, opened 2025-09-26
[ENH] Work Items for `ptf-v2`


phoeenniixx:
We already have a beta version released for `ptf-v2`, but still a lot of work is to be done. 
This issue aims to collect and track the pending work items and improvements that need to be discussed and addressed.

Work Items (not in any specific priority):

- [ ] `predict` function and `_pkg` wrapper (as mentioned in #1870 and sktime/enhancement-proposals/pull/46) implementation
- [ ] How to handle `categorical` variables ?
- [ ] Add `scalers` for preprocessing
- [ ] Add more `train_test_split` strategies
- [ ] Add `label_encoders` to D1 Layer?
- [ ] Add support for datasets which cannot fit in memory
- [ ] Move the v1 models to v2 (should be of low priority)

phoeenniixx:
FYI @fkiraly, @agobbifbk, @PranavBhatP 
Please feel free to add more tasks that I may have missed out

phoeenniixx:
Also, I was thinking about the `datamodules`: 
Right now, we have 2 - one for `ptf` native models and one for `tslib`, so I was thinking, maybe moving forward we should also add a Base Class for these data modules as well?

fkiraly:
Preliminary thoughts on priorities. I would advise three parallel tracks:

1. v2 feature parity, with priority in decreasing order:

* `predict` interface
* scalers
* categorical inputs and label encoder

2. adding more models - v1 and/or v2

incl writing an extension template

I think this is important to:

* validate the v2 API as it progresses
* validate the v1 testing framework (and with it v2)
* encourage new contributors

Noting that we will likely have to revisit v2 estimators as we are changing API.

3. attempt full integration with dsipts - using the 2nd party integration pattern as between sktime and prophetverse

agobbifbk:
My two cents for the numerical scalers and not-in-memory data. We can leverage on `partial_fit` method of most of the scalers (unfortunately not for categorical variable) without reinventing the wheel 
https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler.partial_fit

ojuschugh1:
Hi Team, I hope you are doing well. I have choosen to pick on  Add label_encoders to D1 Layer task, i have started the implementation , will open the PR tommorrow.

Thanks

PR #2082, opened 2026-02-21
[ENH] Added Support for Categorical Variable Handling

### LINKED ISSUE #2062: [ENH] Support for Categorical Variable Handling
Referencing #1974 

Currently, `ptf-v2` lacks a standardized, built-in pipeline for handling categorical variables. As we move out of beta, users need a consistent way to transform non-numeric data (like strings or labels) into a format the models can actually process without manually writing boilerplate preprocessing code for every project.

I'd like to see a dedicated module or utility within the preprocessing layer that handles common categorical encoding tasks. Specifically:

- Standardized Encoders: Implementation of One-Hot Encoding and Ordinal/Label Encoding that integrates directly with the ptf workflow.

- API Design: A clear choice on whether this should be Automated (where the framework detects string columns and encodes them) or User-Guided (where the user explicitly defines which columns to encode in the DataModule).

- Consistency: The solution should ensure that the same encoding mappings used during training are preserved and applied during the predict function to avoid data leakage or shape mismatches.

Muhammad-Rebaal:
Closes #2062  which references #1974

Hi @PranavBhatP , @phoeenniixx !

I've resolved the issue to add support for categorical variable handling in the ptf-v2 preprocessing pipeline. Could you please review the PR?

- **Created Encoders:** Added `pytorch_forecasting/data/categorical_encoders.py` which implements `PTFOrdinalEncoder` and `PTFOneHotEncoder`.
- **Integrated into DataModules:** Added the categorical_encoders parameter, initialization logic, and preprocessing transformations to both **EncoderDecoderTimeSeriesDataModule** (`pytorch_forecasting/data/data_module.py`) and **TslibDataModule** (`pytorch_forecasting/data/_tslib_data_module.py`). This allows for both user-guided and automated mapping.
- **Inference Consistency:** Ensured the categorical mapping state is saved during setup(stage="fit") and seamlessly applied during `predict()` to prevent mapping mismatch and data leakage.

Thank You !


docs/source/tutorials/stallion.ipynb
```diff
@@ -41,7 +41,6 @@
    "source": [
     "import warnings\n",
     "\n",
-    "\n",
```
PranavBhatP: Why is this file being edited?

PranavBhatP (commented):
@Muhammad-Rebaal thanks for the contribution! I don't see why we would need to create a separate file for categorical encoders. We already have one - `encoders.py` which contains both the scalers and encoders of v1. In the current plan for v2 we are thinking of reusing the existing ones in `encoders.py`. 

- I would suggest trying to make a working encoding pipeline in one of the v2 data modules (preferably focus on `EncoderDecoderTimeSeriesDataModule` first), using existing encoders.
- Once that is done, write tests for the data module to check whether the usage of encoders is working as intended.
- Then move onto the next data module and repeat the same. 

For the sake of this PR, I would recommend you to focus on implementing label encoding (using the existing encoders) on the `EncoderDecoderTimeSeriesDataModule`. Maybe then you can add a few more types of encoding techniques and we will also have a good understanding of whether v1 encoders work well with the D2 layer of PTF v2.

FYI @phoeenniixx @agobbifbk any thoughts?

docs/source/tutorials/stallion.ipynb
Muhammad-Rebaal: It happened because of running the pre-commit, although I'll change it back.

Muhammad-Rebaal:
> @Muhammad-Rebaal thanks for the contribution! I don't see why we would need to create a separate file for categorical encoders. We already have one - `encoders.py` which contains both the scalers and encoders of v1. In the current plan for v2 we are thinking of reusing the existing ones in `encoders.py`.
> 
> * I would suggest trying to make a working encoding pipeline in one of the v2 data modules (preferably focus on `EncoderDecoderTimeSeriesDataModule` first), using existing encoders.
> * Once that is done, write tests for the data module to check whether the usage of encoders is working as intended.
> * Then move onto the next data module and repeat the same.
> 
> For the sake of this PR, I would recommend you to focus on implementing label encoding (using the existing encoders) on the `EncoderDecoderTimeSeriesDataModule`. Maybe then you can add a few more types of encoding techniques and we will also have a good understanding of whether v1 encoders work well with the D2 layer of PTF v2.
> 
> FYI @phoeenniixx @agobbifbk any thoughts?

Thanks for guidance @PranavBhatP ,
The reason to make separate file for categorical is to make code clean although as per your guidance I'll merge the logic within encoder and write tests for it.

agobbifbk:
Ok thx for working on this! 
One observation: you are working only in the D2 layer here if I correctly understand. We need to do something also in the D1 layer (and I think the D2 will be easier then) because we allow the D1 to produce only tensors (i.e. integers and float). So if we have a category that is a string in the original data, we need to put the encoding logic in the D1 to convert them into integers otherwise how we can pass those values to the D2? In such situation the D2 already knows the max id for each class (for example through the metadata) and the implementation will be easier no?

fkiraly:
agree with @agobbifbk.

Muhammad-Rebaal:
@PranavBhatP  and @agobbifbk   Thank you both for the great feedback! I have just pushed an update that implements both of your architectural recommendations perfectly.

**Regarding the Encoders:**
I completely agree with the feedback on keeping the encoders unified. I have deleted the experimental categorical encoders file and injected the categorical encoding logic (`PTFOrdinalEncoder` and `PTFOneHotEncoder`) directly into our main encoders module.

**D1 / D2 Architecture:**
@agobbifbk  were completely right, handling this solely in the D2 layer wasn't the correct approach. I have now refactored the pipeline so the heavy lifting happens entirely in the D1 layer (the `TimeSeries` dataset), exactly as you envisioned.

Here is how the pipeline flows now:

- **D1 encodes everything:** The `TimeSeries` class now natively initializes the encoders and transforms the String/Categorical columns into Integers right from the start. This ensures D1 strictly produces only tensors natively.
- **D1 Metadata exposing Cardinalities:** The initialization inside D1 now calculates and exposes the max vocabulary size via `metadata["categorical_cardinalities"]`.
- **D2 is drastically simplified:** Because D1 handles the string-to-integer conversion, the D2 data modules no longer perform any heavy encoding loops. D2 simply reads `"categorical_cardinalities"` from the metadata and safely casts the feature slices using `.long()`.

One thing I'd like to let you know is that **code-quality issue** is caused in the file  `docs/source/tutorials/stallion.ipynb`  which is because of the extra line previously removed by the `pre-commit` but added as per your @PranavBhatP request, so let me know what to do with that.

CC : @fkiraly @phoeenniixx

Muhammad-Rebaal:
Hi @phoeenniixx , @PranavBhatP , @fkiraly

Kindly review this PR.

I've also `implement standardized categorical variable handling with D1/D2 preprocessing and embedding support` the idea suggested by @agobbifbk on discord.

agobbifbk:
Hello! 
Some comments:
1- 
```python
        for col, n_cats in metadata["categorical_cardinalities"].items():
            emb_dim = min(50, (n_cats + 1) // 2) if n_cats > 0 else 0
            metadata["categorical_embedding_sizes"][col] = (n_cats, emb_dim)
``` 
I think this is not something we want to define at D1/D2 layer, we will delegate it to the model layer.
2- since this D1 layer is related to `pandas` dataframe, should we leverage on sklearn encorders or not? I mean, we have a lot of free stuff and we can still use it so that in the D2 there is nothing to do 
3- Onehotencoding seems akward, we are introducing a lot of new channels for the categorical variables...


The idea, in my opinion, is that we load a csv, with some categorical variables, we need to convert them into integers in order to correctly create the batch in the D2. And the metadata is needed for initializing the models. Am I missing something?

Muhammad-Rebaal:
> Hello! Some comments: 1-
> 
> ```python
>         for col, n_cats in metadata["categorical_cardinalities"].items():
>             emb_dim = min(50, (n_cats + 1) // 2) if n_cats > 0 else 0
>             metadata["categorical_embedding_sizes"][col] = (n_cats, emb_dim)
> ```
> 
> I think this is not something we want to define at D1/D2 layer, we will delegate it to the model layer. 2- since this D1 layer is related to `pandas` dataframe, should we leverage on sklearn encorders or not? I mean, we have a lot of free stuff and we can still use it so that in the D2 there is nothing to do 3- Onehotencoding seems akward, we are introducing a lot of new channels for the categorical variables...
> 
> The idea, in my opinion, is that we load a csv, with some categorical variables, we need to convert them into integers in order to correctly create the batch in the D2. And the metadata is needed for initializing the models. Am I missing something?

Hi @agobbifbk , thanks for the feedback! I've been thinking about your points, and here is my perspective on how we can refine this:

**1. Embedding Sizes in D2:** The main reason I added **categorical_embedding_sizes** to the metadata is to provide a DRY (Don't Repeat Yourself) default. If we delegate this entirely to the model layer, every single model initialization (TFT, TimeXer, etc.) or the end-user will have to duplicate the `min(50, (n_cats + 1) // 2)`  heuristic entirely. By having D2 provide it in metadata, models can read it as a **suggested default** while still allowing users to override it if they want. I think this keeps the model code much cleaner. Does that separation make sense, treating D2's value as a suggested default?

**2. Leveraging Sklearn Encoders:** Fair point, we shouldn't reinvent the wheel with `PTFOrdinalEncoder`. Since we need to robustly map **NaNs** and unseen categories to index 0 for embeddings, standard **Sklearn** encoders can sometimes be a bit clunky to configure. Instead, I think we can just drop `PTFOrdinalEncoder` and use the library's already-existing `NaNLabelEncoder`as the default string-to-integer converter in D1. This gives us the robust Sklearn-like behavior without adding arbitrary new encoders to the codebase.

**3. One-Hot Encoding:** I completely agree with you on this. One-hot encoding explodes the feature space and complicates the batching logic unnecessarily since models strictly use **nn.Embedding** anyway. I will drop `PTFOneHotEncoder` completely from this implementation so we keep the D2 pipeline strictly integer-based.

Let me know if you are on board with this revised approach! I can quickly strip out the one-hot logic and refactor D1 to use `NaNLabelEncoder` today.

agobbifbk:
1- fair point, so far I have seen that the final size of embedding layer is not class dependent (meaning that all the categorical variables are mapped in the same dimension so that you can sum up all the contributors reducing the number of channels dedicated to the categorical variables). Since most of the architecture does not have a specific layer that deal with categorical variable is up to us decide how to threat them. If we go for different shape embedding at the end you need to stack them ending with a huge block, I'm a little worry about that. We need to came up with an idea for the model layers for a unified processing of this block: I don't want to push to the direction of same-shape-then-sum (current DSIP implementation) because maybe here is the place where we can have more clever ideas :-) Delegating to the D1 layer this choice is somehow risky because we are coupling the layers and this goes against the idea of having these layers, @phoeenniixx  and @PranavBhatP maybe we can discuss it in the next community meeting?
2- ok for me but for uniformity we should do all the calculation in the D1 layer, as it seems to be in this PR right? 
3- top

Clean the code and then let's have a look! THX!

Muhammad-Rebaal:
Hi @agobbifbk @PranavBhatP @phoeenniixx ,

Kindly review this PR.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -125,6 +127,44 @@ def __init__(
...
+        # Initialize encoders if set to "auto"
+        if self.categorical_encoders == "auto":
+            self.categorical_encoders = {
+                col: NaNLabelEncoder(add_nan=True) for col in self._cat
+            }
+        else:
+            # Ensure we have an encoder for every categorical column
+            for col in self._cat:
+                if col not in self.categorical_encoders:
+                    self.categorical_encoders[col] = NaNLabelEncoder(add_nan=True)
```
phoeenniixx: I think we should add a warning that we are adding a default here if the col is not present.. 
Also please add `categorical_encoders` to the docstring and explain how the dict should be initialized and which "encoders" are allowed here

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -125,6 +127,44 @@ def __init__(
...
+        for col, encoder in self.categorical_encoders.items():
+            # If the encoder is untrained (no classes_ attribute)
+            if not hasattr(encoder, "classes_") or len(encoder.classes_) == 0:
+                # Fit on training data
+                encoder.fit(self.data[col])
+
+            # Transform the categorical column to integers
+            encoded = encoder.transform(self.data[col])
+            if isinstance(encoded, torch.Tensor):
+                self.data[col] = encoded.numpy()
```
phoeenniixx: - why are we changing the tensor back to numpy here? I think D1 layer changes the numpy to tensors right? why are we going backwards here? 
- Should we transform in the `__getitem__` directly if the encoders expect the tensors... When the data is transformed to tensors, we send it to the encoders after that?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -89,6 +90,7 @@ def __init__(
         known: list[str | list[str]] | None = None,
         unknown: list[str | list[str]] | None = None,
         static: list[str | list[str]] | None = None,
+        categorical_encoders: dict | str = "auto",
```
phoeenniixx: Which encoders are we going to use here? Just the `ptf` ones or from `sklearn` as well?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
phoeenniixx, line 131: I think you should collect the param sent from the `__init__` here as well.
```python
self.categorical_encoders = categorical_encoders
```
And then if we need to make any updates we could use `self._categorical_encoders` instead, which means once `self.categorical_encoders` is set, it is not changed after that

(see how `self.num` and `self._num` are used)

pytorch_forecasting/data/_tslib_data_module.py
```diff
@@ -300,6 +300,7 @@ def __init__(
         | list[NORMALIZER]
         | tuple[NORMALIZER]
         | None = "auto",  # noqa: E501
+        categorical_encoders: dict[str, Any] | str | None = "auto",
```
phoeenniixx: How is it different from D1 layer `categorical_encoders`? Please add the docstrings to explain the params added clearly.

pytorch_forecasting/data/data_module/_tslib_data_module.py
```diff
@@ -520,8 +531,39 @@ def _prepare_metadata(self) -> dict[str, Any]:
...
+        If categorical_encoders="auto" (the default), this method pulls
+        the already-fitted encoders from the D1 TimeSeries dataset.
+        This ensures the same mappings are used during prediction.
+        """
+        if self._categorical_encoders == "auto":
+            self._categorical_encoders = (
+                self.time_series_dataset.categorical_encoders.copy()
+            )
+
+    def get_categorical_encoders(self) -> dict:
```
phoeenniixx: where (and why) is this method used?

pytorch_forecasting/data/encoders.py
```diff
@@ -1576,3 +1577,106 @@ def func(*args, **kwargs):
...
+        return np.array([self.inverse_mapping_.get(idx, None) for idx in y])
+
+    @property
+    def n_classes(self) -> int:
+        if self.add_unknown:
+            return len(self.mapping_) + 1
+        return len(self.mapping_)
+
+
+class PTFOneHotEncoder(BaseEstimator, TransformerMixin, CategoricalEncoderMixin):
```
phoeenniixx: Can we also use the `sklearn` One Hot encoder here? Or maybe the [`torch.one_hot`](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html) adapter for `ptf`?

phoeenniixx (changes requested):
Thanks! I have added some comments
Also, please have a look at some other doubts and suggestions:
- Please add the new params to docstrings and explain what are these params are and what are the expected input? Like what should I pass to `categorical_encoders` - `ptf` encoders or the `sklearn` ones or both?
- If you add a new param to `__init__`, please collect it in the param name of same name like:  `self.param = param`. If you need to make any change to that param, please donot change `self.param`, rather create a new one as `self._param` and then perform the operations on it
- I am not able to understand why are we accepting the `categorical_encoders` in D2 as well as in D1. Are they expected to be different?
- Also, I can't see how `categorical_enocders` are used in D2, can you please explain that?
- Also, I think we could also use the `sklearn` encoders and other already implemented enocders? Do we really need to reimplement encoders like `OrdinalEncoder` and `OneHotEndcoder`? I think we could also add adapters for it from `sklearn` so that it can handle tensors if needed? or maybe we could `fit` and `transform` it before changing to tensors if we are using only in D1 layer - in this case we might not even need adapters? (I am still not sure why D2 layer need encoders?) What do you think @Muhammad-Rebaal @agobbifbk ?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: Great catch, removed entirely. **NaNLabelEncoder.transform()** already returns numpy arrays directly, so the isinstance(encoded, torch.Tensor) check was unnecessary. The code is now simply 
`self.data[col] = encoder.transform(self.data[col])`. D1's `getitem`  handles the numpy→tensor conversion later via torch.tensor(), so there's no need to go backwards.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: Done! Added a **UserWarning** inside the if col not in self._categorical_encoders block. Also added `categorical_encoders` to the class docstring explaining it only accepts  `NaNLabelEncoder`  instances (not PTF or sklearn encoders), defaults to "auto", and how to pass pre-fitted encoders for prediction.

pytorch_forecasting/data/encoders.py
Muhammad-Rebaal: Both **PTFOneHotEncoder** and **PTFOrdinalEncoder** have been removed as per @agobbifbk's feedback, one-hot expands feature dimensions unnecessarily and all models use `nn.Embedding` for categoricals. If one-hot were needed in a future PR, `torch.nn.functional.one_hot` would be the right approach it works directly on the integer output from `NaNLabelEncoder` with zero conversion overhead. Thanks for notifying its presence here.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: After discussion with @agobbifbk, we're using only the existing `NaNLabelEncoder` from **encoders.py**. `PTFOrdinalEncoder` and `PTFOneHotEncoder` have been removed entirely. Since encoding runs at the pandas DataFrame level in D1 (before any tensor conversion), no sklearn adapters are needed `NaNLabelEncoder` already handles NaN/unknown tokens natively.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: Done, added `self.categorical_encoders = categorical_encoders` directly after `self.static = static`, matching the pattern used by `self.num/self._num`. All internal operations now go through `self._categorical_encoders` only.

pytorch_forecasting/data/_tslib_data_module.py
Muhammad-Rebaal: It's not different, D2 doesn't do any encoding itself. D1 is where we fit and transform the categorical columns into integers using `NaNLabelEncoder`. D2's `categorical_encoders` defaults to "auto", which just copies the already-fitted encoders from D1 during `setup()`. The only reason D2 has this param is so users can call `get_categorical_encoders()` after training and pass those fitted encoders to a new TimeSeries  for prediction, keeping the category→integer mappings consistent. I've added docstrings to both layers to make this clear.

pytorch_forecasting/data/data_module/_tslib_data_module.py
Muhammad-Rebaal: It's used in the prediction workflow. 
After `dm.setup(stage="fit"), a user calls fitted_encoders = dm.get_categorical_encoders()` and passes them into a new `TimeSeries(categorical_encoders=fitted_encoders)`  for prediction data. This guarantees unseen categories (e.g. "purple" at inference) map consistently to index 0 (the NaN/unknown token), preventing label leakage. This is validated in test_predict_encoder_consistency in test_data_module.py.

Muhammad-Rebaal:
Hi @phoeenniixx, thanks for the detailed review! Here's a summary addressing all your points:

1. **Docstrings & param convention:** Done, added `categorical_encoders` to docstrings in D1 and both D2 modules. Also followed the self.param = param / self._param pattern (same as self.num / self._num).

2. **Which encoders:** We're using only the existing `NaNLabelEncoder`. As @agobbifbk feedback, I've dropped `PTFOrdinalEncoder` and `PTFOneHotEncoder` entirely. Since encoding happens at the pandas level in D1 (before tensors), sklearn adapters aren't needed either `NaNLabelEncoder` already handles NaN/unknown tokens natively.
 
3. **Why D2 also accepts** `categorical_encoders`: D2 does not encode anything. It defaults to "auto" and just copies the already-fitted encoders from D1 during `setup()`. The reason it has this param is for the prediction workflow — after training, users call **dm.get_categorical_encoders()** and pass them to a new `TimeSeries` for inference, keeping the mappings consistent. D2 also reads *categorical_cardinalities* from** D1's metadata and passes it through to models for **nn.Embedding init**.
 
4. **Reimplementing encoders:** Agreed, no need. That's why we dropped the custom ones and stuck with `NaNLabelEncoder`. If one-hot were ever needed, torch.nn.functional.one_hot would be the native choice, but per @agobbifbk, one-hot is out of scope since models use **nn.Embedding**.

codependent:
Hi @Muhammad-Rebaal @phoeenniixx 

I'm working on a Master's thesis that involves training a TimeXer model. I was planning to use [PyTorch Forcasting v2 implementation](https://pytorch-forecasting.readthedocs.io/en/latest/api/pytorch_forecasting.models.timexer._timexer_v2.TimeXer.html?utm_source=chatgpt.com) but found this comment on the documentation:

```
[1] This implementation handles only continuous variables in the context length. Categorical variables
support will be added in the future.
```

Categorical support is a must for the kind of time series I'm working on which includes them as past and future covariates. Would this PR address this limitation? In that case, could you provide some estimation for a release that includes it?

Thank you for providing this awesome library and the hard work!

phoeenniixx:
> Hi @Muhammad-Rebaal @phoeenniixx
> 
> I'm working on a Master's thesis that involves training a TimeXer model. I was planning to use [PyTorch Forcasting v2 implementation](https://pytorch-forecasting.readthedocs.io/en/latest/api/pytorch_forecasting.models.timexer._timexer_v2.TimeXer.html?utm_source=chatgpt.com) but found this comment on the documentation:
> 
> ```
> [1] This implementation handles only continuous variables in the context length. Categorical variables
> support will be added in the future.
> ```
> 
> Categorical support is a must for the kind of time series I'm working on which includes them as past and future covariates. Would this PR address this limitation? In that case, could you provide some estimation for a release that includes it?

Hi @codependent! Welcome to `pytorch-forecasting` and thanks for using it!
The v2 is still unstable and it is still a WiP, we are working to incorporate all the features asap, but it would still need 
some time to get stable. For categorical support, I assume we could have it by the next release - after `v1.7.0` release I mean (or maybe the one after that). We currently release beta versions for v2 along with v1 patches :)

For the time being, I would suggest you use [v1](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer.py) for `TimeXer` training. It is stable and has categorical support! The training process would be different from the v2 though - there you would need to use `TimeSeriesDataset` in place of datamodules for training. You could look at [tutorials](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials.html) for more info!

Please let us know if you need any help. You can ping us on discord as well: https://discord.gg/ytC8dqVqF

codependent:
> Hi @codependent! Welcome to `pytorch-forecasting` and thanks for using it! The v2 is still unstable and it is still a WiP, we are working to incorporate all the features asap, but it would still need some time to get stable. For categorical support, I assume we could have it by the next release - after `v1.7.0` release I mean (or maybe the one after that).
> 
> For the time being, I would suggest you use [v1](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer.py) for `TimeXer` training. It is stable and has categorical support! The training process would be different from the v2 though - there you would need to use `TimeSeriesDataset` in place of datamodules for training. You could look at [tutorials](https://pytorch-forecasting.readthedocs.io/en/stable/tutorials.html) for more info!
> 
> Please let us know if you need any help. You can ping us on discord as well: https://discord.gg/ytC8dqVqF

@phoeenniixx I do appreciate your quick answer :-) I'll stick with v1 then, I mistakenly thought it didn't support categorical features either.

Kind regards

pytorch_forecasting/data/timeseries/_timeseries_v2.py
phoeenniixx: I have a doubt: if for label-encoding we only need `NaNLabelEncoder`, do we need to ask it from the user? Or are there any other label encoder can also be used?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -187,6 +229,15 @@ def _prepare_metadata(self):
 
             self.metadata["col_known"][col] = "K" if col in self._known else "U"
 
+        # Expose cardinalities for each categorical variable
+        self.metadata["categorical_cardinalities"] = {}
+        for col in self._cat:
+            encoder = self._categorical_encoders.get(col)
+            if encoder is not None and hasattr(encoder, "classes_"):
```
phoeenniixx: Question:
If encoder has no `classes_` attribute, would that mean that it is not fitted? and it is not possible to reach this `if` statement due to line 160? Am I missing something here?

pytorch_forecasting/data/_tslib_data_module.py
phoeenniixx: Should the D2 layer not infer this from D1? Is it necessary that user has to pass this?

pytorch_forecasting/data/encoders.py
```diff
@@ -1576,3 +1577,14 @@ def func(*args, **kwargs):
                     return [getattr(norm, name) for norm in self.normalizers]
             else:  # attribute does not exist for all normalizers
                 raise e
+
+
+class CategoricalEncoderMixin:
```
phoeenniixx: Where are we using this mixin class?

Muhammad-Rebaal: It was added speculatively as a future extension point for giving encoders consistent tensor I/O in the v2 pipeline, but the current implementation handles type coercion directly in the D1 layer. I'll remove it to avoid dead code.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: Yes you're right, by line 160, every encoder is guaranteed to be fitted and have `classes_`. The guard is redundant. I've simplified it.

pytorch_forecasting/data/_tslib_data_module.py
PranavBhatP: yes I have a similar concern here, the user should not be allowed to override the info which flows down from D1 to the D2. It should be automatically inferred from the D1.

Muhammad-Rebaal: yes I got your point regarding D2 I remove the categorical parameter so that D2 has to restrictly use the D1 `categorical_encoder` and user can't `override` that.

Muhammad-Rebaal:
Hi @phoeenniixx, 
Kindly review this PR.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -98,7 +100,7 @@ def __init__(
         | list[NORMALIZER]
         | tuple[NORMALIZER]
         | None = "auto",
-        categorical_encoders: dict[str, NaNLabelEncoder] | None = None,
+        categorical_encoders: dict[str, Any] | str | None = "auto",
```
phoeenniixx: Why are we accepting the encoders here? If we decided that this would happen in D1? Why not just remove this param?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,6 +136,36 @@ def __init__(
         self._unknown = _coerce_to_list(unknown)
         self._static = _coerce_to_list(static)
 
+        if (
+            isinstance(self.categorical_encoders, str)
+            and self.categorical_encoders == "auto"
```
phoeenniixx: if `auto` means `NanLabelEncoder` in all cases then why not just make that default? How does `auto` work in v1?

tests/test_data/test_d1.py
```diff
@@ -397,3 +397,31 @@ def test_group_index():
...
+        }
+    )
+
+    ts = TimeSeries(
+        data=data,
+        time="timestamp",
+        target="target",
+        group=["group"],
+        cat=["cat_feature", "group"],
+        categorical_encoders="auto",
```
phoeenniixx: please check for multiple cases - with and without encoding, different encoders etc

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -73,6 +74,15 @@ class TimeSeries(Dataset):
     static : list of str, optional, default = all variables not in known, unknown
         list of variables that do not change over time,
         list may also contain list of str, which are then grouped together.
+    categorical_encoders : dict[str, NaNLabelEncoder] or str, default="auto"
```
phoeenniixx: why not support `sklearn` encoders as well here? like in v1: 
https://github.com/sktime/pytorch-forecasting/blob/c4f5aac341082c56951bed2f352dc3bf6087ce35/pytorch_forecasting/data/timeseries/_timeseries.py#L345-L352

phoeenniixx (changes requested):
Thanks!
I think we still need to work on this:
- Why are we collecting encoders in d2, if everything is happening in D1?
- The `auto` option seems misleading - it just applies `NanLabelEncoder` and according to the docstring, `dict[str, NanLabelEncoder]`, it seems only `NanLabelEncoder` is supported, then this `auto` makes no sense?
- Why not add support for other encoders from `sklearn` as well? As this is happening in D1, we could apply `sklearn` encoders before changing to tensors

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
Muhammad-Rebaal: Yes, this was a lingering oversight from an earlier iteration. Since D1 is handling all the encoding and tensor conversion, D2 doesn't need to accept this config.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: ok sure.

tests/test_data/test_d1.py
Muhammad-Rebaal: ok, I've expanded the test suite in **test_d1.py** to include a dedicated test case `test_custom_sklearn_encoders`. This test injects a custom `sklearn.preprocessing.LabelEncoder` into the dataset to prove that custom non-PTF encoders are fully supported, properly calculate cardinalities, and output valid tensors.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: right. I'll adjust that as well.

Muhammad-Rebaal:
Hi @phoeenniixx,

I have removed the `categorical_encoders` parameter from the **D2** DataModules as requested, and updated **D1** to support `sklearn` encoders natively.

However, removing the parameter from the D2 `__init__` caused some CI tests to fail specifically in `docs/source/tutorials/ptf_V2_example.ipynb` and some tests in test_all_v2. The failure happens because those notebooks/tests define a `datamodule_cfg` dictionary that still passes `categorical_encoders` to the `DataModule`.

What is your suggestion in?

1. Should I go through the **ptf_V2_example.ipynb** notebook and the test config files to manually remove categorical_encoders from all `datamodule_cfg` dictionaries?

**OR**

2. Should I add `**kwargs` to the D2 DataModule `__init__` signatures so it safely absorbs and ignores deprecated parameters passed from older configs?

phoeenniixx:
> 1. Should I go through the **ptf_V2_example.ipynb** notebook and the test config files to manually remove categorical_encoders from all `datamodule_cfg` dictionaries?

this is how I would do it. Passing `kwargs` is not safe!

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -260,8 +253,31 @@ def _prepare_metadata(self):
             }
         )
 
+        metadata["categorical_cardinalities"] = self.time_series_dataset.metadata.get(
+            "categorical_cardinalities", {}
+        )
+
         return metadata
 
+    def get_categorical_encoders(self) -> dict:
```
phoeenniixx, line 319: where are we using this method?
If it is used for new `TimeSeries` class, how is it connected to the D2 layer?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,6 +136,23 @@ def __init__(
         self._unknown = _coerce_to_list(unknown)
         self._static = _coerce_to_list(static)
 
+        self._categorical_encoders = _coerce_to_dict(self.categorical_encoders)
+        for col in self._cat:
+            if col not in self._categorical_encoders:
+                self._categorical_encoders[col] = NaNLabelEncoder(add_nan=True)
```
phoeenniixx: what if i dont want to apply the encoders?

tests/test_data/test_d1.py
```diff
@@ -397,3 +397,60 @@ def test_group_index():
...
+    assert ts.metadata["categorical_cardinalities"]["cat_feature"] == 4
+
+    item = ts[0]
+    # Assert D1 emits purely tensors, no strings!
+    assert isinstance(item["x"], torch.Tensor)
+
+
+def test_custom_sklearn_encoders():
+    """Test injection of standard scikit-learn encoders instead of NaNLabelEncoder."""
+    from sklearn.preprocessing import LabelEncoder
```
phoeenniixx, line 457: why dont we use `parameterize` and test all the `sklearn` encoders as well?

tests/test_data/test_data_module.py
```diff
@@ -470,3 +471,149 @@ def test_multivariate_target():
 
     x, y = dm.train_dataset[0]
     assert len(y) == 2
+
+
+def test_v2_categorical_with_nan_label_encoder():
```
phoeenniixx: do we need this test? I mean the encoding is happening in d1, D2 has nothing to do with that?

phoeenniixx (changes requested):
I have some doubts please see below.
- I think we should think of most basic options as well - no encoders as well.
- And also document ALL the ooptions user have to pass to `categorical_encoders`!

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
Muhammad-Rebaal: The `get_categorical_encoders()` method on **D2** is a convenience accessor for the prediction workflow. During inference, users need to pass the fitted encoders from training to a new TimeSeries instance to ensure consistent category-to-integer mapping. Since **D2**  is the user-facing API, this method exposes the `D1's` internal encoders without requiring users to reach into `dm.time_series_dataset._categorical_encoders` directly.

You can see the intended workflow in `test_predict_encoder_consistency`:

```
fitted_encoders = dm.get_categorical_encoders()
ts_predict = TimeSeries(..., categorical_encoders=fitted_encoders)
```

This is the same pattern that v1's `TimeSeriesDataSet` uses.

tests/test_data/test_data_module.py
Muhammad-Rebaal: Thanks. I'll trim this test to only verify the D2-specific behavior: that `encoder_cat` arrives as torch.long in the batch. This ensures D2's `preprocessing` correctly handles the integer-encoded data from D1.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: Thanks I've handled this case as well

tests/test_data/test_d1.py
Muhammad-Rebaal: Thanks for the suggestion, I've implemented it that way

Muhammad-Rebaal:
Hi @phoeenniixx, @agobbifbk,  I've made the following changes, kindly have a look
- Parmeterize the `sklearnEncoder` tests.
- Document the options the user have to pass.
- If the user don't want to use encoders aka `no-encoders`.

I think the feature pipeline is ready & we can introduce this feature in the v2 models in new issue.

tests/test_data/test_data_module.py
```diff
@@ -726,3 +726,144 @@ def test_group_normalizer_uses_groups():
...
+        time_series_dataset=ts,
+        max_encoder_length=2,
+        max_prediction_length=1,
+        batch_size=2,
+    )
+    datamodule.setup(stage="fit")
+
+    batch = next(iter(datamodule.train_dataloader()))
+    x_batch, y_batch = batch
+    assert x_batch["encoder_cat"].dtype == torch.long
```
agobbifbk, line 762: Great! Just one comment, by default all the features are unknown so I can find them in the `encoder_` key right?

Muhammad-Rebaal: Yes

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,6 +154,26 @@ def __init__(
...
+                if col not in self._categorical_encoders:
+                    self._categorical_encoders[col] = NaNLabelEncoder(add_nan=True)
+
+        # Fit and transform categorical columns to integer codes
+        for col, encoder in self._categorical_encoders.items():
+            if not hasattr(encoder, "classes_") or len(encoder.classes_) == 0:
+                encoder.fit(self.data[col])
+
+            # NaNLabelEncoder.transform() returns numpy ints directly.
+            # D1's __getitem__ converts to tensors later via torch.tensor().
```
phoeenniixx: do we need this comment?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,6 +154,26 @@ def __init__(
         self._unknown = _coerce_to_list(unknown)
         self._static = _coerce_to_list(static)
 
+        if self.categorical_encoders is False:
+            self._categorical_encoders = {}
+        else:
+            self._categorical_encoders = _coerce_to_dict(self.categorical_encoders)
```
phoeenniixx, line 195: here shall we check if the cols passed actually belong to `cat` or not?

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: No, It just an explanatory notes to explain what's going on.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: I think its good thing to add because there can be cols in the data where some cols belong to cat and some aren't so apply `cat_encoders` to only on those cols that belong to cat.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,6 +154,44 @@ def __init__(
...
+
+        # Fit and transform categorical columns to integer codes
+        for col, encoder in self._categorical_encoders.items():
+            if not hasattr(encoder, "classes_") or len(encoder.classes_) == 0:
+                encoder.fit(self.data[col])
+
+            self.data[col] = encoder.transform(self.data[col])
+
+            if self.data_future is not None and col in self.data_future.columns:
+                self.data_future[col] = encoder.transform(self.data_future[col])
```
phoeenniixx, line 214: I think this should be moved into a method instead of everything inside the `__init__` and then we call it? This feels like too much indentation (especially the allocation of cat encoders when encoders are not specified)

phoeenniixx (changes requested):
Thanks
- Can you please explain the tests for D2? I think categorical encoders are more for D1?
- Also, I still dont understand completely the use of `get_categorical_encoders`. Can you please give me a usecase where it can be useful?

Muhammad-Rebaal:
> * Can you please explain the tests for D2? I think categorical encoders are more for D1?

As D2 is responsible for taking D1's output and passing it to the model, So I think we need to test that D2 doesn't mess it up:

- test_v2_batch_categorical_dtype :  Tests that D2 actually keeps the categorical data as integers when creating batches. If D2 accidentally turns them into floats, PyTorch will crash.
- test_d2_metadata_has_cardinalities :  Tests that D2 correctly grabs the number of categories from D1 and passes them along in its metadata. The model needs this information to build its layers.
- test_predict_encoder_consistency : This is a full workflow test. It tests the whole process of getting the encoders from D2 and use them for new dataset


> * Also, I still dont understand completely the use of `get_categorical_encoders`. Can you please give me a usecase where it can be useful?

Actually, it's a helper function that users can use to extract the mapped encoders during training, save them to a variable or file and reuse them later when new data comes in for prediction to guarantee the category numbers match.

It has nothing to do with the pipeline its a helper function

pytorch_forecasting/data/timeseries/_timeseries_v2.py
Muhammad-Rebaal: ok, I'll clean the `__init__` block. and add those indentations into separate  methods to be called internally

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,9 +154,71 @@ def __init__(
...
+
+        # overwrite __init__ params for upwards compatibility with AS PRs
+        # todo: should we avoid this and ensure classes are dataclass-like?
+        self.group = self._group
+        self.target = self._target
+        self.num = self._num
+        self.cat = self._cat
+        self.known = self._known
+        self.unknown = self._unknown
+        self.static = self._static
```
phoeenniixx, line 169: why are we doing this? These variables are already used no?

phoeenniixx: why are they  moved here? This code is as it is copy pasted from one part to another, why not keep this code stationary and add you new code to some other place - to keep the diff at minimum

pytorch_forecasting/data/timeseries/_timeseries_v2.py
```diff
@@ -124,9 +154,71 @@ def __init__(
...
+        for col, encoder in self._categorical_encoders.items():
+            if not hasattr(encoder, "classes_") or len(encoder.classes_) == 0:
+                encoder.fit(self.data[col])
+
+            self.data[col] = encoder.transform(self.data[col])
+
+            if self.data_future is not None and col in self.data_future.columns:
+                self.data_future[col] = encoder.transform(self.data_future[col])
+
+    def _prepare_group_index(self):
```
phoeenniixx, line 216: i see you added this new method and that is why the diff is showing that the code is moved.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
phoeenniixx: I see this part is already stationary. just a change in method led to this kind of diff

phoeenniixx (changes requested):
I think it is ready, i am just not very convinced about the d2 tests 
- they are jsut checking if the the batch passed or the final pred is tensor or not - that is already tested we dont need to add tests for it. 
- Or if the metadata is passed from d1 to d2 (`metadata_cardinalities`) is correct or not - again that is tested, we dont need to check that again. 
- The only thing that this PR changes in d2 is provide d1 encoders in d2 (which i am still a bit worried about due to concerns of coupling). We should add tests for this only if you want.
     - On this accont, i have a suggestion - why not provide a method in d1 directly to getr the fitted encoders? this si where the enocders are fitted anyway? that may be cleaner no?

Muhammad-Rebaal:
Hi @phoeenniixx,
I've made the following changes : 
- I've removed the D2 tests for the concern of coupling.
- I've added the getter method from D2 to D1

PR #2084, opened 2026-02-22
[ENH] Add dedicated MQF2 and MultiLoss integration test coverage; expose skip_trainer_test on DistributionLoss

### LINKED ISSUE #1933: [ENH] Add support for `MultiLoss` and `MQF2DistributionLoss` to the test framework for models v1
Right now, the test framework for models doesn't test models against `MQF2DistributionLoss` and `MultiLoss`. We need to add support for these losses.
There are some `todo` comments for `MQF2DistributionLoss` as well there.

GunaPalanivel:
#### Reference Issues/PRs

- Closes #1933.

---

## Summary

Issue #1933 identified two test coverage gaps in the v1 model test framework:

1. `MQF2DistributionLoss` never exercised the full `trainer.test()` path due to a hardcoded `isinstance` skip inside `_integration()`.
2. `MultiLoss([QuantileLoss(), QuantileLoss()])` lacked explicit coverage with a multi-target dataloader.

This PR resolves both by making trainer-test skipping explicit at the loss level and by strengthening integration test coverage.

---

### 1. `DistributionLoss` — explicit skip API

`pytorch_forecasting/metrics/base_metrics/_base_metrics.py`

```python
skip_trainer_test: bool = False
```

* Graduates the previous hardcoded skip into a class-level override.
* Default remains `False`, so existing subclasses are unaffected.

### 2. `MQF2DistributionLoss` override

`pytorch_forecasting/metrics/distributions.py`

```python
skip_trainer_test: bool = True
```

* MQF2 cannot execute `trainer.test()` without `cpflows` shape errors.
* The skip is now declared declaratively rather than checked via `isinstance`.

---

### 1. `dataloaders_for_mqf2` fixture

`tests/test_models/conftest.py`

* Enforces MQF2 preconditions:

  * integer-rounded targets (`.round()`)
  * `GroupNormalizer(transformation="log1p")`
* Guarded with `pytest.importorskip("cpflows")`.

### 2. `LossFactory` sentinel

`test_temporal_fusion_transformer.py`

* Introduces a lightweight wrapper to distinguish:

  * dataset-dependent loss factories
  * instantiated `torch.nn.Module` losses
* Necessary because `nn.Module` implements `__call__`.

### 3. `_integration()` refactor

* Replaces `loss` with `loss_spec`
* Resolves `LossFactory` instances against the dataset
* Replaces hardcoded MQF2 check with:

```python
getattr(net.loss, "skip_trainer_test", False)
```

### 4. New and updated tests

* `test_multiloss_explicit()` — full trainer path coverage for
  `MultiLoss([QuantileLoss(), QuantileLoss()])` with multi-target data.
* `test_mqf2_loss()` — now uses `dataloaders_for_mqf2` and dynamic `prediction_length`.
* Updated `test_distribution_loss()` and `test_non_causal_attention()` to use `loss_spec=`.

---

## Reviewer Focus

* Is `skip_trainer_test` correctly placed on `DistributionLoss` vs a narrower class?
* Is the `LossFactory` sentinel preferable to signature inspection or typing-based resolution?
* Are the MQF2 data preconditions complete and correct?
* Should the same `skip_trainer_test` + `LossFactory` pattern be propagated to other model test files as part of this PR?

---

## Tests

All production code changes exist solely to enable new test paths.

* `test_multiloss_explicit()` adds missing end-to-end coverage.
* `test_mqf2_loss()` now exercises the full integration path.
* No behavioral changes for existing losses (`skip_trainer_test=False` default).

---

#### Any other comments?

`LossFactory` is defined directly in `test_temporal_fusion_transformer.py`
rather than `conftest.py` intentionally. It is an implementation detail of
`_integration()` and does not belong in the shared fixture namespace.

Pre-commit hooks (`black`, `isort`, `flake8`) were executed via
`pre-commit run --all-files` before committing.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/metrics/distributions.py
+++ b/pytorch_forecasting/metrics/distributions.py
@@ -347,6 +347,7 @@ class MQF2DistributionLoss(DistributionLoss):
     """
 
     eps = 1e-4
+    skip_trainer_test: bool = True
 
     def __init__(
         self,
--- a/tests/test_models/test_temporal_fusion_transformer.py
+++ b/tests/test_models/test_temporal_fusion_transformer.py
@@ -32,6 +32,14 @@ from pytorch_forecasting.models.temporal_fusion_transformer.tuning import (
 )
 
 
+class LossFactory:
+    def __init__(self, fn):
+        self.fn = fn
+
+    def __call__(self, dataset):
+        return self.fn(dataset)
+
+
 def test_integration(multiple_dataloaders_with_covariates, tmp_path):
     _integration(
         multiple_dataloaders_with_covariates,
@@ -45,7 +53,7 @@ def test_non_causal_attention(dataloaders_with_covariates, tmp_path):
         dataloaders_with_covariates,
         tmp_path,
         causal_attention=False,
-        loss=TweedieLoss(),
+        loss_spec=TweedieLoss(),
         trainer_kwargs=dict(accelerator="cpu"),
     )
 
@@ -66,7 +74,16 @@ def test_distribution_loss(data_with_covariates, tmp_path):
     _integration(
         dataloaders_with_covariates,
         tmp_path,
-        loss=NegativeBinomialDistributionLoss(),
+        loss_spec=NegativeBinomialDistributionLoss(),
+    )
+
+
+def test_multiloss_explicit(dataloaders_multi_target, tmp_path):
+    _integration(
+        dataloaders_multi_target,
+        tmp_path,
+        loss_spec=MultiLoss([QuantileLoss(), QuantileLoss()]),
+        trainer_kwargs=dict(accelerator="cpu"),
     )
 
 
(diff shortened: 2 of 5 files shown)


tests/test_models/test_temporal_fusion_transformer.py
```diff
@@ -66,47 +74,43 @@ def test_distribution_loss(data_with_covariates, tmp_path):
...
-    ].dataset.min_prediction_length
+def test_mqf2_loss(dataloaders_for_mqf2, tmp_path):
+    prediction_length = dataloaders_for_mqf2["train"].dataset.min_prediction_length
 
     _integration(
-        dataloaders_with_covariates,
+        dataloaders_for_mqf2,
         tmp_path,
-        loss=MQF2DistributionLoss(prediction_length=prediction_length),
+        loss_spec=MQF2DistributionLoss(prediction_length=prediction_length),
```
phoeenniixx, line 100: why is this changed to `loss_spec`? Any specific reason of the renaming of param?

pytorch_forecasting/metrics/distributions.py
```diff
@@ -347,6 +347,7 @@ class MQF2DistributionLoss(DistributionLoss):
     """
 
     eps = 1e-4
+    skip_trainer_test: bool = True
```
phoeenniixx, line 350: Why is this param added? Can you please explain?

tests/test_models/test_temporal_fusion_transformer.py
GunaPalanivel: Originally, `_integration()` just took a `loss` that was already an instantiated `nn.Module`. That worked until MQF2 — it needs `prediction_length`, which we only know after the dataset is created. So we can’t just do `MQF2DistributionLoss(...)` at test parameterization time because the dataset doesn’t exist yet.

That’s why I introduced `LossFactory`: it’s just a thin wrapper that takes the dataset and builds the loss once the required values are available.

Since `_integration()` now accepts either a ready-made loss or a factory, calling the parameter just `loss` felt misleading — it’s not always a fully constructed module anymore. `loss_spec` seemed clearer because it signals that this is something that gets resolved internally.

If that name feels off, I’m totally fine renaming it to something more explicit like `loss_or_factory`. Just wanted the parameter name to reflect the new behavior honestly.

pytorch_forecasting/metrics/distributions.py
GunaPalanivel: This change directly addresses the root issue in #1933.

Previously, `_integration()` had this hardcoded check:

```python
if not isinstance(net.loss, MQF2DistributionLoss):
    trainer.test(...)
```

That was marked as a TODO in the code. The reason it existed is that `MQF2DistributionLoss` uses normalizing flow gradients (`cpflows`) internally. When `trainer.test()` runs, Lightning wraps parts of execution in `torch.no_grad()` in ways that break MQF2’s flow computation, leading to shape errors. So skipping `trainer.test()` for MQF2 is actually correct behavior — not a workaround.

The problem is how it was implemented. Hardcoding `isinstance(net.loss, MQF2DistributionLoss)` inside a generic test helper creates tight coupling between the test framework and a specific loss class. If another loss ever had the same constraint, we’d have to modify `_integration()` again.

By adding a `skip_trainer_test` flag on `DistributionLoss` (default `False`) and overriding it to `True` in `MQF2DistributionLoss`, the responsibility moves to the loss itself. `_integration()` just checks:

```python
getattr(net.loss, "skip_trainer_test", False)
```

Now the test helper doesn’t need to know anything about specific loss types. Any future loss with similar constraints can opt in without modifying shared test logic.

GunaPalanivel:
> Thanks! I think there might be some misunderstanding, by the test support for `MQF2`, we meant the support in the unified test framework (see [`test_all_estimators`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/tests/test_all_estimators.py)). The unified test framework doesnot have the MQF2 support and we need to do some debugging and add the support there. (see the [`_loss_mapping`](https://github.com/sktime/pytorch-forecasting/blob/602ee157d18bd00ea7e1cfddeb0c8184c138976f/pytorch_forecasting/tests/_loss_mapping.py#L46))

Hey @phoeenniixx — thanks for pointing that out. I scoped this to the legacy per-model tests (`test_temporal_fusion_transformer.py`), but the actual issue is about unblocking MQF2 in the unified framework (`test_all_estimators.py` + `_loss_mapping.py`). That’s on me — I conflated the two systems.

Looking at `_loss_mapping.py` (line 46), the TODO explicitly calls out MQF2. And in `test_all_estimators.py`’s `_integration()`, there’s commented-out code showing where it previously broke.

From what I can see, the two blockers in the unified framework are:

1. `MQF2DistributionLoss` needs `prediction_length` at construction time, but `DISTR_LOSSES_NUMERIC` instantiates losses eagerly — before any dataset exists.
2. `_integration()` calls `trainer.test()` unconditionally, with no skip mechanism for MQF2.

The `skip_trainer_test` flag and the lazy factory pattern I added for the legacy tests are actually the right primitives — they just need to be applied in the unified test path instead.

Happy to move this in that direction. Would you prefer I fold those changes into this PR, or open a fresh one focused purely on the unified framework?

PR #2086, opened 2026-02-22
[BUG] Fix GPU memory leak in PredictCallback (callbacks/predict.py)

ashnaaseth2325-oss:
## Summary

This PR fixes a critical GPU memory accumulation issue in
`pytorch_forecasting/callbacks/predict.py` inside the `PredictCallback` class.

During prediction, batch outputs and related tensors were appended directly to
internal lists **without detaching them from the computation graph and without moving them to CPU**. As a result, all batch tensors remained on GPU for the entire prediction epoch.

Because prediction can span thousands of batches, this caused **monotonic GPU memory growth**, eventually leading to CUDA OOM on moderately large datasets , especially when using:

* `return_x=True`
* `return_y=True`
* `mode="raw"`

This behavior differed from the legacy implementation, which correctly moved batch outputs to CPU immediately.

---

## Where the Problem Occurred

File:

```
pytorch_forecasting/callbacks/predict.py
```

Method:

```
PredictCallback.on_predict_batch_end()
```

### Previous Behavior (Problematic)

```python

if key == "x":
    self.info[key].append(x)
elif key == "y":
    self.info[key].append(y[0])
elif key == "decoder_lengths":
    self.info[key].append(x["decoder_lengths"])
```

Here:

* `processed_output` is a GPU tensor (or dict of tensors)
* `x` is a full GPU batch dictionary
* `y[0]` is a GPU tensor
* All were appended directly to persistent lists

These lists were only cleared at `on_predict_epoch_end()`, meaning GPU memory usage grew linearly with number of batches.

---

## Why This Was Critical

Prediction is commonly used:

* After `trainer.fit()` for evaluation
* In cross-validation
* In hyperparameter sweeps
* In production inference
* In notebooks with repeated `.predict()` calls

For realistic TFT workloads (100K+ samples):

* Full `x` batch dict ≈ ~2 MB per batch
* 1,500+ batches → ~3+ GB accumulated on GPU
* Raw mode worse (attention weights, intermediate states)

Observed behavior:

* `nvidia-smi` shows linear memory growth during prediction
* `torch.cuda.memory_allocated()` increases per batch
* CUDA OOM during `torch.cat()` in `on_predict_epoch_end()`

This makes prediction on larger datasets unreliable on 8–16 GB GPUs.

---

## Steps to Reproduce

```python
predictions = tft.predict(
    dataloader,
    return_x=True,
    return_y=True,
    mode="raw"
)
```

Monitor:

```python
torch.cuda.memory_allocated()
```

Expected behavior:

* Memory remains roughly constant

Actual behavior:

* Memory increases after every batch
* Eventually results in CUDA OOM

---

## Root Cause

The v2 `PredictCallback` implementation did not apply the same pattern used in the legacy implementation:

```python
move_to_device(detach(tensor), "cpu")
```

Instead, it accumulated raw GPU tensors in Python lists.

Since Python lists hold strong references, CUDA memory could not be released until the entire epoch finished — creating an O(dataset_size) GPU footprint instead of O(batch_size).

---

## The Fix

We now detach tensors and move them to CPU immediately upon accumulation.

### Updated Behavior

```python
self.predictions.append(
    move_to_device(detach(processed_output), "cpu")
)

self.info[key].append(
    move_to_device(detach(x), "cpu")
)
```

For specific tensor cases:

```python
self.info[key].append(
    x["decoder_lengths"].detach().cpu()
)
```

This ensures:

* No autograd graph retention
* No GPU tensor accumulation
* GPU memory bounded to a single batch at a time

---

## Impact

After this fix:

* GPU memory plateaus during prediction
* No OOM on large datasets
* `mode="raw"` becomes viable on standard GPUs
* Repeated `.predict()` calls are safe
* Behavior is consistent with legacy implementation

Peak memory now scales with **batch size**, not dataset size.

---

## Result

This restores safe and predictable memory behavior for `BaseModel.predict()` in `_base_model_v2`.
The fix is minimal, localized to `PredictCallback`, and introduces no architectural changes.
It eliminates a regression introduced during the v2 callback refactor and ensures production-scale inference remains stable.

### DIFF
--- a/tests/test_callbacks/test_predict_callback.py
+++ b/tests/test_callbacks/test_predict_callback.py
@@ -1,0 +1,126 @@
+"""Unit tests for ``pytorch_forecasting.callbacks.predict.PredictCallback``.
+
+Two branches in ``on_predict_batch_end`` were introduced by the GPU memory-leak
+fix (PR #2086) but were not exercised by the existing test suite:
+
+* Line 77 – the ``"y"`` branch:
+      ``self.info[key].append(move_to_device(detach(y[0]), "cpu"))``
+* Line 81 – the ``"decoder_lengths"`` branch:
+      ``self.info[key].append(x["decoder_lengths"].detach().cpu())``
+
+These tests directly invoke ``on_predict_batch_end`` to trigger each branch and
+assert that the stored tensors are detached and reside on CPU.
+"""
+
+import pytest
+import torch
+
+from pytorch_forecasting.callbacks.predict import PredictCallback
+
+
+# ---------------------------------------------------------------------------
+# helpers
+# ---------------------------------------------------------------------------
+
+def _make_batch(batch_size: int = 2, pred_len: int = 4):
+    """Return a minimal (x, y) batch pair suitable for raw-mode tests."""
+    x = {
+        "decoder_lengths": torch.full((batch_size,), pred_len, dtype=torch.long),
+    }
+    target = torch.randn(batch_size, pred_len)
+    index = torch.arange(batch_size)
+    return x, (target, index)
+
+
+def _batch_end(callback: PredictCallback, batch, outputs):
+    """Call ``on_predict_batch_end`` with null trainer/pl_module (raw mode)."""
+    callback.on_predict_batch_end(
+        trainer=None,
+        pl_module=None,
+        outputs=outputs,
+        batch=batch,
+        batch_idx=0,
+        dataloader_idx=0,
+    )
+
+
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @fkiraly @PranavBhatP @phoeenniixx  ,this is a small improvement to the earlier GPU memory fix.
I noticed decoder_lengths and batch[1] were still staying on GPU when return_y=True, so this ensures everything collected during prediction is consistently moved to CPU.

PR #2087, opened 2026-02-23
[ENH] Add more `train_test_split` strategies

### LINKED ISSUE #2064: [ENH] Add more `train_test_split` strategies
Hi @PranavBhatP ,

Kindly assign this to me ,

Referencing  #1974 

Currently, the `ptf-v2` framework relies on a basic training/testing split mechanism. While this works for simple datasets, it’s insufficient for more complex real-world scenarios—such as imbalanced classification, time-series forecasting, or grouped data—where a random split can lead to data leakage or unrepresentative evaluation metrics.

I would like to expand the train_test_split capabilities within the ptf-v2 DataModules. Specifically, the implementation should include:

- Stratified Splitting: To ensure that class distributions are preserved in both training and validation sets.

- Time-Series Splitting: Implementing a "sliding window" or "expanding window" split to respect the temporal order of data.
 
- Group-based Splitting: Ensuring that all data points from a specific group (e.g., the same user or location) stay within the same fold to test model generalization.
 
- Pattern Alignment: The implementation should follow the existing v2 pattern seen in the D1 Layer and current DataModule structures.

Muhammad-Rebaal:
Closes #2064  which references #1974

Hi @PranavBhatP , @phoeenniixx !

I've resolved the issue to expand the train_test_split capabilities within the ptf-v2 DataModules. Could you please review the PR?

*   **Created Splitters:** Added `pytorch_forecasting/data/splitters.py` which implements three new splitting functions: `random_series_split` (acts as group-based split), `stratified_series_split`, and `temporal_window_split`.
*   **Integrated into DataModules:** Added the `split_strategy` parameter ("random", "stratified", "temporal", "group") and delegated splitting logic to both `EncoderDecoderTimeSeriesDataModule` (`pytorch_forecasting/data/data_module.py`) and `TslibDataModule` (`pytorch_forecasting/data/_tslib_data_module.py`).
*   **Backwards Compatibility:** Ensured the previous randomized series split logic remains fully intact and acts as the default behavior when `split_strategy="random"`.

Thank You !


pytorch_forecasting/data/_tslib_data_module.py
```diff
@@ -695,51 +697,101 @@ def setup(self, stage: str | None = None) -> None:
...
-        else:
-            self._train_size = int(self.train_val_test_split[0] * total_series)
-            self._val_size = int(self.train_val_test_split[1] * total_series)
-
-            self._train_indices = self._indices[: self._train_size]
-            self._val_indices = self._indices[
-                self._train_size : self._train_size + self._val_size
-            ]
+        # Ensure categorical encoders are fitted
+        if getattr(self, "_categorical_encoders", "auto") == "auto":
```
phoeenniixx: Can you please specify where are you defining `_categorical_encoders`?

phoeenniixx: I think this should be handled sperately? or maybe you can stack both PRs?

phoeenniixx (changes requested):
Thanks!
There are some unrelated changes to `layers` (like changes to the documentation) please remove that

pytorch_forecasting/data/_tslib_data_module.py
Muhammad-Rebaal: Hi @phoeenniixx! You're absolutely right `_categorical_encoders` was never defined in it. But it is in this my PR #2082 . So, could you review that and we'd stack over after merging that I'll continue from here and make this `categorical_encoder's train_test_split` feature more stable. I just added that as a part of reference for future implementation.

Muhammad-Rebaal:
> Thanks! There are some unrelated changes to `layers` (like changes to the documentation) please remove that

Thanks for guiding @phoeenniixx, Kindly review the PR. I've removed the un-related changes.

Muhammad-Rebaal:
Hi @phoeenniixx , @PranavBhatP , @fkiraly 

Kindly review this PR

pytorch_forecasting/data/_tslib_data_module.py
phoeenniixx: I think `categorical_encoders` are out of scope for this PR? If you want to use them, I think you should stack on the the current PR using `git merge`?

phoeenniixx (changes requested):
Thanks!
I think if you want to access `categorical_enocders`, you should merge with the other PR?
Also, I have some doubts: 
- I can't see how the categorical encoders actually being used in the splitting, can you please explain that? I think they are currently a lit bit independent?
- If you added the logic for categorical encoders in `tslib_data_module` why not add it to `EncoderDecoderDataModule`? I think it would also need it?? :)

Muhammad-Rebaal:
Hi @phoeenniixx, thank you for the feedback and for catching this!

To answer your questions:

The original purpose of adding the `_categorical_encoders` here was to make the categorical features ready and processed prior to performing the train_test_split (especially for stratified splitting on categorical static features).
However, following your earlier comment advising that categorical encoders are out of scope for this specific PR and should be stacked, I removed the logic from the splitting strategy (which is why it isn't in EncoderDecoderDataModule either).
I simply forgot to remove the residual setup code from `_tslib_data_module.py`. This is why they seemed completely independent of the splitting logic here!

I have completely removed the lingering `_categorical_encoder` code from `_tslib_data_module.py` in my latest commit. This keeps the PR cleaned up and strictly scoped to expanding the train_test_split capabilities. We can properly handle the categorical encoders in the stacked PR (#2082) where I'll make sure they are added consistently.

Let me know if everything else looks good with the new splitter strategies!

pytorch_forecasting/data/data_module/_tslib_data_module.py
```diff
@@ -695,51 +697,69 @@ def setup(self, stage: str | None = None) -> None:
...
+            self._val_indices = torch.arange(total_series)
+            self._test_indices = torch.arange(total_series)
+        else:
+            raise ValueError(f"Unknown split_strategy: {self.split_strategy}")
 
         if stage == "fit" or stage is None:
             if not hasattr(self, "_train_dataset") or not hasattr(self, "_val_dataset"):
-                self._train_windows = self._create_windows(self._train_indices)
-                self._val_windows = self._create_windows(self._val_indices)
+                if self.split_strategy == "temporal":
```
agobbifbk, line 747: Can you provide a visual example (or a snippet of code) with let say:
1 - 3 groups
2- group 1 has 2 series with variable length
3- group 2 has 1 series with a given length
4- group 3 has 4 series with variable length
So we can visually see how the split will apply for each group? THX

Muhammad-Rebaal: **Diagram :** 


Untitled Diagram-2026-06-30T13-24-27


**Code :**
```
def temporal_window_split(
    windows: list[tuple[int, int, int, int]],
    train_val_test_split: tuple[float, float, float],
) -> tuple[
    list[tuple[int, int, int, int]],
    list[tuple[int, int, int, int]],
    list[tuple[int, int, int, int]],
]:
 
    # Group windows by series_idx
    series_windows = {}
    for w in windows:
        s_idx = w[0]
        if s_idx not in series_windows:
            series_windows[s_idx] = []
        series_windows[s_idx].append(w)

    train_windows, val_windows, test_windows = [], [], []

    for s_idx, sw in series_windows.items():
        # Ensure windows are sorted by time (start_idx: w[1])
        sw.sort(key=lambda x: x[1])
        total_w = len(sw)

        train_end = int(np.round(train_val_test_split[0] * total_w))
        if train_end == 0 and train_val_test_split[0] > 0 and total_w > 0:
            train_end = 1

        val_end = train_end + int(np.round(train_val_test_split[1] * total_w))
        if val_end > total_w:
            val_end = total_w

        train_windows.extend(sw[:train_end])
        val_windows.extend(sw[train_end:val_end])
        test_windows.extend(sw[val_end:])

    return train_windows, val_windows, test_windows

```
For the **Temporal split**, the filtering happens at the Window level. The **DataModule** assigns ALL series indices to Train, Val, and Test. It generates ALL possible windows for the entire dataset, and then passes that massive list of windows to `temporal_window_split()`. This function groups the windows by their `series_idx`, sorts them chronologically, and slices the window arrays based on our split ratios (e.g., the first 70% of windows for Series A go to Train).

agobbifbk: Nice for the representation, it is much easier now to understand :-) 
I have a point to raise: if we call it `temporal_split`  and the two series of group 1 are consecutive (for example yearly csv data) this procedure will break the data leakage paradigm, see for example https://arxiv.org/html/2412.11376v1

> To avoid information leakage during the evaluation phase in Section 4.3, each dataset is chronologically split into training, validation, and test sets with a ratio of 6:2:2

We can still have this splitting methodology, but I won't call it `temporal_split` because it is not a real temporal split. You should look at the timestamp and decide which samples goes in the correct place.
Any thoughts @phoeenniixx ?

phoeenniixx:
Hi @agobbifbk @Muhammad-Rebaal 
Can we try something like `sktime` implementation: https://github.com/sktime/sktime/blob/main/sktime/split/temporal_train_test_split.py
THere are multiple ways, we can start from there?
FYI @fkiraly

Muhammad-Rebaal:
> Hi @agobbifbk @Muhammad-Rebaal Can we try something like `sktime` implementation: https://github.com/sktime/sktime/blob/main/sktime/split/temporal_train_test_split.py THere are multiple ways, we can start from there? FYI @fkiraly

I've analyzed `sktime's` approach, and I think we can implement a custom **TemporalTrainTestSplitter** class in splitters.py tailored for our window-based pipeline.

This approach allows us to support anchor ("start"/"end") semantics and 3-way splits (train/val/test) using global timestamp cutoffs. The main benefit is that it simplifies the `setup()` logic across all DataModules and gives us a clean, reusable API without adding an external dependency.

Does this direction look good to you? If so, I can start on a backward-compatible PR.

agobbifbk:
What if we have 2 groups but with different lifespan? If we use temporal cutoff we may end with one series in train and the other in validation, but maybe we don't want this. We should also think to have a percentage approach that computes the cutoff PER GROUP and then use the temporal cutoff, what do you think?

Muhammad-Rebaal:
> What if we have 2 groups but with different lifespan? If we use temporal cutoff we may end with one series in train and the other in validation, but maybe we don't want this. We should also think to have a percentage approach that computes the cutoff PER GROUP and then use the temporal cutoff, what do you think?

Hi @agobbifbk  Could you please explain to me this what do you mean by 2 groups. If there are 2 groups then they should be separated first into group 1 and group 2 and then the temporal split applies. So, what is the situation you're talking about that we end up?

agobbifbk:
If group 1 goes from 2000 to 2010 and group 2 goes from 2005 to 2015 and you put a global cutoff at 2005, for the train/validation split you will end up with half of group 1 on train and half in test, the group 2 entirely in validation. But what if I want half of both in train and half in validation? I'm saying that global cutoff can be used if I want to end up in situation 1, but we need also to use the percentage and compute local cutoff if I want to end up with situation 2.

Muhammad-Rebaal:
Hi @phoeenniixx, @agobbifbk, I've made the following changes, kindly have a look:

- The user can now apply their own temporal cutoff.

If data is provided from 2020 to 2027 andthe user applied the temporal cutoff by themselves
```
dm = TslibDataModule(
    ds,
    context_length=30,
    prediction_length=7,
    split_strategy="temporal",
    temporal_cutoffs={"end_train": 2022.0, "start_test": 2023.0}
)
```
Train: [2020, 2022] 
Validation: (2022, 2023) 
Test: [2023, 2027]

- There's a default global cutoff is available if the user don't specify their cutoff in that case : `70/15/15`

agobbifbk:
Nice one @Muhammad-Rebaal! Just one comment. Since we can pass both the percentage and the global cutoff, maybe it worth to print a warning that says, when both are not None, that global cutoff is checked first. Did you check if the temporal split works also for datetime, date, and other non-numerical timestamps? From a first review it seems that the percentage split is applied per series in case of temporal split instead of per group isn't it? 
```python
for s_idx, timestamps in series_timestamps.items():
        t_min = float(np.min(timestamps))
        t_max = float(np.max(timestamps))
        t_range = t_max - t_min
        if t_range == 0:
            series_cutoffs[s_idx] = None
        else:
            train_cutoff = t_min + train_val_test_split[0] * t_range
            val_cutoff = train_cutoff + train_val_test_split[1] * t_range
            series_cutoffs[s_idx] = (train_cutoff, val_cutoff)
    all_zero = all(v is None for v in series_cutoffs.values())
```

If you do it per series, it may be that one group has some future data in validation no? Or I'm reading it wrongly?

Muhammad-Rebaal:
> Since we can pass both the percentage and the global cutoff, maybe it worth to print a warning that says, when both are not None, that global cutoff is checked first. 

Yes I've added the warning for that.

and you're right for the rest I've added a fix to tackle that.

pytorch_forecasting/data/splitters.py
```diff
@@ -0,0 +1,244 @@
...
+    train_windows, val_windows, test_windows = [], [], []
+    for w in windows:
+        end_time = _get_window_end_time(w, series_timestamps)
+        if end_time <= train_cutoff:
+            train_windows.append(w)
+        elif end_time <= val_cutoff:
+            val_windows.append(w)
+        else:
+            test_windows.append(w)
+    return train_windows, val_windows, test_windows
```
Muhammad-Rebaal: Going with default split (70,15,15)
We've 2 series : 
S1: [1,2,3,4,5,6] , S2 : [4,6,7,8] 
then combine that via `np.concatenate()` -> [1,2,3,4,5,6,7,8] -> length of 8
train_pos = global_timeline[6] -> [6]
val_pos = global_timeline[7] -> [7] 

train_cutoff = 7 
val_cutoff = 8 

Which later by looping we get 
end_time = 8 

train_windows = [1-7)
val_windows = [8] 
test_windows = [empty in this case]

agobbifbk: Almost: this concatenation is performed by group or regardless the group? Your logic works if S1 and S2 are from the same group even if this is not possible (it can not be that the same group has the same timestamp in two series). 
if S1 is a series of group 1 and S2 of group 2 you should not concatenate the TS you should perform the split per series (as it was at the beginning).

I give another example:
percentage: (80,10,10) (for convenience)
 
G1: S1: [1,2,3,4,5,6] , S2 : [7,8,9,10] --> 1-8 in train, 9 validation 10 test 
G2: S3: [11,12,13,14,15,16] , S4 : [17,18,19,20]  --> 11-18 in train, 19 validation 20 test 

So the concatenation must be per group, hard for me to read if from the code, if it is already like this good :-) 
Moreover, if the user goes for a temporal split we should suggest to use the group as categorical feature (known).


Since you are at this point, can we implement also the group-time split? In this case first you divide the groups into train, validation and test, then the groups in train are divided using the temporal logic into train validation and test!  In this case the group CAN not be used as categorical variable otherwise the model can hardly generalize on new groups! 

Hope this clarifies!

Muhammad-Rebaal: Actually its not that way implemented but I completely understand what you mean so I'd adjust that accordingly then we'd move towards the group-time split

phoeenniixx (changes requested):
Would it make sense to move the splitting logic to `split` folder and then you access it from there, in that way you dont have to implement the same thing twice

Muhammad-Rebaal:
> Would it make sense to move the splitting logic to `split` folder and then you access it from there, in that way you dont have to implement the same thing twice

Yeah, I think that's a good idea to do so but I think first the implementation should be done then we'd refactor.

pytorch_forecasting/data/splitters.py
Muhammad-Rebaal: Just one question :
> In this case first you divide the groups into train, validation and test

It would be train_val_test_split: tuple = (0.8,0.1,0.1) or train_val_test_split: tuple = (0.7, 0.15, 0.15)

agobbifbk: Maybe we should call `train_val_test_split` differently if is referred to the temporal split or group split, so we need one more parameter here

Muhammad-Rebaal: Yeah I agree,  I've added a param for that called `group-split` to handle it.

phoeenniixx:
> > Would it make sense to move the splitting logic to `split` folder and then you access it from there, in that way you dont have to implement the same thing twice
> 
> Yeah, I think that's a good idea to do so but I think first the implementation should be done then we'd refactor.

But rn (when the thing is being actually designed) - it is easier to refactor no? Once it is released, we would need deprecastions etc to make this same refactor. Also, this will overcomplicate the datamodule class once we want to add more splitting algos (which will happen soon ig)

phoeenniixx:
Also, the code-quality is failing :)

Muhammad-Rebaal:
> > > Would it make sense to move the splitting logic to `split` folder and then you access it from there, in that way you dont have to implement the same thing twice
> > 
> > 
> > Yeah, I think that's a good idea to do so but I think first the implementation should be done then we'd refactor.
> 
> But rn (when the thing is being actually designed) - it is easier to refactor no? Once it is released, we would need deprecastions etc to make this same refactor. Also, this will overcomplicate the datamodule class once we want to add more splitting algos (which will happen soon ig)

Yes you're right I didn't mean the release thing ? I mean if there are split strategies or this is the last one for this `PR` then I'd refactor them all in this PR

let's say there 3 functions of splitting and there is another that needs to be done so I wrote the 4th one and then refactor in the same PR.

What is your suggestion?

phoeenniixx:
Sorry, I dont understand what you mean? Did you plan to first reach to a consensus and then refactor in the same PR? That is also a good idea... Everything depends on you as you are the main person on this feat rn

Muhammad-Rebaal:
> Sorry, I dont understand what you mean? Did you plan to first reach to a consensus and then refactor in the same PR? That is also a good idea... Everything depends on you as you are the main person on this feat rn

No worries! Yes, I meant reaching a consensus first and then refactoring it all in the same PR. Thanks for understanding.

Muhammad-Rebaal:
Hi @agobbifbk, these are the following changes:

- Implemented `group-time-split`.
- Fix the concatination issue in the `_split_percentage` method.

Is there any other split strategies that needs to be implemented. Kindly let me know.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -959,21 +964,60 @@ def _resolve_target_normalizer(self, train_indices: torch.Tensor) -> None:
         self._target_normalizer = ScalerAdapter(normalizer)
 
     def _ensure_split(self):
```
phoeenniixx, line 966: please add all the split statergies here

phoeenniixx (changes requested):
Can you please add the docstrings for all the methods expalining exactly what they do?

Muhammad-Rebaal:
Hi @phoeenniixx @agobbifbk ,
I've added the comments and support in both data-modules.

Have we reached the consensus or is there any other splitting technique left to implement so kindly let me know. If `group-time-split` is the last one then I'd move forward towards refactoring....

agobbifbk:
For now it is sufficient, please go ahead with the refactoring, I will have a look when finished!
THX

Muhammad-Rebaal:
Hi @agobbifbk ,

I've refactored the code. Kindly, have a look

agobbifbk:
I will do it asap, but until the 28-th I will be very busy!

pytorch_forecasting/data/tests/test_splitters.py
```diff
@@ -0,0 +1,383 @@
...
+
+        total = len(train_w) + len(val_w) + len(test_w)
+        assert total == len(windows)
+        assert len(train_w) > 0
+        assert len(val_w) > 0
+        assert len(test_w) > 0
+
+    def test_identical_timestamps_fallback(self):
+        """When all timestamps are identical, should fall back gracefully."""
+        timestamps = np.array([5, 5, 5, 5, 5])
```
agobbifbk, line 112: This should never happen but i prefer 1 test more than 1 less

pytorch_forecasting/data/splitters.py
```diff
@@ -0,0 +1,298 @@
...
+    list[tuple[int, int, int, int]],
+    list[tuple[int, int, int, int]],
+]:
+    """Split windows temporally to prevent data leakage.
+    Supports two modes:
+    - **Absolute mode** (when ``temporal_cutoffs`` is provided):
+      Uses global timestamp boundaries. All series share the same cutoffs.
+    - **Percentage mode** (default, when ``temporal_cutoffs`` is None):
+      Computes cutoffs from a global timeline across all series,
+      ensuring a single boundary that prevents cross-series leakage.
```
agobbifbk: This confounds me: if percentage is defined, it should act per group no? What do you mean by `global timeline across all series` do you mean `across all series for the same group`?

pytorch_forecasting/data/splitters.py
```diff
@@ -0,0 +1,298 @@
...
+        (no gap between val and test).
+
+    Returns
+    -------
+    train_windows, val_windows, test_windows
+        Three lists of window tuples.
+    """
+    if not windows:
+        return [], [], []
+    if temporal_cutoffs is not None and train_val_test_split != (0.7, 0.15, 0.15):
```
agobbifbk: Why this check on the percentage? I can not have 0.6 0.2 0.2?

Muhammad-Rebaal: Yes, I agree with you and I've removed that hardcoded check completely.

pytorch_forecasting/data/splitters.py
Muhammad-Rebaal: Yes, you're right, the docstring was misleading. The implementation already computes cutoffs per series from each series' own timeline, not globally. Updated the docstring to reflect this.

Muhammad-Rebaal:
Hi @agobbifbk, these are the changes made : 
- Corrected the docstring.
- Removed the test code for extra check of hardcode split block

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -135,6 +135,9 @@ def __init__(
         batch_size: int = 32,
         num_workers: int = 0,
         train_val_test_split: tuple = (0.7, 0.15, 0.15),
+        split_strategy: str = "random",
+        temporal_cutoffs: dict[str, float] | None = None,
+        group_split: tuple[float, float, float] = (0.7, 0.15, 0.15),
```
phoeenniixx, line 140: Can you please add docstrings for this?

pytorch_forecasting/data/splitters.py
phoeenniixx, line 1: this should go in `split` folder?

phoeenniixx (changes requested):
Thanks, I just skimmed the PR, so based on it, I have some suggestions/requests:
- Please add docstrings for all the params (explaining the splt startegies etc)
- Why not make splitters as classes and accept them in `__init__` in place of it being split strategies?
   - This will make the design more extensible (you wont have to add more code to the DMs whenever you add a new splitter)
   - Also, i can see some "split strategy specific" params like `temporal_cutoff` etc
      - I think we should accept them as params to the splitter classes instead. The diff inside the DM should not be that large, just accepting the classes and rest happens inside the classes itself.

What do you think on this? Please feel free to object

Also FYI @agobbifbk @fkiraly

agobbifbk:
Ye at this point is more elegant and maintainable to have a class!

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
Muhammad-Rebaal: ok sure, I'll add that

pytorch_forecasting/data/splitters.py
Muhammad-Rebaal: I think you forget to specify the codeblock?

Muhammad-Rebaal:
Thanks for the review,

> * Why not make splitters as classes and accept them in `__init__` in place of it being split strategies?

Do you mean something like this :
```
class SplitterA:
    pass

class SplitterB:
    pass

class DataModule:
    def __init__(self, splitter):
        self.splitter = splitter

    def verify_splitter(self):
        return self.splitter.__class__.__name__


module_a = DataModule(splitter=SplitterA())
print(module_a.verify_splitter()) 

module_b = DataModule(splitter=SplitterB())
print(module_b.verify_splitter()) 
```

Muhammad-Rebaal:
Hi @phoeenniixx , @agobbifbk !
I've refactored the code kindly have a look.

PR #2104, opened 2026-02-27
[MNT] Improve error message for invalid reduction argument in groupby_apply

Gyanam1310:
#### Reference Issues/PRs

None.

This PR improves the error message raised when an invalid `reduction`
argument is passed to `groupby_apply`.

Previously, the error message only stated that the reduction was unknown.
This update clarifies the expected valid options: {'mean', 'sum'}.

No functional behavior is changed.

- Clarity and wording of the updated error message.
- Whether the expected options are correctly specified.

No. This change only improves the error message wording and does not
modify functionality.

#### Any other comments?

This is a minor maintenance improvement to make debugging clearer
for users passing invalid arguments.

### DIFF
--- a/pytorch_forecasting/utils/_utils.py
+++ b/pytorch_forecasting/utils/_utils.py
@@ -68,7 +68,9 @@ def groupby_apply(
     elif reduction == "sum":
         reduce = torch.sum
     else:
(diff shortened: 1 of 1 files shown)


phoeenniixx (approved):
Thanks!

PR #2113, opened 2026-02-28
[BUG] Fix dimension assumption in QuantileLoss.loss for 2D inputs

### LINKED ISSUE #1795: [BUG]: Error from non existent dim in QuantileLoss.loss
**WHERE**
In pytorch_forecasting.metrics.quantile
In QuantileLoss.loss (shown below)

**WHAT**
torch.cat with dim=2 does not work as tensors are expected to only have 2 dimensions.
It works fine with dim=1.

(By the way I'm not sure about the criterion's scaling "2 *" logic. I don't find it in other references)

**CODE**
def loss(self, y_pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
    # calculate quantile loss
    losses = []
    for i, q in enumerate(self.quantiles):
        errors = target - y_pred[..., i]
        losses.append(torch.max((q - 1) * errors, q * errors).unsqueeze(-1))
    losses = _2 *_ torch.cat(losses, **dim=2**)

    return losses

Kundan-CR7:
#### Reference Issues/PRs

Fixes #1795

This PR fixes a dimension assumption in `QuantileLoss.loss`.

The current implementation concatenates per-quantile losses using:

    torch.cat(losses, dim=2)

This assumes that `y_pred` is always 3D (e.g., [batch, time, quantiles]).  
However, when `y_pred` is 2D (e.g., [batch, quantiles]), this raises:

    IndexError: Dimension out of range

To make the implementation dimension-safe, concatenation has been changed to:

    torch.cat(losses, dim=-1)

Since the quantile dimension is always the last dimension of `y_pred`, using `dim=-1` ensures compatibility with both:

- 2D inputs: [batch, quantiles]
- 3D inputs: [batch, time, quantiles]

This resolves the issue without altering the loss computation logic.

- Correctness of replacing `dim=2` with `dim=-1`
- Consistency with expected quantile dimension semantics
- Regression test coverage for 2D inputs
- Backward compatibility with existing 3D use cases

Yes.

Added regression tests in `test_quantile_loss.py` to verify:

- 2D input no longer raises `IndexError`
- 3D input behavior remains unchanged
- Output shapes are correct for both cases

All existing tests pass locally:

    862 passed, 2 skipped

#### Any other comments?

This change only removes the fixed-dimension assumption and makes concatenation robust to varying input dimensionality.  
No changes were made to the quantile loss computation itself.

#### PR checklist

- [x] Added/modified tests


PranavBhatP:
@Kundan-CR7 thanks for the contribution! Your change is noted, and I think it would make sense to have `dim=-1` instead of `dim=2`. But the reason why `dim=2` was because it was ensured that we always have the input as `[batch, time, quantiles]` into the `torch.cat` function. That is my understanding, of course your change still stands.

Kundan-CR7:
@PranavBhatP Thanks a lot for the clarification and the quick review!

That makes sense — I understand the original assumption about the input always being `[batch, time, quantiles]`. I’ve moved the regression test into the existing `tests/test_metrics.py` file as suggested to keep the test structure consistent.

This is my first PR in this repository, so I really appreciate the feedback and guidance.

phoeenniixx:
I have a doubt:
should the `QuantileLoss` support 2D input? I mean we always assume it to be 3D no?

Issue #2066, opened 2026-02-20
[DOC] Migrate the documentation to numpydocstyle


phoeenniixx:
#### Describe the issue linked to the documentation
Currently in some older parts of the `pytorch-forecasting`, we still have Google style of docstrings, but I think we should move it numpydoc style now.
Some areas I could find the issues:

- [ ] [`_timeseries.py`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/data/timeseries/_timeseries.py)- @karthikch947 is working on this starting from this line: https://github.com/sktime/pytorch-forecasting/blob/602ee157d18bd00ea7e1cfddeb0c8184c138976f/pytorch_forecasting/data/timeseries/_timeseries.py#L1040

- [ ] [`examples.py`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/data/examples.py) @AyushDineshRathi 
- [ ] [`pytorch_forecasting/layers/_attention`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/layers/_attention/) - most files - @omkar-334 is working on this
- [ ] [`pytorch_forecasting/layers/_decomposition/_series_decomp.py`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/layers/_decomposition/_series_decomp.py) @Meet-Ramjiyani-10 
- [x] [`pytorch_forecasting/layers/_embeddings/`](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/layers/_embeddings) - most files - @Soham-47 is working on this #2079 
- [x] [`pytorch_forecasting/layers/_encoders/`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/layers/_encoders) - most files - @Skvmqq #2121 
- [x] [`pytorch_forecasting/layers/_filter/`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/layers/_filter) - most files - @AyushDineshRathi #2116 
- [ ] [`pytorch_forecasting/layers/_output/`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/layers/_output/) - most files @Meet-Ramjiyani-10 
[...]

PR #2115, opened 2026-03-01
[DOC] migrate baseline.py docstrings to numpydocstyle 

amruth6002:
#### Reference Issues/PRs

Addresses #2066

Migrates the docstrings in [pytorch_forecasting/models/baseline.py](cci:7://file:///home/amruth/amruth/projects/pytorch-forecasting/pytorch_forecasting/models/baseline.py:0:0-0:0) from Google style  to NumPy docstring style , as part of the repo-wide numpydocstyle migration tracked in #2066.

No — this is a documentation-only change with no functional code modifications.

#### Any other comments?

None.

#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/baseline.py
+++ b/pytorch_forecasting/models/baseline.py
@@ -35,11 +35,15 @@ class Baseline(BaseModel):
         """
         Network forward pass.
 
-        Args:
-            x (Dict[str, torch.Tensor]): network input
+        Parameters
+        ----------
+        x : dict of {str : torch.Tensor}
+            network input
 
-        Returns:
-            Dict[str, torch.Tensor]: network outputs
+        Returns
+        -------
+        dict of {str : torch.Tensor}
+            network outputs
         """
         if isinstance(x["encoder_target"], tuple | list):  # multiple targets
             prediction = [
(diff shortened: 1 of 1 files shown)


amruth6002:
few test case failed with HTTP Error 504: Gateway Timeout , could you please re-run them. @phoeenniixx @PranavBhatP @fkiraly

pytorch_forecasting/models/baseline.py
```diff
@@ -35,11 +35,15 @@ def forward(self, x: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
         Network forward pass.
 
-        Args:
-            x (Dict[str, torch.Tensor]): network input
-
-        Returns:
-            Dict[str, torch.Tensor]: network outputs
+        Parameters
+        ----------
+        x : dict of {str : torch.Tensor}
```
phoeenniixx: should it be `Dict[str, torch.Tensor]`?

amruth6002: the numpydocstyle standard for dictionaries is technically dict of {key_type: value_type}

amruth6002: but if we need to change it to square bracket i will do it, should i?

phoeenniixx: I mean its a common convention we've been following since the start, that's why I suggested it. I understand what you mean, but I would prefer following already set convention in docs to prevent any misunderstandings :)

phoeenniixx (approved):
Thanks!

PR #2122, opened 2026-03-01
[ENH]added multiloss support for v2 interface of base model class

Soham-47:
Added MultiLoss support for to_prediction and to_quantiles methods of base model v2 class

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries.py
@@ -1003,7 +1003,9 @@ class TimeSeriesDataSet(Dataset):
         ), "Timeseries index should be of type integer"
         # numeric categoricals which can cause issues in tensorborad logging
         category_columns = data.head(1).select_dtypes("category").columns
-        object_columns = data.head(1).select_dtypes(object).columns
+        object_columns = (
+            data.head(1).select_dtypes(include=["object", "string"]).columns
+        )
         for name in self.flat_categoricals:
             if name not in data.columns:
                 raise KeyError(f"variable {name} specified but not found in data")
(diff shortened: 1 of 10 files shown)


pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -139,23 +141,23 @@ def predict(
...
         """Converts raw model output to point forecasts."""
-        # todo: add MultiLoss support
         try:
-            out = self.loss.to_prediction(out["prediction"], **kwargs)
-        except TypeError:  # in case passed kwargs do not exist
-            out = self.loss.to_prediction(out["prediction"])
-        return out
+            return self.loss.to_prediction(out["prediction"], **kwargs)
+        except TypeError:
+            return self.loss.to_prediction(out["prediction"])
```
phoeenniixx: What is the difference between the changes? I cant see any logical difference?

Soham-47: it adds multivariate support by updating prediction return types to handle multiple targets. Apart from that, nothing's changed.

phoeenniixx: I understand that, but I am unable to spot the difference. I mean before it returned `self.loss.to_prediction(out["prediction"], **kwargs)` and `self.loss.to_prediction(out["prediction"])` ( in case of `TypeError`), the code is still doing the same. If there is no logical difference, why are we making this change?

phoeenniixx: Please correct me if I have missed something here

Soham-47: yeah there isn't any changes ......i will just revert that

PranavBhatP:
Hi, if the PR is still not ready for review, you can revert to a draft PR, I don't see any reviewable code right now.

Issue #2126, opened 2026-03-02
[ENH] Add LightTS model to v2 interface


Sylver-Icy:
LightTS (from the THUML Time-Series-Library) is currently not available in pytorch-forecasting. Adding it would expand the library with a lightweight transformer-based architecture that has shown strong performance in long-term time-series forecasting tasks.
I would like to implement LightTS using the new _base_model_v2.BaseModel interface.

Migrating existing v1 models to v2 was considered, but several of those are already being worked on. Adding LightTS avoids duplication and expands model coverage.

LightTS originates from the THUML Time-Series-Library and is designed as a lightweight and efficient transformer variant for forecasting.

PR #2135, opened 2026-03-04
[ENH] Add LightTS model implementation (v2)

Sylver-Icy:
#### Reference Issues/PRs
#2126 

#### What does this implement?

This PR adds an implementation of the LightTS model for the PyTorch Forecasting v2 model framework.
Implemented the LightTS architecture based on the original paper.
The implementation follows the same design and integration patterns used by existing models such as DLinear.

Yes.

Added `tests/test_models/test_lightts_v2.py` which includes tests for:

- Model initialization with different hyperparameters
- Forward pass prediction shape
- Quantile loss output shape
- Univariate forecasting behavior

#### Any other comments?

The implementation aims to stay consistent with existing v2 model implementations (e.g., DLinear) in structure and integration patterns.

Feedback on architectural alignment with the original LightTS design is welcome.


pytorch_forecasting/models/lightts/_lightts_pkg_v2.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
```
phoeenniixx: I dont think you need this now. Please have a look at other models: like [timexer](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py)

pytorch_forecasting/models/lightts/_lightts_pkg_v2.py
```diff
@@ -0,0 +1,148 @@
...
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class LightTS_pkg_v2(Base_pkg):
+    """LightTS package container."""
+
+    _tags = {
+        "info:name": "LightTS",
+        "info:compute": 2,
+        "authors": ["Abhinav Singh"],
```
phoeenniixx: I think you should add you GH acc here

pytorch_forecasting/models/lightts/_lightts_v2.py
```diff
@@ -0,0 +1,283 @@
...
+
+import torch
+import torch.nn as nn
+from torch.optim import Optimizer
+
+from pytorch_forecasting.metrics import QuantileLoss
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class _IEBlock(nn.Module):
```
phoeenniixx: should this go into [`layers`](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/layers) module?

phoeenniixx (dismissed):
Thanks!
Please add docstrings to each class and atleast to all the public methods.

Sylver-Icy:
Thanks for the feedback! I’ll add docstrings
For the `_get_test_datamodule_from` part I’ll check the implementation in `timexer` and adjust/remove it if needed.

And good point about `_IEBlock`  moving it to the `layers` module makes sense. I’ll update the structure accordingly.

PranavBhatP (dismissed):
LGTM!

Sylver-Icy:
@phoeenniixx I kept _IEBlock inside the lightts module for now since it’s currently a private helper used only by lightts. In this repo it looks like layers is mainly used for reusable/shared components across multiple models

since no other model is using it yet, I thought keeping it local would keep things a bit simpler for now xd
If you’d still prefer it moved to layers (or another shared location) just lmk I’m happy to update the structure

PranavBhatP:
> since no other model is using it yet, I thought keeping it local would keep things a bit simpler for now xd
If you’d still prefer it moved to layers (or another shared location) just lmk I’m happy to update the structure

please move it to the layers folders, all layer modules are meant to be a part of it regardless of whether it is shared or not.

pytorch_forecasting/layers/_blocks/_ie_block.py
```diff
@@ -0,0 +1,74 @@
+import torch
+import torch.nn as nn
+
+
+class _IEBlock(nn.Module):
```
phoeenniixx: I think this should be public now?

pytorch_forecasting/models/lightts/_lightts_v2.py
```diff
@@ -0,0 +1,304 @@
...
+from torch.optim import Optimizer
+
+from pytorch_forecasting.layers._blocks import _IEBlock
+from pytorch_forecasting.metrics import QuantileLoss
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class LightTS(TslibBaseModel):
+    """
+    LightTS model for long-term time-series forecasting.
```
phoeenniixx, line 19: Please add the references (Paper, code, etc) here

pytorch_forecasting/models/lightts/_lightts_pkg_v2.py
```diff
@@ -0,0 +1,94 @@
...
+    Package container describing the LightTS model.
+
+    This class registers metadata, links the model implementation,
+    and provides helper utilities used during testing.
+    """
+
+    _tags = {
+        "info:name": "LightTS",
+        "info:compute": 2,
+        "authors": ["Sylver-Icy"],
```
phoeenniixx: Add the authors as well here (from the GH?)

pytorch_forecasting/models/lightts/_lightts_v2.py
```diff
@@ -0,0 +1,313 @@
...
+            return output
+
+        if output.size(-1) != 1:
+            raise ValueError(
+                "Quantile output expects a single target channel, but got "
+                f"{output.size(-1)}."
+            )
+
+        batch_size = output.shape[0]
+        output = output.squeeze(-1)
```
phoeenniixx: why are we using it? I mean `reshape` will handle the shapes anyhow - not sure if this squeeze helps or not?

pytorch_forecasting/models/lightts/_lightts_v2.py
```diff
@@ -0,0 +1,313 @@
...
+            output = output[:, :, target_indices]
+
+        return self._reshape_output(output)
+
+    def _reshape_output(self, output: torch.Tensor) -> torch.Tensor:
+        """
+        Reshape model output to match v2 prediction format.
+        """
+        if self.n_quantiles is None:
+            return output
```
phoeenniixx, line 277: I have a doubt, if `n_quantiles` is 0 then what will be the shape here? I think for point prediction losses the shape should be `(batch,time,1)` - as much as I remember. @PranavBhatP please correct me if I am wrong here

phoeenniixx (changes requested):
Added some doubts and suggestions

pytorch_forecasting/models/lightts/_lightts_v2.py
Sylver-Icy: `n_quantiles` cannot actually be 0 in this implementation. It is either
`None`  or
`len(self.loss.quantiles)` when using `QuantileLoss`

PranavBhatP: > I think for point prediction losses the shape should be (batch,time,1) - as much as I remember. 

No, point prediction losses simply expect `(batch_size, timesteps)` - a 2d tensor.

phoeenniixx: i meant the output format of the `forward` it was 3D no? When using a PP loss

PranavBhatP: yes model outputs are standardized as a 3d tensor.

phoeenniixx: > `n_quantiles` cannot actually be 0 in this implementation. It is either
> `None` or
> `len(self.loss.quantiles)` when using `QuantileLoss`

Actually that's what I was asking: if `n_quantiles` is `None`, what will be the shape here? As I see if `n_quantiles` is `None`, you dont reshape the output, is it in the expected shape already here?

Sorry for not being clear before

Sylver-Icy: umm when n_quantiles is none, the encoder already returns (batch, prediction_length, target_dim) _reshape_output() only applies reshaping for quantile outputs, so it correctly returns the tensor unchanged for point prediction losses

pytorch_forecasting/layers/_blocks/_ie_block.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if input_dim <= 0:
+            raise ValueError("input_dim must be positive.")
+        if hidden_dim <= 0:
+            raise ValueError("hidden_dim must be positive.")
+        if output_dim <= 0:
+            raise ValueError("output_dim must be positive.")
+        if num_nodes <= 0:
+            raise ValueError("num_nodes must be positive.")
+
```
phoeenniixx, line 40: you should collect all the `__init__` params like `self.param=param`. Please do so wherever applicable

pytorch_forecasting/models/lightts/_lightts_v2.py
```diff
@@ -0,0 +1,312 @@
...
+                ],
+                dim=1,
+            )
+
+        # Highway (autoregressive) branch.
+        highway = self.ar(x.permute(0, 2, 1)).permute(0, 2, 1)
+
+        # Continuous sampling.
+        x1 = x.reshape(batch_size, self.num_chunks, self.chunk_size, n_features)
+        x1 = x1.permute(0, 3, 2, 1).reshape(-1, self.chunk_size, self.num_chunks)
```
phoeenniixx, line 252: I have a doubt: I remember for we dont have 4D tensors, we actually use 3D tensors and a `list` of 3D tensors if multi-target, then when does this list turned to 4D tensor?

phoeenniixx (changes requested):
I think there is some logic blunder, I am not sure, but please check it. I am not sure if we handle 4D tensors?
FYI @PranavBhatP

pytorch_forecasting/models/lightts/_lightts_v2.py
Sylver-Icy: The 4D tensor here is only an internal LightTS sampling view, not dataloader/model input contract. The input to _encoder is still 3D: (batch, context_length, features)
or continuous sampling we reshape it to (B, num_chunks, chunk_size, N), then immediately reshape it back to 3D as (B * N, chunk_size, num_chunks) before passing it to IEBlock. Interval sampling does the same. So no multi-target list is being converted to 4D here

pytorch_forecasting/layers/_blocks/_ie_block.py
Sylver-Icy: kay will do that

Issue #1992, opened 2025-11-27
[ENH] Wishlist of forecasters and integrations


fkiraly:
We should start adding models to the `v2` interface as designed (possibly with `predict` adaptation later on).

A good starting point is to look at the implementations of DLinear, SAMformer, and TFT in the v2 implementation.

The "recipe" is trying to add a new estimator which is compliant with the v2 interface. For each model, it should also be decided whether interfacing (import and dependency management) or forking (copy and adapt) is better.

Layers go in `layers`, the models go in `models`, and models import layers from `layers`.

Contributors should pick one model and post about it here:

List of models to look at:

* current models in `pytorch-forecasting` that do not have a v2 implementation
* the DSIPTS package https://github.com/DSIP-FBK/DSIPTS
* the thuml package https://github.com/thuml/Time-Series-Library
* forecasting foundation models from this list https://github.com/sktime/sktime/issues/6177 - note that this is difficult until weight handling is designed in v2, but API design work is appreciated!

Also, anyone should suggest models below to add to this wishlist.

Vishal35198:
Hi! I would like to contribute by adding a new model to the v2 interface.

I am interested in implementing **PatchTST** based on the implementation from:
https://github.com/thuml/Time-Series-Library

My tentative plan is:
- Study the current v2 implementations such as DLinear, SAMformer, and TFT to follow the expected structure.
- Separate reusable components into the `layers/` module.
- Implement the estimator under `models/` and ensure it imports the necessary components from `layers/`.
- Evaluate whether interfacing with the upstream implementation or forking and adapting the model would be more appropriate.

Please let me know if this model is suitable to work on or if there are any specific guidelines I should follow before starting.

Vishal35198:
> Hi! I would like to contribute by adding a new model to the v2 interface.
> 
> I am interested in implementing **PatchTST** based on the implementation from: https://github.com/thuml/Time-Series-Library
> 
> My tentative plan is:
> 
> * Study the current v2 implementations such as DLinear, SAMformer, and TFT to follow the expected structure.
> * Separate reusable components into the `layers/` module.
> * Implement the estimator under `models/` and ensure it imports the necessary components from `layers/`.
> * Evaluate whether interfacing with the upstream implementation or forking and adapting the model would be more appropriate.
> 
> Please let me know if this model is suitable to work on or if there are any specific guidelines I should follow before starting.

@fkiraly @phoeenniixx @christy

[...]

PR #2080, opened 2026-02-21
[ENH] Relax naming convention in test_pkg_linkage

PalakB09:
#### Reference Issues/PRs

Fixes #2065

#### What does this implement/fix? Explain your changes.

This PR relaxes the naming constraint in `test_pkg_linkage` within `test_all_estimators_v2.py`.

The test strictly required the package class name to follow: `estimatorClass_v2` -> `estimatorClass_v2_pkg_v2`
However, when migrating models from v1 to v2, this constraint becomes restrictive and leads to redundant naming.

A better and clearer naming pattern would be `estimatorClass_v2` -> `estimatorClass_pkg_v2`

The test has been updated to allow both valid patterns:
- `estimatorClass_v2_pkg_v2`
- `estimatorClass_pkg_v2` (when the estimator class ends with `_v2`)

#### Did you add any tests for the change?
The existing `test_pkg_linkage` test was updated

#### Any other comments?
I was facing the similar issue [here](https://github.com/sktime/pytorch-forecasting/pull/2043#discussion_r2831361892). After applying the changes all the tests pass locally.

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [x] Added/modified tests
- [x]  Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.
  To run hooks independent of commit, execute `pre-commit run --all-files`

PR #2136, opened 2026-03-04
[ENH] Add v2 interface support for RecurrentNetwork (RNN)

### LINKED ISSUE #2128: [ENH] Add v2 interface support for RecurrentNetwork (RNN)
Referencing #1992 

Adding v2 interface support for the RecurrentNetwork (RNN) model .

The plan is to 
- Creating `_rnn_v2.py` following the TslibBaseModel pattern (similar to DLinear, SAMformer v2)
- Creating `_rnn_pkg_v2.py` with appropriate tags and test parameters
CC: @fkiraly 
@phoeenniixx

Meet-Ramjiyani-10:
#### Reference Issues/PRs

Closes #2128. Part of #1992.

Adds v2 interface support for the RecurrentNetwork (RNN) model.

- Added `_rnn_v2.py` implementing `RecurrentNetwork_v2` using `TslibBaseModel`
- Added `_rnn_pkg_v2.py` with tags, datamodule, and test parameters
- Supports both LSTM and GRU cell types
- Supports QuantileLoss for probabilistic forecasting

I followed the v2 pattern established by DLinear and SAMformer and have made no changes to existing v1 implementation.

-  forward pass input preparation
- Whether the output shape is consistent with other v2 models
- Tags in the pkg file

Test parameters added in `get_test_train_params` covering LSTM, GRU, single layer, and multi-layer configurations.

#### Any other comments?

Happy to iterate based on feedback.

#### PR checklist

- [✓ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

- [ ✓] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.

### DIFF
--- a/pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
+++ b/pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
@@ -9,7 +9,7 @@ class RecurrentNetwork_v2_pkg(Base_pkg):
     """RecurrentNetwork v2 package container."""
 
     _tags = {
-        "info:name": "RecurrentNetwork_v2",
+        "info:name": "RecurrentNetwork",
         "info:compute": 2,
         "authors": ["Meet-Ramjiyani-10"],
         "capability:exogenous": True,
--- a/pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
+++ b/pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
@@ -1,0 +1,139 @@
+"""
+Package container for RecurrentNetwork v2 model.
+"""
+
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class RecurrentNetwork_pkg_v2(Base_pkg):
+    """RecurrentNetwork v2 package container."""
+
+    _tags = {
+        "info:name": "RecurrentNetwork",
+        "info:compute": 2,
+        "authors": ["Meet-Ramjiyani-10"],
+        "capability:exogenous": True,
+        "capability:multivariate": True,
+        "capability:pred_int": True,
+        "capability:flexible_history_length": True,
+        "capability:cold_start": False,
+    }
+
+    @classmethod
+    def get_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models.rnn._rnn_v2 import RecurrentNetwork_v2
+
+        return RecurrentNetwork_v2
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
+        """Create test dataloaders from trainer_kwargs."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+        from pytorch_forecasting.tests._data_scenarios import (
+            data_with_covariates_v2,
+            make_datasets_v2,
+        )
+
+        data_with_covariates = data_with_covariates_v2()
+        data_loader_default_kwargs = dict(
+            target="target",
+            group_ids=["agency_encoded", "sku_encoded"],
+            add_relative_time_idx=True,
+        )
+
+        data_loader_kwargs = trainer_kwargs.get("data_loader_kwargs", {})
+        data_loader_default_kwargs.update(data_loader_kwargs)
+
+        datasets_info = make_datasets_v2(
+            data_with_covariates, **data_loader_default_kwargs
+        )
+
+        training_dataset = datasets_info["training_dataset"]
+        validation_dataset = datasets_info["validation_dataset"]
+
+        context_length = data_loader_kwargs.get("context_length", 8)
+        prediction_length = data_loader_kwargs.get("prediction_length", 2)
+        batch_size = data_loader_kwargs.get("batch_size", 2)
+
+        train_datamodule = TslibDataModule(
+            time_series_dataset=training_dataset,
+            context_length=context_length,
+            prediction_length=prediction_length,
+            add_relative_time_idx=data_loader_kwargs.get("add_relative_time_idx", True),
+            batch_size=batch_size,
+            train_val_test_split=(0.8, 0.2, 0.0),
+        )
+
+        val_datamodule = TslibDataModule(
+            time_series_dataset=validation_dataset,
+            context_length=context_length,
+            prediction_length=prediction_length,
+            add_relative_time_idx=data_loader_kwargs.get("add_relative_time_idx", True),
+            batch_size=batch_size,
+            train_val_test_split=(0.0, 1.0, 0.0),
+        )
+
+        test_datamodule = TslibDataModule(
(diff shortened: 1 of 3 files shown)


pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
```
phoeenniixx: I dont think you need this method anymore? Please have a look at other pkgs (like [timexer](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer_pkg_v2.py)) for more info

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
```diff
@@ -0,0 +1,139 @@
...
+        from pytorch_forecasting.models.rnn._rnn_v2 import RecurrentNetwork_v2
+
+        return RecurrentNetwork_v2
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
```
phoeenniixx: Should it be `EncoderDecoderDataModule`?

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_test_train_params(cls):
+        """
+        Return testing parameter settings for the trainer.
+
+        Returns
+        -------
+        params : list of dict
+            Parameters to create testing instances of the class.
+        """
+        from pytorch_forecasting.metrics import MAE, SMAPE, QuantileLoss
```
phoeenniixx, line 48: Did you mean to use them in the params?
I think it would be good if we could chekc which type of losses this can support

phoeenniixx (changes requested):
Thanks!
Added some comments

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
Meet-Ramjiyani-10: thanks  for pointing. After checking _timexer_pkg_v2.py. this method is not needed . Will remove it

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
Meet-Ramjiyani-10: @phoeenniixx  After checking _timexer_pkg_v2., it also uses TslibDataModule in get_datamodule_cls. Should I still switch to EncoderDecoderDataModule, or is TslibDataModule correct here?

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
Meet-Ramjiyani-10: i will update get_test_train_params to include MAE, SMAPE, and QuantileLoss directly in the params to verify loss compatibility,  as in  _timexer_pkg_v2.py

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
PranavBhatP: Please switch to `EncoderDecoderDataModule` for models with v1 native implementations. The rule of thumb is to stick to `TslibDataModule` only for direct migrations of models from `thuml`(tslib).

Meet-Ramjiyani-10: @PranavBhatP  
thanks for clarifying!. But it seems EncoderDecoderDataModule doesn't exist in the codebase yet. Should I keep TslibDataModule for now, or is there a different existing class I should use?

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
Meet-Ramjiyani-10: @phoeenniixx  done 👍

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
Meet-Ramjiyani-10: @phoeenniixx  done 👍

pytorch_forecasting/models/rnn/_rnn_pkg_v2.py
PranavBhatP: `EncoderDecoderTimeseriesDataModule` is the exact name of the D2 data module I am referring, it is interchangeably referred to as `EncoderDecoderDataModule` (u get what i mean?).

Meet-Ramjiyani-10: yes @PranavBhatP got it .  I was searching for EncoderDecoderDataModule as a single class and couldn't find it , making the changes.. EncoderDecoderTimeSeriesDataModule

Meet-Ramjiyani-10:
@phoeenniixx  and @PranavBhatP 
I’ve updated the RNN v2 package so that info:name is "RecurrentNetwork_v2", which fixes the first part of test_pkg_linkage. However, I’m still seeing failures from the second assertion:

assert object_pkg.__name__ == object_class.__name__ + "_pkg_v2", msg
For RNN v2 we currently have:

object_class.__name__ == "RecurrentNetwork_v2"
object_pkg.__name__ == "RecurrentNetwork_v2_pkg"
So the test expects "RecurrentNetwork_v2_pkg_v2", which doesn’t match the current naming. For the other v2 models (e.g. TFT, DLinear, Samformer), the class name stays the same and only the package class gets the _pkg_v2 suffix, so this rule fits them but not RNN v2.

Could you please advise what naming you’d prefer here? I see two options:

Rename the model class back to RecurrentNetwork and keep the package as RecurrentNetwork_pkg_v2 (matching the convention used for the other v2 models), or
Keep RecurrentNetwork_v2 / RecurrentNetwork_v2_pkg and relax/special‑case the test for this model.
Once I know which direction you’d like, I can adjust the code and tests accordingly.

phoeenniixx:
Yes this is a known issue, see #2080  by @PalakB09. You can:
- Either wait for some time and update once #2080 is merged
- Or maybe name the model as `RNN` in place of `RecurrentNetwork_v2` so that the names doesnt class with v1

The decision is depends on you
FYI @PranavBhatP @fkiraly

Meet-Ramjiyani-10:
okay, for now i will rename the model class to RNN

Meet-Ramjiyani-10:
@phoeenniixx I have updated the implementation and renamed the model to `RNN` with `RNN_pkg_v2` following the current v2 naming convention.
All CI checks are passing now. Please let me know if any further adjustments are needed.

PR #2139, opened 2026-03-04
[MNT] Remove `pandas 3` warnings

phoeenniixx:
`pandas 3` introduced some deprecations which were clogging up the CI. This PR tries to remove these warnings
Currently removed warnings:

- [x] ```In a future version, the keys of `groups` will be a tuple with a single element, e.g. (group_id,) , instead of a scalar, e.g. group_id, when grouping by a list with a single element. Use ``df.groupby(by='a').groups`` instead of ``df.groupby(by=['a']).groups`` to avoid this warning```
- [x] ```For backward compatibility, 'str' dtypes are included by select_dtypes when 'object' dtype is specified. This behavior is deprecated and will be removed in a future version. Explicitly pass 'str' to `include` to select them, or to `exclude` to remove them and silence this warning. ```
- [x] ```The 'mode.copy_on_write' option is deprecated. Copy-on-Write can no longer be disabled (it is always enabled with pandas >= 3.0), and setting the option has no impact. This option will be removed in pandas 4.0.```

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries.py
@@ -1003,7 +1003,9 @@ class TimeSeriesDataSet(Dataset):
         ), "Timeseries index should be of type integer"
         # numeric categoricals which can cause issues in tensorborad logging
         category_columns = data.head(1).select_dtypes("category").columns
-        object_columns = data.head(1).select_dtypes(object).columns
+        object_columns = (
+            data.head(1).select_dtypes(include=["object", "string"]).columns
+        )
         for name in self.flat_categoricals:
             if name not in data.columns:
                 raise KeyError(f"variable {name} specified but not found in data")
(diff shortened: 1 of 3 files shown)


phoeenniixx:
Not sure about the `LeafSpec()` warning, but I think that is coming from upstream? maybe `torch` or `lightning`? See https://github.com/Lightning-AI/pytorch-lightning/issues/21326

phoeenniixx:
Hi @fkiraly @PranavBhatP @fnhirwa, can you please review the PR?

fkiraly (approved):
Looks good to me - I think we can now release another version with `pandas 3` support - should we?

phoeenniixx:
> I think we can now release another version with `pandas 3` support

Yes, I was going to propose this after this PR got merged :)
Just one thing I am confused about: should it be a minor release or a patch release?

PR #2165, opened 2026-03-09
[ENH] `Units_v2` Model added

### LINKED ISSUE #2158: [ENH] Add v2 interface support for UniTS
Referencing #1992

This issue tracks adding v2 interface support for the UniTS (Unified Time Series Model).

UniTS is a state-of-the-art framework that provides a unified approach to time series forecasting, classification, and imputation. Given the v2 interface's goal of modularity, UniTS is a significant addition as it represents the shift toward foundation-like models in the forecasting space.

**Plan:**
- Implement the core UniTS Transformer blocks in the `layers` directory.
- Implement the forecasting-specific estimator in the `models` directory, following the v2 pattern used by `DLinear` and `TFT`.
- Ensure compliance with the v2 requirement for explicit dependency management/interfacing.

**Reference:**
Zhang, M., et al. (2024). UniTS: Building a Unified Time Series Model. [https://arxiv.org/abs/2403.00131](https://arxiv.org/abs/2403.00131)

CC: @fkiraly @PranavBhatP @phoeenniixx

Muhammad-Rebaal:
Fixes #2158 

Hi @fkiraly , @phoeenniixx, @PranavBhatP !

I have implemented the UniTS (Unified Time Series Model)  within the PyTorch Forecasting v2 architecture.

Could you please review the PR?

- **Isolated Layer Abstraction (`pytorch_forecasting/layers/_units/`)**: Implemented the core neural network components (`_PatchEmbedding`, `_PositionalEncoding`, `_TransformerBlock`) completely isolated from the standard base estimating logic to strictly adhere to the project's v2 layer standards.
- **Model Construction (`pytorch_forecasting/models/units/`)**: Created the main `UniTS` wrapper which inherently inherits from `TslibBaseModel` and handles feature input alignments, channel-independence, and forward pass routing to the abstracted layers. 
- **V2 Package Management**: Established the `UniTS_pkg_v2` class representing model metadata (`authors`, `tags`) and defined automated testing configurations (`get_test_train_params`).
- **Comprehensive Test Suite**: Included a full Pytest harness for the model (`tests/test_models/test_units_v2.py`) validating parameter exceptions, tensor shapes, gradient flow bounds, and default hyperparameter states.

Thank you!


pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,56 @@
...
+
+class UniTS_pkg_v2(Base_pkg):
+    """
+    UniTS: Unified Time Series Model.
+    Reference: https://arxiv.org/abs/2403.00131
+    """
+
+    _tags = {
+        "info:name": "UniTS",
+        "authors": ["Muhammad-Rebaal"],
```
phoeenniixx: I think we should also credit the actual author of the model? The one who implemented it originally?

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,56 @@
...
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Define varied configurations for auto-testing."""
+        return [
+            {
```
phoeenniixx, line 57: I think we should also have the default fixture, ie, `{}`

pytorch_forecasting/models/__init__.py
```diff
@@ -20,6 +20,7 @@
 )
 from pytorch_forecasting.models.tide import TiDEModel
 from pytorch_forecasting.models.timexer import TimeXer
+from pytorch_forecasting.models.units import UniTS_pkg_v2
```
phoeenniixx: I think you should add the model (and not the `pkg`) here?

pytorch_forecasting/models/units/_units_v2.py
```diff
@@ -0,0 +1,203 @@
...
+    _TransformerBlock,
+)
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class UniTS(TslibBaseModel):
+    """
+    UniTS: Unified Time Series Model.
+
+    Patch-based transformer for multivariate time series forecasting.
```
phoeenniixx, line 28: I think it would be good if you could also add the references (like GH link and the paper) here?

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: yeah sure, I'll add him as well

Muhammad-Rebaal:
Hi @phoeenniixx ,

Kindly review the PR, I've updated the PR with the requested changes.

Thank You.

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: Thanks, for letting me know. I've added the default fixture.

pytorch_forecasting/layers/_units/_units.py
```diff
@@ -0,0 +1,127 @@
...
+Core Neural Network Layers for the UniTS architecture.
+"""
+
+import math
+
+import torch
+import torch.nn as nn
+
+
+class _PatchEmbedding(nn.Module):
```
phoeenniixx: would it make sense to add it to [`_embeddings`](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/layers/_embeddings)?

pytorch_forecasting/layers/_units/_units.py
```diff
@@ -0,0 +1,127 @@
...
+        B, num_patches, C, P = patches.shape
+        patches = patches.permute(0, 2, 1, 3).contiguous().view(B * C, num_patches, P)
+        emb = self.drop(self.projection(patches))
+        emb = emb.view(B, C, num_patches, self.projection.out_features)
+
+        # Channel independence: average across channels as per UniTS logic
+        return emb.mean(dim=1)
+
+
+class _PositionalEncoding(nn.Module):
```
phoeenniixx: it could go to `layers/_encoders`?

pytorch_forecasting/layers/_units/_units.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        pe[:, 0::2] = torch.sin(position * div_term[: pe[:, 0::2].size(1)])
+        pe[:, 1::2] = torch.cos(position * div_term[: pe[:, 1::2].size(1)])
+        self.register_buffer("pe", pe.unsqueeze(0))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.drop(x + self.pe[:, : x.size(1), :])
+
+
+class _TransformerBlock(nn.Module):
```
phoeenniixx: should we add it to `layer/_transforms` or something?

pytorch_forecasting/layers/_units/_units.py
Muhammad-Rebaal: Yes I think it would make complete sense if we'd place it there in a file called **_patch_embedding.py**, as the `_embeddings directory` is already the designated home for embedding abstractions **(like _data_embedding.py and _en_embedding.py)**.

pytorch_forecasting/layers/_units/_units.py
Muhammad-Rebaal: I think creating a new `_transforms` folder wouldn't be necessary here because it would be confusing for new contributors as **transform** and **transformers** are 2 different terminologies. Furthermore, we already have an established `layers/_blocks/ directory`. A transformer block is standard neural network block logic, so keeping it grouped with other blocks in `layers/_blocks/_transformer_block.py` works perfectly entirely to our existing layout.

pytorch_forecasting/layers/_units/_units.py
Muhammad-Rebaal: Actually its just a misinterpretation, it is an embedding not an encoder. We already have a **PositionalEmbedding** class in `_embeddings/_positional_embedding.py` doing the exact same math. Instead of duplicating that logic in an `_encoder.py` file, I created a **_PositionalEmbedding** child class inside the existing **_positional_embedding.py** file. It inherits the fixed sinusoidal buffer from the parent and adds the specific dropout and additive forward logic required for **UniTS**.

Muhammad-Rebaal:
Hi @phoeenniixx ,
Kindly review the PR

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,74 @@
...
+    Reference: https://arxiv.org/abs/2403.00131
+    """
+
+    _tags = {
+        "info:name": "UniTS",
+        "authors": ["Muhammad-Rebaal", "sohamukute"],
+        "capability:exogenous": True,
+        "capability:multivariate": True,
+        "capability:pred_int": False,
+        "capability:flexible_history_length": False,
```
phoeenniixx: I think some tags are missing here. PLease look at the [extension-templates](https://github.com/sktime/pytorch-forecasting/tree/main/extension_templates/v2)

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,74 @@
...
+    def get_cls(cls):
+        from pytorch_forecasting.models.units._units_v2 import UniTS
+
+        return UniTS
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
```
phoeenniixx: Is this a `tslib` model? Why not use `EncoderDecoderDataModule`?

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    def get_cls(cls):
+        from pytorch_forecasting.models.units._units_v2 import UniTS
+
+        return UniTS
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
```
phoeenniixx: the path for the data modules have changed. Please use `pytorch_forecasting.data.data_module` instead

phoeenniixx (changes requested):
Thanks!
I have added some suggestions based on a shallow review

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: Updated

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: Yes, this is a **tslib** model. However, we are not using the `EncoderDecoderDataModule` because **UniTS** is an encoder-only architecture. Since it doesn't use a decoder, wrapping it in that specific module is unnecessary.

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,80 @@
...
+        return UniTS
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data.data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def get_base_test_params(cls):
```
phoeenniixx: Why are we using this method instead of `get_test_train_params`?
I feel like an AI hallucination? It was correct earlier

tests/test_models/test_units_v2.py
```diff
@@ -0,0 +1,225 @@
...
+        instance = UniTS_pkg_v2()
+        assert instance.get_tag("info:name") == "UniTS"
+        assert instance.get_tag("capability:multivariate") is True
+        assert instance.get_tag("capability:exogenous") is True
+        assert instance.get_tag("capability:pred_int") is False
+
+    def test_get_datamodule_cls(self):
+        assert UniTS_pkg_v2.get_datamodule_cls() is not None
+
+    def test_get_base_test_params(self):
```
phoeenniixx: why are we checking if it returns the correct params? I dont understand this test, the params are itself used as fixtures for testing?

Can you please explain these tests. 
Thanks

tests/test_models/test_units_v2.py
```diff
@@ -0,0 +1,225 @@
...
+            "scale": torch.ones(B, 1, C),
+            "center": torch.zeros(B, 1, C),
+        },
+    }
+    if Cc > 0:
+        batch["history_cont"] = torch.randn(B, T, Cc)
+    return batch
+
+
+class TestUniTSPkg:
```
phoeenniixx: I think `pkg` layer IS for testing, why are we testing it separately like this?

phoeenniixx (changes requested):
Thanks, I think the tests in `test_units_v2` is a bit of hallucination or something of an misunderstanding. 

Please look at other model tests (like [`test_timexer_v2`](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_models/test_timexer_v2.py) ) for more info

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    UniTS: Unified Time Series Model.
+    Reference: https://arxiv.org/abs/2403.00131
+    Github: https://github.com/mims-harvard/UniTS
+    """
+
+    _tags = {
+        "info:name": "UniTS",
+        "info:pred_type": ["point"],
+        "info:y_type": ["numeric"],
```
phoeenniixx: You might want to remove this for now. As v2 doesnt support all the losses rn, we had to create a separate collection mechanisms.  these tags (if both are present) moves the pkg to the v1 test framework.

Muhammad-Rebaal: Thanks for the review, I've updated that.

Muhammad-Rebaal:
Hi @phoeenniixx, 

Kindly review this and let me know if we're done with this one so I'd add the `api-reference` for this one.

pytorch_forecasting/layers/_units/__init__.py
phoeenniixx, line 1: what is the use of this folder?

tests/test_models/test_units_v2.py
```diff
@@ -0,0 +1,175 @@
...
+        dropout=0.1,
+        patch_len=8,
+        stride=4,
+        logging_metrics=[SMAPE()],
+        optimizer="adam",
+        metadata=basic_metadata,
+    )
+
+
+def test_basic_model_initialization(model, basic_metadata):
```
phoeenniixx: i think test framework is enough for these tests?

phoeenniixx (changes requested):
I think it is almost ready. Just please de-duplicate the tests that can already being tested using the test framework

tests/test_models/test_units_v2.py
Muhammad-Rebaal: Yes, the extra tests are redundant I've removed that.

pytorch_forecasting/layers/_units/__init__.py
Muhammad-Rebaal: The model file already imports directly from the canonical locations. I'll remove the `_units` layer folder entirely.

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,83 @@
...
+    UniTS: Unified Time Series Model.
+    Reference: https://arxiv.org/abs/2403.00131
+    Github: https://github.com/mims-harvard/UniTS
+    """
+
+    _tags = {
+        "info:name": "UniTS",
+        "info:compute": 4,
+        "authors": ["Muhammad-Rebaal", "gasvn", "sohamukute"],
+        "python_dependencies": ["torch"],
```
phoeenniixx: why do we need this? isnt `torch` already a core dep here?

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,83 @@
...
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            ``MyClass(**params)`` or ``MyClass(**params[i])`` creates a valid test
+            instance. ``create_test_instance`` uses the first (or only) dictionary in
+            ``params``.
+        """
+        params = [
+            {},
+            {
+                "patch_len": 8,
+                "stride": 4,
```
phoeenniixx, line 59: please add some loss functions as well here - Is this model only compatible with point prediction losses, or can it also handle quantile and distribution losses?

tests/test_models/test_units_v2.py
phoeenniixx, line 1: why do we have this file? I dont understand, it has just have a few fixtures and no clear tests are written, 
Can you explain what this file is exactly doing?

phoeenniixx (changes requested):
Please remove the test file, or add some tests there - currently it feels like an AI hallucination

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: Actually I implemented this seeing the model extension v2. So I think I its an essential tag to be added torch as dependencies, not knowing the `core-dep` can't be added. Although I removed that.

pytorch_forecasting/models/units/_units_pkg_v2.py
Muhammad-Rebaal: In the paper the model it is mentioned only about point prediction losses but we can extend it so I extend it to both quantile and distribution losses.

tests/test_models/test_units_v2.py
Muhammad-Rebaal: Actually the model was initially implemented by the use of `TslibBaseModel` which later converted to `EncoderDecoder` so I removed those tests manually and new tests yet to be added which I added in my latest commit.

Muhammad-Rebaal:
###  Changes made in the recent commit:
- removed the extra tag.
- Added support of losses (quantile, distribution)
- Write tests for the model.

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,96 @@
...
+    @classmethod
+    def get_datamodule_cls(cls):
+        from pytorch_forecasting.data.data_module import (
+            EncoderDecoderTimeSeriesDataModule,
+        )
+
+        return EncoderDecoderTimeSeriesDataModule
+
+    @classmethod
+    def get_base_test_params(cls):
```
phoeenniixx: As I mentioned earlier, `get_base_test_params` is not going to work for now for v2
use `get_test_train_params` instead.
for that you need to remove `pred_type`, `y_type` tags

phoeenniixx: looping over all the losses is not a good idea rn for v2

Muhammad-Rebaal: Aren't these tags  `pred_type`, `y_type` be removed  from here : https://github.com/sktime/pytorch-forecasting/blob/main/extension_templates/v2/model_simple/model_pkg.py if these are part of `get_base_test_params` ?

> looping over all the losses is not a good idea rn for v2

Can you suggest me the preferred ones ? so I'd add those

phoeenniixx: > Aren't these tags `pred_type`, `y_type` be removed from here : https://github.com/sktime/pytorch-forecasting/blob/main/extension_templates/v2/model_simple/model_pkg.py if these are part of `get_base_test_params` ?

I have reservations about this :)
Yes, we can remove this, but we will have to add it again, once the work on updating the test framework is done - which is the next work item on my list. As, we dont have much contributors around at this time, we can manage to ignore this. wdyt?

Muhammad-Rebaal: Hmm, right

Muhammad-Rebaal:
Hi @phoeenniixx ,
I've made the following changes as mentioned:

- Remove the tags and renamed `get_base_test_params` to `get_test_train_params `

Also I add another change:

- Removed the `Distribution loss` because I didn't find any implementation of the `Distribution loss` in base_model_v2 ? If there's present kindly let me know I'd implement it...

pytorch_forecasting/models/units/_units_pkg_v2.py
```diff
@@ -0,0 +1,89 @@
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            ``MyClass(**params)`` or ``MyClass(**params[i])`` creates a valid test
+            instance. ``create_test_instance`` uses the first (or only) dictionary in
+            ``params``.
+        """
+        from pytorch_forecasting.metrics import NormalDistributionLoss, QuantileLoss
```
phoeenniixx: I am not sure if `DistributionLoss` works well with v2 rn, does it?

Muhammad-Rebaal: No I didn't find any implementation as well that's why I asked you above. That's why in my last commit I remove the code implementation as well but forget to remove this dead import sorry for confusion

Muhammad-Rebaal: Is there any open issue regarding the support of `Distribution loss` in **v2 models** ?

pytorch_forecasting/models/units/_units_v2.py
```diff
@@ -0,0 +1,224 @@
...
+        self.n_quantiles = None
+        self.n_dist_args = None
+
+        if isinstance(self.loss, QuantileLoss):
+            self.n_quantiles = len(self.loss.quantiles)
+
+        elif isinstance(self.loss, DistributionLoss):
+            self.n_dist_args = len(self.loss.distribution_arguments)
+        output_dim = self.prediction_length * self.target_dim
+
```
phoeenniixx: i think we should remove this part as well?

phoeenniixx (changes requested):
I think it is almost ready - just please make sure to remove the part related to `DistributionLoss` completely for now. And add `todo` there!

Muhammad-Rebaal:
Hi @phoeenniixx,

I've removed the distribution loss implementation and add the todos.

Thanks for the review.

pytorch_forecasting/layers/_embeddings/_positional_embedding.py
```diff
@@ -42,3 +42,29 @@ def __init__(self, d_model, max_len=5000):
 
     def forward(self, x):
         return self.pe[:, : x.size(1)]
+
+
+class _PositionalEmbedding(PositionalEmbedding):
```
phoeenniixx: I have a doubt:
Why do we need this class? All i can see is it adds dropout and add `x` to the return of `PositionalEmbedding.forward()`

Why dont update the existing class? Atmost the difference is additon of `x`, but are you sure that is important enough to be a new class?

Muhammad-Rebaal: Yes, I agree with your concern up here. I've made the `dropout` additive `x` as an optional  Thx

pytorch_forecasting/layers/_blocks/_transformer_block.py
```diff
@@ -0,0 +1,45 @@
+"""
+Pre-norm Transformer Encoder Block for PTF.
+"""
+
+import torch
+import torch.nn as nn
+
+
+class _TransformerBlock(nn.Module):
```
phoeenniixx, line 9: why keep it private?

pytorch_forecasting/layers/_embeddings/_patch_embedding.py
```diff
@@ -10,6 +10,55 @@
 )
 
 
+class UNITS_PatchEmbedding(nn.Module):
```
phoeenniixx, line 13: A doubt:
Can the already implemented one very different from this? and cant handle what this one does? I feel this one is just a linear layer so the already implemented one is a better option?

pytorch_forecasting/layers/_blocks/_transformer_block.py
Muhammad-Rebaal: oh, I just forget the convention we're following, I felt its an internal class no user-end facing thing that's why.

tests/test_models/test_units_v2.py
phoeenniixx, line 1: Can you remove some tests here - the ones that check if "the code that you implemented is actually implemented".
Like checking if there are correct metadata keys? why do we need that here? this is just a model implementation?

There are a lot of tests which i feel may not be necessary. We should just add the tests that check if the model behaves how it should be.  The `forward` pass, and prediciton etc are already tested by the test framework, you dont need to add those. The new tests should just be checking something that the test framework doesnt check (like how the model behaves in a specific setting - which is very specific to this model etc)

pytorch_forecasting/layers/_blocks/_transformer_block.py
phoeenniixx: We should keep it public as it may be used by some other models as well and by the users themselves

pytorch_forecasting/layers/_embeddings/_patch_embedding.py
Muhammad-Rebaal: Actually the patch_embedding for another model gets merge earlier, that's why let me check if it follow that embedding follow the research paper and get back to you.

tests/test_models/test_units_v2.py
Muhammad-Rebaal: ok, I'll check and remove those ones

phoeenniixx (approved):
Thanks!

PR #2169, opened 2026-03-10
[ENH] Extend v2 BaseModel with AdamW optimizer and additional LR schedulers

### LINKED ISSUE #2132: [ENH] Extend v2 BaseModel with AdamW optimizer and additional LR schedulers
While looking into the v2 BaseModel for foundation model work (#1959, #2051), I noticed the optimizer and scheduler support is pretty limited compared to v1.

Right now, [_get_optimizer()](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model_v2.py#L304-L310) only handles "adam" and "sgd" as strings. 
v1 already supports "adamw", "ranger", generic torch.optim class name lookup, and even callable optimizers ([v1 code here](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model.py#L1401-L1484)).

Same story for [_get_scheduler()](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model_v2.py#L335-L344) only ReduceLROnPlateau and StepLR are supported. 
For fine-tuning pretrained models or even just modern training loops, you'd typically want cosine annealing or warm restarts, and v2 doesn't have those yet.

Bring  _get_optimizer() and _get_scheduler() closer to v1 parity + add a couple of schedulers that are commonly needed:

- "adamw" in  _get_optimizer(), basically a backport from v1.
- Generic torch.optim class name lookup (v1 does hasattr(torch.optim, name) which is nice)
- Callable optimizer support (pass a class or function instead of a string)
- "cosine_annealing" → CosineAnnealingLR
- "cosine_annealing_warm_restarts" → CosineAnnealingWarmRestarts
- Tests for the new paths

All changes scoped to  _base_model_v2.py, just the two methods.

One option would be to handle this at the FoundationModelBase level (i.e., only add these for foundation models in a future subclass). But since v1 already supports these and they're useful for any v2 model, it makes more sense to put them in the shared BaseModel so all v2 models benefit.

Another option is to just support callable optimizers and let users pass whatever they want  but having string shortcuts for common ones ("adamw", "cosine_annealing") is much better for the typical user experience and keeps things consistent with v1.

**Additional context**
This came up while working on the foundation model integration (#1959). Models like Chronos, TTM, and Moirai all use AdamW with cosine/warmup schedules for fine-tuning, and the current v2 base doesn't support any of that. It's also tracked as part of the v2 work items (#1974).

Following the guidance of @agobbifbk in #2051 to bring HF Trainer capabilities into the v2 Lightning module, this brings the optimizer/scheduler layer to parity with what foundation models expect.

I'd be happy to start working on this and submit a PR for the same.
cc @phoeenniixx @fkiraly

StrikerEureka34:
#### Reference Issues/PRs

Closes #2132

Picked up the work discussed in #2132. The v2 `BaseModel` only supported "adam" and "sgd" via an if-else chain, with no path for AdamW or any cosine annealing schedulers. Additions of the same were finalized following a registry based approach.

So that is what this does. Two class-level dicts replace the if-else logic:

- `_OPTIMIZER_REGISTRY`: "adam", "adamw", "adagrad", "sgd"
- `_SCHEDULER_REGISTRY`: "reduce_lr_on_plateau", "step_lr", "cosine_annealing", "cosine_annealing_warm_restarts"

If a string is not in the registry, `_get_optimizer()` still checks `hasattr(torch.optim, name)` as a fallback, which is the same behaviour as the old v1 base model. Passing a callable factory or a pre-instantiated `Optimizer` object also still works, nothing was removed.

The `BaseModel.__init__` and `TslibBaseModel.__init__` docstrings are updated to list all supported strings.

Files changed:
- pytorch_forecasting/models/base/_base_model_v2.py
- pytorch_forecasting/models/base/_tslib_base_model_v2.py
- pytorch_forecasting/models/dlinear/_dlinear_pkg_v2.py
- pytorch_forecasting/models/timexer/_timexer_pkg_v2.py
- pytorch_forecasting/models/temporal_fusion_transformer/_tft_pkg_v2.py
- tests/test_models/test_base_model_v2.py (new)
&nbsp;

- Branch ordering in `_get_optimizer()`: callable is checked first, then string (registry then `torch.optim` fallback), then pre-instantiated `Optimizer`. Happy to swap the order if there is a preference.
- Whether the registry shape matches what was agreed in #2132. 
- The ReduceLROnPlateau case: it still injects monitor: "val_loss" into the Lightning config dict, same as before. Just confirming that is still the expected behaviour.
&nbsp;

Yes, `tests/test_models/test_base_model_v2.py` is a new unit test file with 7 test functions, 11 cases total when parametrized:

- `torch.optim` fallback path (e.g. "RMSprop", which is not in the registry)
- Callable optimizer factory and pre-instantiated `Optimizer` passthrough
- Invalid optimizer input (bad string and bad type) merged into one parametrized test
- Parametrized checks for all 4 scheduler strings
- `ValueError` on bad scheduler string 
- `ReduceLROnPlateau` config includes monitor: "val_loss"

Integration coverage via pkg fixtures: 
- DLinear, TimeXer and TFT each got two new get_test_train_params() entries covering adamw + cosine_annealing and adagrad through actual Lightning training loops.

&nbsp;

#### Any other comments?

Since the registries are class-level attributes, subclasses can extend them by overriding `_OPTIMIZER_REGISTRY` or `_SCHEDULER_REGISTRY` without touching `configure_optimizers`. That felt like a nice side effect worth keeping.
&nbsp;

#### PR checklist

- [x] Added/modified tests


StrikerEureka34:
I have merged the branch to the most recent update, pleases let me know if this implementation hold, thanks! @phoeenniixx @PranavBhatP

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -30,18 +30,36 @@ class BaseModel(LightningModule):
...
+        Supported values: "reduce_lr_on_plateau", "step_lr",
+        "cosine_annealing", "cosine_annealing_warm_restarts".
     lr_scheduler_params : Optional[Dict], optional
         Parameters for the learning rate scheduler.
     """
 
+    _OPTIMIZER_REGISTRY = {
+        "adam": torch.optim.Adam,
+        "adamw": torch.optim.AdamW,
+        "sgd": torch.optim.SGD,
```
phoeenniixx, line 52: I think we could also use optimizers like `Adagrad`? what do you think?

phoeenniixx (changes requested):
Thanks! This looks really good!
Could you also try adding some fixtures to the current test framework?
(Making changes to `pkg` classes' test fixtures of v2 models and trying different schedulers and optimizers there?)
Just to make sure this change doesnt break anything?

pytorch_forecasting/models/base/_base_model_v2.py
StrikerEureka34: That's a great idea, I'll do the needful.

StrikerEureka34:
> Thanks! This looks really good! Could you also try adding some fixtures to the current test framework? (Making changes to `pkg` classes' test fixtures of v2 models and trying different schedulers and optimizers there?) Just to make sure this change doesnt break anything?

Okay, I've written the test changes as requested and then finally tracked my whole test execution plan with claude too (so that I don't miss any pkg test files).
Rest of the work and test files are same as before the review was requested.
Please let me know if I missed anything and if this implementation works, thanks!

StrikerEureka34:
I've updated the branch, description of PR and added the pkg tests as requested, please let me know if this implementation holds, thanks! @phoeenniixx

tests/test_models/test_base_model_v2.py
phoeenniixx, line 1: Do we really need this file? I mean it just tests the optimizer registery right? And they can be tested using the params from the `pkg` classes?

StrikerEureka34: Thanks for the review! @phoeenniixx 
I thought about this carefully and I think the unit test file is still needed even with the pkg fixtures in place.
The pkg fixtures and unit tests are actually testing different things. 
 - The pkg fixtures run full training loops but only exercise 2 optimizer strings and 1 scheduler, and never assert the actual optimizer or scheduler types.

 - The unit tests cover paths pkg fixtures cannot reach: ValueError paths for invalid inputs, the callable factory branch, the pre-instantiated optimizer passthrough, the torch.optim fallback, and 3 schedulers (reduce_lr_on_plateau, step_lr, cosine_annealing_warm_restarts) not used in any pkg fixture. 
 
I think removing this file would leave the error handling, edge cases, and most of the scheduler registry completely untested.

In essence, the pkg fixtures test that training works, the unit test file tests that the API contract is correct, including the parts that would never surface during normal training. They're not duplicating each other.

tests/test_models/test_base_model_v2.py
```diff
@@ -0,0 +1,162 @@
...
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UserWarning)
+        return DummyModel(**kwargs)
+
+
+# --- optimizer tests ---
+
+
+@pytest.mark.parametrize("name", ["adam", "adamw", "adagrad", "sgd"])
+def test_optimizer_registry_strings(name):
```
phoeenniixx: I think this test is redundant no? if the optimizers are not of expected cls, they would fail in `pkg` class itself?

tests/test_models/test_base_model_v2.py
```diff
@@ -0,0 +1,162 @@
...
+    assert isinstance(cfg["optimizer"], torch.optim.AdamW)
+
+
+def test_optimizer_invalid_string():
+    model = _make_model(optimizer="not_a_real_optimizer")
+    with pytest.raises(ValueError, match="not supported"):
+        model.configure_optimizers()
+
+
+def test_optimizer_invalid_type():
```
phoeenniixx: I think the `test_optimizer_invalid_string` itself can test this one?

tests/test_models/test_base_model_v2.py
```diff
@@ -0,0 +1,162 @@
...
+    model = _make_model(
+        optimizer="adam",
+        optimizer_params={"lr": 1e-3},
+        lr_scheduler="bogus_scheduler",
+    )
+    with pytest.raises(ValueError, match="not supported"):
+        model.configure_optimizers()
+
+
+def test_reduce_lr_on_plateau_returns_monitor():
```
phoeenniixx, line 106: I think this and the one after that has kinda already been tested in `pkg` class

phoeenniixx (changes requested):
Thanks! I think you are right, it would be good to have a unit test file. I would just suggest to remove some redundant tests and also add the docstrings to the tests - what exactly the test is about?
I think it is then ready to merge!

tests/test_models/test_base_model_v2.py
StrikerEureka34: Makes sense, I tried running a test suite before writing this test and found that the pkg tests only verify that training completes, not that the right optimizer class was used. So if "adamw" accidentally resolved to Adam, both have the same interface so training would pass fine. This test is the only thing explicitly checking the registry maps to the correct class.

I can attach an end-to-end script if needed, that I used to verify this silent failing behaviour.

tests/test_models/test_base_model_v2.py
StrikerEureka34: Actually these test two different branches in _get_optimizer with different error messages:

- `test_optimizer_invalid_string` passes a string that's not in the registry, which raises "not supported"
- `test_optimizer_invalid_type` passes an int, which hits a separate else branch raising "must be a string, a callable, or an instance of torch.optim.Optimizer"

So one won't catch the other since they're completely separate code paths.

phoeenniixx: Actually I meant could we merge them? sorry if I was not clear enough

tests/test_models/test_base_model_v2.py
StrikerEureka34: Nice catch, I removed the one after it. 
Kept this one since it checks that `configure_optimizers` returns `{"monitor": "val_loss"}` in the scheduler config, which Lightning requires at runtime for ReduceLROnPlateau. 

I cross checked and found that no pkg fixture currently uses that scheduler so nothing else catches this if it regresses.

tests/test_models/test_base_model_v2.py
phoeenniixx: `AdamW` cant resolve to `Adam` as you have "hardcoded" the logic and have mapped everything correctly :) 
We dont have to check every nitpick I mean - like this mapping? But if you feel this is necessary, I have no issues!

tests/test_models/test_base_model_v2.py
StrikerEureka34: oh, that's a great idea, I'll do that, also could you please check the latest commit, I've retained a few tests and removed the redundant ones, and added my reasoning for the ones I retained, still happy to remove them if the coverage is not needed.

tests/test_models/test_base_model_v2.py
StrikerEureka34: Fair point, actually I got confused between the testing branches while checking the runs, rest assured I'll remove this, thanks for catching this! :)

StrikerEureka34:
Thanks a lot for the feedback, it really helped! 

I've cleaned up the redundant tests, merged the two invalid input ones into a single parametrized test, and added docstrings to everything remaining which finally gets us 7 test functions, 11 cases in total. 
I've also updated the PR description to reflect the current state. 

Please let me know if anything else needs changing, thanks! @phoeenniixx

phoeenniixx (approved):
Thanks a lot!

StrikerEureka34:
I've updated the Branch, thanks! @phoeenniixx

agobbifbk:
Nice job and thx @phoeenniixx for the great review!

Issue #2157, opened 2026-03-08
[ENH] Add TSMixer (or NLinear) model implementation


Sylver-Icy:
Hi! My LightTS PR is almost ready and I was thinking of working on another model next. I found **TSMixer** and **NLinear** as possible additions. Would either of these be good candidates to add to the repo, or is there another model you’d prefer next? @phoeenniixx

PR #2171, opened 2026-03-11
[ENH] Add NLinear model implementation

Sylver-Icy:
Reference Issues/PRs
#2157 

Reference implementation based on the LTSF-Linear paper:
https://github.com/cure-lab/LTSF-Linear
https://github.com/cure-lab/LTSF-Linear/blob/main/models/NLinear.py
⸻
This PR adds an implementation of the NLinear model for the PyTorch Forecasting v2 estimator API.

Main changes:
	•	Added NLinear model implementation using TslibBaseModel
	•	Implemented support for:
	•	point forecasting
	•	quantile forecasting via QuantileLoss
	•	optional per-channel linear layers (individual=True)
	•	Ensured compatibility with TslibDataModule and v2 estimator workflows

Reference paper:
https://arxiv.org/abs/2205.13504

Added a new test module:

tests/test_models/test_nlinear_v2.py

The tests cover:
	•	model initialization
	•	forward pass execution
	•	correct output tensor shapes
	•	behavior for both point forecasting and quantile forecasting
	•	compatibility with the v2 estimator pipeline

This implementation closely follows the structure used by existing v2 models (e.g., DLinear) to maintain consistency with the repository architecture.

Happy to adjust structure or implementation details if there are preferred patterns for new models.

### DIFF
--- a/pytorch_forecasting/models/base/_tslib_base_model_v2.py
+++ b/pytorch_forecasting/models/base/_tslib_base_model_v2.py
@@ -76,17 +76,17 @@ class TslibBaseModel(BaseModel):
         self.unknown_indices = feature_indices.get("unknown", [])
         self.target_indices = feature_indices.get("target", [])
 
-        feature_dims = self.metadata.get("n_features", {})
+        feature_dims = metadata.get("n_features", {})
         self.cont_dim = feature_dims.get("continuous", 0)
         self.cat_dim = feature_dims.get("categorical", 0)
         self.static_cat_dim = feature_dims.get("static_categorical", 0)
         self.static_cont_dim = feature_dims.get("static_continuous", 0)
         self.target_dim = feature_dims.get("target", 1)
 
-        self.feature_names = self.metadata.get("feature_names", {})
+        self.feature_names = metadata.get("feature_names", {})
 
         # feature-mode
-        self.features = self.metadata.get("features", "MS")
+        self.features = metadata.get("features", "MS")
 
     def _init_network(self):
         """
--- a/pytorch_forecasting/models/nlinear/_nlinear_v2.py
+++ b/pytorch_forecasting/models/nlinear/_nlinear_v2.py
@@ -91,10 +91,6 @@ class NLinear(TslibBaseModel):
 
     def _init_network(self):
         """Initialize NLinear network layers for single-target input only."""
-        if not self.metadata:
-            raise ValueError(
-                "NLinear requires `metadata` from a fitted datamodule to initialize."
-            )
         if self.context_length <= 0 or self.prediction_length <= 0:
             raise ValueError(
                 "NLinear requires positive `context_length` and `prediction_length` "
@@ -190,7 +186,7 @@ class NLinear(TslibBaseModel):
         input_data = self._prepare_input_data(x)
         prediction = self._encoder(input_data)
 
-        if "target_scale" in x and hasattr(self, "transform_output"):
+        if "target_scale" in x:
             prediction = self.transform_output(prediction, x["target_scale"])
 
         return {"prediction": prediction}


pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
```
phoeenniixx: Why are you using this method?

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
```diff
@@ -0,0 +1,133 @@
...
+        """Create test dataloaders from trainer_kwargs - following v1 pattern."""
```
phoeenniixx: I think you were adding v2 model? why v1? Please make sure to manually check the output what AI is giving you, if you are using one

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
Sylver-Icy: mb this method shouldn’t be here
I initially scaffolded pkg file by copying the structure from dlinear so the package layout and flow stayed consistent while implementing the mode  _get_test_datamodule_from() came along with that scaffold and I forgot to remove it afterward xd
 I’ll remove it rq ty for spotting it

Sylver-Icy:
TslibBaseModel initializes metadata as
self.metadata = metadata or {}
However, several attributes were still being read from the raw metadata argument rather than self.metadata. When metadata=None was passed, this could lead to an attributeerror when calling .get() on none

This change updates those accesses to consistently use self.metadata, ensuring the normalized metadata object is used throughout the class and preventing potential crashes when metadata is not provided

pytorch_forecasting/models/base/_tslib_base_model_v2.py
```diff
@@ -69,24 +69,24 @@ def __init__(
         self.context_length = self.metadata.get("context_length", 0)
         self.prediction_length = self.metadata.get("prediction_length", 0)
 
-        feature_indices = metadata.get("feature_indices", {})
+        feature_indices = self.metadata.get("feature_indices", {})
         self.cont_indices = feature_indices.get("continuous", [])
```
phoeenniixx, line 73: I think these changes are out of scope for this PR?

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
+
+    @classmethod
+    def get_test_dataset_from(cls, **kwargs):
```
phoeenniixx, line 37: why are you adding this method?

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
```diff
@@ -0,0 +1,85 @@
...
+            "train": TimeSeries(training_data, **ts_kwargs),
+            "predict": TimeSeries(validation_data, **ts_kwargs),
+        }
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Return testing parameter settings for the trainer."""
+        params = [
+            {},
+            dict(datamodule_cfg=dict(context_length=12, prediction_length=3)),
```
phoeenniixx, line 76: I think there should be some fixtures making changes to the model params itself?

phoeenniixx (changes requested):
I  have a doubt: are you sure there are no model specific params (like n_layers or something) here?

phoeenniixx:
Ok, I looked at the implementation, it seems like a basic linear model?

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
phoeenniixx: where is this method coming from?

Sylver-Icy: the default test datasets include covariates and categorical features, while NLinear v2 enforces a strict target-history-only input contract
using the default dataset would therefore trigger the validation errors for unsupported inputs

this helper creates a dataset containing only `time_idx`, group identifiers, nd the target column so that the package tests run with inputs that match the model's constraints

I avoided modifying the shared datamodule and kept the change scoped to the model package instead

pytorch_forecasting/models/base/_tslib_base_model_v2.py
Sylver-Icy: I explained it in above comment 👆🏻
the change does not affect the existing behavior of the codebase. It only replaces direct access to the raw metadata argument with self.metadata, which is already normalized earlier as metadata or {}

I added it because the class was still reading from the raw argument in a few places, which could raise an attributeerror if metadata=None and .get() is called

If you prefer keeping this PR strictly scoped to the NLinear implementation, I can remove the change and open a separate PR for it just lmk

Sylver-Icy:
> Ok, I looked at the implementation, it seems like a basic linear model?

Yes sir the current implementation follows the original paper which is just a single linear projection from
context_length → prediction_length

So there are currently no model-specific hyperparameters like n_layers
or hidden_size to vary in fixtures the only configurable dimensions
are determined by the datamodule (context_length / prediction_length)
and the loss (e.g. quantile count)

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
Sylver-Icy: this follows the original LTSF-NLinear formulation from the paper. It is just a single linear projection from context_length to prediction_length after the normalization step

pytorch_forecasting/models/nlinear/_nlinear_pkg_v2.py
phoeenniixx: Is this method even being used anywhere? I mean how the test framework calls it?

Sylver-Icy: It’s used by the v2 test utilities in test_all_v2/utils.py
Since the tests operate on the package class, this override is picked up through dynamic dispatch and used to construct the dataset

phoeenniixx: Ahh thanks! I forgot about that :)
But if you look at `make_datasets_v2`, you can see categorical vars is already an [empty list](https://github.com/sktime/pytorch-forecasting/blob/edbdeb49f422cacfb7faeed7b85c523c79475693/pytorch_forecasting/tests/_data_scenarios.py#L197). (we still dont have categorical support in v2).
So do we really need this? I can see even the numerical vars and known vars are kept empty. Why can you please explain?

Sylver-Icy: Even though `make_datasets_v2` sets `cat/num/known` to empty lists, the underlying dataframe still contains additional columns

`TimeSeries` derives feature columns directly from the DataFrame (not those config lists), so those extra columns are included in `feature_cols` and then into the model inputs

Since my model enforces a strict target-history-only contract and rejects any continuous/categorical features, the default dataset leads to invalid inputs

This helper selects only ["time_idx", group ids, "target"], ensuring no feature columns are created, and the dataset matches the model constraints.

Hope that clears it

pytorch_forecasting/models/base/_tslib_base_model_v2.py
phoeenniixx: yes please remove this, maybe we could use a new PR for this (make sure noone else has already not done this:))

pytorch_forecasting/models/nlinear/_nlinear_v2.py
```diff
@@ -0,0 +1,220 @@
...
+            Reshaped output tensor.
+        """
+        if self.n_quantiles is not None:
+            if output.size(-1) != 1:
+                raise ValueError(
+                    "Quantile output expects a single target channel, but got "
+                    f"{output.size(-1)}."
+                )
+            batch_size = output.shape[0]
+            output = output.squeeze(-1).reshape(
```
phoeenniixx: do we need this `sequeeze`? I think `reshape` should handle it quite effectively?

pytorch_forecasting/models/nlinear/_nlinear_v2.py
```diff
@@ -0,0 +1,220 @@
...
+
+    NLinear normalizes each series by subtracting the last observed value,
+    applies a linear projection from context length to prediction length,
+    and then restores scale by adding the last observed value back.
+
+    This v2 implementation is intentionally narrow and follows the paper's
+    target-history-only formulation. It supports a single target variable with
+    a fixed context length and does not currently support exogenous,
+    categorical, static, or future-known features.
+
```
phoeenniixx, line 28: Could you also add some refernces here?

pytorch_forecasting/models/nlinear/_nlinear_v2.py
Sylver-Icy: done I removed it :)

pytorch_forecasting/models/nlinear/_nlinear_v2.py
Sylver-Icy: done this too

pytorch_forecasting/models/base/_tslib_base_model_v2.py
Sylver-Icy: I removed the out of scope changes as you said.

pytorch_forecasting/models/nlinear/_nlinear_v2.py
```diff
@@ -0,0 +1,228 @@
...
+        """
+        # Detach to match the original NLinear reference implementation
+        # In this implementation history_target is a leaf tensor from the dataloader
+        # so this detach has no effect on gradient flow but preserves paper fidelity
+        seq_last = x[:, -1:, :].detach()
+        x = x - seq_last
+
+        output = self.linear(x.permute(0, 2, 1)).permute(0, 2, 1)
+
+        output = output + seq_last.expand(-1, output.size(1), -1)
```
phoeenniixx, line 142: I was looking at the official implementation you mentioned in the PR description, based on [this line](https://github.com/cure-lab/LTSF-Linear/blob/0c113668a3b88c4c4ee586b8c5ec3e539c4de5a6/models/NLinear.py#L37):
I have a doubt, why are we expanding the `seq_last`, it didnt happen in the official implementation, if the rest of the `forward` is the absic copy paste, why are we then expanding it here?

PR #2178, opened 2026-03-12
[ENH] add configurable window stride to v2 data module

lordaarush:
#### Reference Issues/PRs

~~Implements the ptf-v2 "conditional sampling" idea mentioned on Discord by @agobbifbk~~
Implements a configurable window stride for the v2 native datamodule.

This PR adds a `window_stride` parameter to `EncoderDecoderTimeSeriesDataModule` in the v2 data pipeline.

Currently, `_create_windows(...)` generates every valid sliding window by advancing one step at a time. This PR makes that behavior configurable by allowing users to keep every `k`-th valid window start instead.

With this change:
- `window_stride=1` preserves the current default behavior
- larger values reduce the density of generated windows, e.g. keeping every 6th or 24th valid start

This adds a configurable window-generation control to the v2 datamodule layer by letting users choose the stride between valid window starts. Similar window-selection controls are common in modern forecasting libraries and are useful for reducing redundant overlapping samples or matching deployment schedules.

Concretely, the PR:
- adds the `window_stride` argument to `EncoderDecoderTimeSeriesDataModule`
- validates that `window_stride >= 1`
- applies `window_stride` inside `_create_windows(...)` when iterating candidate start indices
- updates tests for the native v2 datamodule accordingly

Yes.

I added tests in `tests/test_data/test_data_module.py` to cover:
- the default `window_stride=1` behavior
- reduced window density when `window_stride > 1`
- validation failure when `window_stride < 1`

I also ran:

```bash
pytest tests/test_data/test_data_module.py

### DIFF
--- a/pytorch_forecasting/data/data_module.py
+++ b/pytorch_forecasting/data/data_module.py
@@ -73,6 +73,10 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
 
     randomize_length : Union[None, Tuple[float, float], bool], default=False
         Whether to randomize input sequence length.
+    window_stride : int, default=1
+        Step size between consecutive candidate windows. A value of 1 keeps the
+        current dense sliding-window behavior, while larger values only keep every
+        k-th valid window start.
     batch_size : int, default=32
         Batch size for DataLoader.
     num_workers : int, default=0
@@ -104,6 +108,7 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
         ]
         | None = None,
         randomize_length: None | tuple[float, float] | bool = False,
+        window_stride: int = 1,
         batch_size: int = 32,
         num_workers: int = 0,
         train_val_test_split: tuple = (0.7, 0.15, 0.15),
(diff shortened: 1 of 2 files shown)


agobbifbk:
Hello there, this is not conditional, it is just a parametrized stride (that is something useful, especially in case of large datasets). For me conditionals means that the starting point of a sliding window should be controlled by a couple variable-value (for example the starting hour = 0). In most of the case you can control it with the stride, but if you think about real application, where there are leap days or solar/legal hours, this can not be ensure by the stride. For sure the parametrized stride is a nice-to-have feature (thx!), but not what I had in mind.

lordaarush:
Thanks for the clarification @agobbifbk , that makes sense and clears things up for me. You’re right that this is better described as configurable window stride rather than true conditional sampling, so I’ll update the PR title and description to avoid misleading reviewers.
I’m still interested in the actual conditional sampling idea you described, where the start of a sliding window is controlled by variable-value conditions rather than only by stride. I’d like to do a bit more research on that and work on it separately in another PR.
If this stride-based version is still considered useful as a standalone enhancement, I’m happy to keep this PR scoped that way.

agobbifbk:
ye sure, it is helpful but the name of the PR is misleading, can you change it? I'm not so skillful in git :-)

lordaarush:
Of course, I’ll update the title and description to reflect this.

PR #2179, opened 2026-03-12
[BUG]: Fix TimeXer autograd graph detachment and deepcopy crash

### LINKED ISSUE #2110: [BUG] TimeXer integration test failing due to requires_grad
Describe the bug

The automated CI tests are currently failing on the TimeXer model during the integration testing phase. Specifically, the test TestAllPtForecasters.test_integration[TimeXer-base_params-3-RMSE] crashes during the backward pass with a RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn. This indicates that the loss calculation or the output tensor has become detached from the PyTorch autograd computation graph.

To Reproduce
This occurs during the standard pytest suite run. You can reproduce it by running the specific failing test:

""pytest pytorch_forecasting/tests/test_all_estimators.py::TestAllPtForecasters::test_integration[TimeXer-base_params-3-RMSE]"""

Expected behavior
The TimeXer model should successfully complete the backward pass and the test should pass alongside the other estimator integration tests.

Additional context
I encountered this failure during the CI run for my Pull Request (upgrading the TFT v2 architecture). Since my PR only touched the _tft_v2.py files and the corresponding TFT tests, this appears to be a pre-existing issue within the TimeXer model's forward/loss logic.

Here is the relevant snippet from the CI traceback:
/opt/hostedtoolcache/Python/3.13.12/x64/lib/python3.13/site-packages/torch/autograd/__init__.py:364: in backward
    _engine_run_backward(

t_outputs = (tensor(12909.9453),), args = ((None,), False, False, ())
kwargs = {'accumulate_grad': True, 'allow_unreachable': True}
attach_logging_hooks = False

    def _engine_run_backward(
        t_outputs: Sequence[Union[torch.Tensor, GradientEdge]],
        *args: Any,
        **kwargs: Any,
    ) -> tuple[torch.Tensor, ...]:
        attach_logging_hooks = log.getEffectiveLevel() <= logging.DEBUG
        if attach_logging_hooks:
            unregister_hooks = _register_logging_hooks_on_whole_graph(t_outputs)
        try:
            return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
                t_outputs, *args, **kwargs
            )  # Calls into the C++ engine to run the backward pass

Versions
This occurred on the GitHub Actions CI runners across multiple environments. Here is an approximation based on the CI logs:
OS: Linux (Ubuntu-latest) / macOS / Windows
Python: 3.10 - 3.14 (Failed across the matrix, specific log from 3.13.12)
PyTorch / PyTorch Lightning: Latest CI versions

Siddhazntx:
#### Reference Issues/PRs

Fixes #2110 

This PR resolves the RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn that caused the automated CI to fail during TimeXer integration tests (specifically [TimeXer-base_params-3-RMSE]).

**Root Cause & Fixes**:

1. Autograd Graph Detachment (Shared State Bug): When initializing TimeXer for multivariate forecasting (features="M"), MultiLoss was being initialized using list multiplication ([MAE()] * len(...)). This created multiple pointers to the exact same metric instance. During the forward pass for multiple targets, the internal state was overwritten, breaking the autograd graph and detaching the final loss tensor.
Change: Updated the initialization to use a list comprehension ([MAE() for _ in range(...)]), ensuring every target receives an independent metric instance.

2. Deepcopy Crash on Hyperparameter Save: self.save_hyperparameters() was attempting to deepcopy initialized, complex metric objects and live tensors, which PyTorch natively blocks, causing local test crashes.
Change: Added ignore=["loss", "logging_metrics"] to save_hyperparameters() and ensured it executes before super().__init__() wraps them.

- The updated initialization order in TimeXer.__init__.

- The list comprehension logic for MultiLoss to ensure it aligns with the expected metric tracking structure in PyTorch Lightning.

No new tests were added, as this PR fixes the existing, failing CI integration tests. Validated the changes locally by running the full test suite in headless mode (pytest pytorch_forecasting/tests/test_all_estimators.py -k "TimeXer"), resulting in a clean pass for all 42 variants.

<details>
<summary><b>Click to expand: Local Test Run Output (All 42 TimeXer tests passing)</b></summary>

```text
ptf-dev) PS D:\pytorch-forecasting> $env:MPLBACKEND="Agg"; pytest pytorch_forecasting/tests/test_all_estimators.py -k "TimeXer"
Test session starts (platform: win32, Python 3.10.19, pytest 9.0.2, pytest-sugar 1.1.1)
cachedir: .cache
rootdir: D:\pytorch-forecasting
configfile: pyproject.toml
plugins: cov-7.0.0, dotenv-0.5.2, sugar-1.1.1, xdist-3.8.0
collected 317 items / 275 deselected / 42 selected                                                                                                                                                                

 pytorch_forecasting\tests\test_all_estimators.py::TestAllPtForecasters.test_doctest_examples[TimeXer] ✓                                     2% ▎         
 pytorch_forecasting\tests\test_all_estimators.py::TestAllPtForecasters.test_integration[TimeXer-base_params-0-MAE] ✓                        5% ▌         
...
 pytorch_forecasting\tests\test_all_estimators.py::TestAllPtForecasters.test_integration[TimeXer-base_params-3-RMSE] ✓                      24% ██▍       
...
 pytorch_forecasting\tests\test_all_estimators.py::TestAllPtForecasters.test_integration[TimeXer-base_params-4-QuantileLoss] ✓              98% █████████▊
 pytorch_forecasting\tests\test_all_estimators.py::TestAllPtForecasters.test_pkg_linkage[TimeXer-TimeXer] ✓                                100% ██████████
=================================================================== warnings summary =================================================================== 
pytorch_forecasting/tests/test_all_estimators.py: 48 warnings
  C:\Users\hp\anaconda3\envs\ptf-dev\lib\site-packages\lightning\pytorch\utilities\_pytree.py:21: `isinstance(treespec, LeafSpec)` is deprecated, use `isinstance(treespec, TreeSpec) and treespec.is_leaf()` instead.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

Results (119.35s (0:01:59)):
      42 passed
     275 deselected
```
</details>

#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/timexer/_timexer.py
+++ b/pytorch_forecasting/models/timexer/_timexer.py
@@ -193,7 +193,7 @@ class TimeXer(BaseModelWithCovariates):
         if loss is None:
             output_size = kwargs.get("output_size", 1)
             if isinstance(output_size, (list, tuple)) and len(output_size) > 1:
-                loss = MultiLoss([MAE() for _ in range(len(self.target_positions))])
+                loss = MultiLoss([MAE() for _ in range(len(output_size))])
             else:
                 loss = MAE()
 
@@ -487,11 +487,10 @@ class TimeXer(BaseModelWithCovariates):
             # which is the length of a list of float. In case of MAE, MSE, etc.
             # n_quantiles = 1 and it mimics the behavior of a point prediction.
             # for multi-target forecasting, the output is a list of tensors.
-            if len(target_positions) > 1:
-                prediction = [prediction[..., i, :] for i in target_indices]
-            else:
+            if len(target_positions) == 1:
                 prediction = prediction[..., 0, :]
-
+            else:
+                prediction = [prediction[..., i, :] for i in target_indices]
             prediction = self.transform_output(
                 prediction=prediction, target_scale=x["target_scale"]
             )


pytorch_forecasting/models/timexer/_timexer.py
```diff
@@ -191,14 +191,13 @@ def __init__(
         if logging_metrics is None:
             logging_metrics = nn.ModuleList([SMAPE(), MAE(), RMSE(), MAPE()])
         if loss is None:
-            if features == "M":
-                loss = MultiLoss([MAE()] * len(self.target_positions))
-            else:
-                loss = MAE()
+            loss = MAE()
```
phoeenniixx: why are you removing the `MultiLoss` from here?

pytorch_forecasting/models/timexer/_timexer.py
```diff
@@ -484,10 +483,20 @@ def forward(self, x: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
             # which is the length of a list of float. In case of MAE, MSE, etc.
             # n_quantiles = 1 and it mimics the behavior of a point prediction.
             # for multi-target forecasting, the output is a list of tensors.
-            if len(target_positions) == 1:
-                prediction = prediction[..., 0, :]
+            if len(target_positions) > 1:
+                prediction = torch.stack(
```
phoeenniixx: why are you using `torch.stack`? I think for multi-target, it should be a `list` of tensors?

pytorch_forecasting/models/timexer/_timexer.py
Siddhazntx: Originally took it out because the old way `([MAE()] * len(...))` shared the exact same metric object in memory, which was actually what caused the autograd graph to detach. But you're totally right. I'll add it back, but I'll use a list comprehension `([MAE() for _ in range(...)])` so each target gets its own independent metric. That should fix the bug and keep the graph stable!

pytorch_forecasting/models/timexer/_timexer.py
Siddhazntx: My bad on that one! I was trying to format the 3D tensor output and just defaulted to torch.stack without thinking. A list is definitely the right way to go here. I'll update it!

Siddhazntx:
Hey @phoeenniixx, Just pushed a quick follow-up commit. The last update tripped a unit test because I tried to check `self.target_positions` to initialize MultiLoss before `super().__init__()` actually created it (which threw an AttributeError).

I swapped it to safely check `kwargs.get("output_size") `instead. All the local unit and integration tests are completely green on my end now! Let me know if everything looks good to you.

pytorch_forecasting/models/timexer/_timexer.py
```diff
@@ -484,10 +487,11 @@ def forward(self, x: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
             # n_quantiles = 1 and it mimics the behavior of a point prediction.
             # for multi-target forecasting, the output is a list of tensors.
-            if len(target_positions) == 1:
-                prediction = prediction[..., 0, :]
-            else:
+            if len(target_positions) > 1:
                 prediction = [prediction[..., i, :] for i in target_indices]
+            else:
+                prediction = prediction[..., 0, :]
+
```
phoeenniixx: why are you making these changes? I feel like they are doing the same things as before? ARe you sure we need these changes?

Siddhazntx: Thanks for pointing out. While I was originally hunting down the `RuntimeError` autograd bug, I suspected the graph was breaking because of how the 3D tensors were being sliced and passed to` MultiLoss`. I was experimenting with `torch.stack` and reshaping to try and fix it.

Once I discovered the actual root cause was `save_hyperparameters` deepcopying the metrics, I should have reverted this specific prediction block back to the original upstream implementation, as these shape changes are no longer needed.

I will revert these lines back to the original base code right now so the PR is clean! Sorry for the back-and-forth.

pytorch_forecasting/models/timexer/_timexer.py
```diff
@@ -191,14 +191,17 @@ def __init__(
         if logging_metrics is None:
             logging_metrics = nn.ModuleList([SMAPE(), MAE(), RMSE(), MAPE()])
         if loss is None:
-            if features == "M":
-                loss = MultiLoss([MAE()] * len(self.target_positions))
+            output_size = kwargs.get("output_size", 1)
+            if isinstance(output_size, (list, tuple)) and len(output_size) > 1:
+                loss = MultiLoss([MAE() for _ in range(len(output_size))])
```
phoeenniixx, line 196: This is still changing the condition no? I agree it should be for `output_size>1`, but can you explain `features=="M"` would not work and its better to change the condition?

Siddhazntx: The main reason is that `features == "M"` only tells us that it's a multi-target dataset, but it doesn't tell us how many targets there are.
Also checking `output_size `from` kwargs` does both jobs at once. It safely confirms it's a multi-target dataset while simultaneously giving us the exact loop length we need to build independent MAE objects, completely avoiding the `AttributeError` caused by calling `self.target_positions` too early.

PR #2181, opened 2026-03-13
[BUG] Fix TslibDataModule ignoring target_normalizer and add_target_scales

### LINKED ISSUE #2164: [BUG] TslibDataModule ignores target_normalizer/add_target_scales - Tslib v2 models never receive target_scale
**Describe the bug**

`TslibDataModule` accepts `target_normalizer`, `add_target_scales`, and `scalers` as constructor parameters but never applies them. `_preprocess_data` always returns an empty `target_scale = {}` regardless of configuration, so targets are never normalized and `x["target_scale"]` is never populated in any batch.

`TslibBaseModel.transform_output` - which is called in `DLinear.forward` as `self.transform_output(prediction, x["target_scale"])` to invert normalization on model outputs can therefore never function correctly for any Tslib v2 model. Any user passing `add_target_scales=True` gets silently wrong results with no warning or error.

**To Reproduce**

```python
import pandas as pd
import torch
from pytorch_forecasting.data.timeseries._timeseries_v2 import TimeSeries
from pytorch_forecasting.data._tslib_data_module import TslibDataModule

df = pd.DataFrame({
    "time":  list(range(50)),
    "value": [float(i) for i in range(50)],
    "group": ["a"] * 50,
})
ts = TimeSeries(df, target="value", time="time", group="group")

dm = TslibDataModule(
    ts,
    context_length=10,
    prediction_length=5,
    add_target_scales=True,   # <-- should activate scaling
)
dm.setup()
x, y = next(iter(dm.train_dataloader()))

print("target_scale in x:", "target_scale" in x)
# Actual:   False
# Expected: True

# With add_target_scales=True, targets in the batch should be normalized
# and x["target_scale"] should be a dict with "center" and "scale" tensors
# so that transform_output can invert normalization on model predictions.
```

Root cause: `self._target_normalizer` is stored in `__init__` but `setup()` never calls `.fit()` on it, and `_preprocess_data()` has no fitted state to apply. Additionally, `collate_fn` attempts to `torch.stack` `target_scale` entries assuming they are tensors, which silently fails when they are dicts.

**Expected behavior**

When `add_target_scales=True`:

1. `setup()` fits the configured normalizer (default: `RobustScaler`) on training targets only no leakage into val/test.
2. `_preprocess_data()` applies the fitted normalizer to each series target and populates `target_scale = {"center": tensor, "scale": tensor}` matching the format expected by `TslibBaseModel.transform_output`.
3. `collate_fn` correctly stacks the per-sample `target_scale` dicts into batch-level tensors of shape `(batch_size, 1)`.
4. The default path (`add_target_scales=False`) is completely unchanged.

When `add_target_scales=False` (current default), behaviour is unchanged, `target_scale` is absent from the batch and targets are unscaled.

**Additional context**

- Directly addresses the v2 roadmap item *"Addition of preprocessing steps - scaling, transformations etc in D1/D2 layers"* from #1993.
- The `add_target_scales` flag has been a public API parameter since the D2 layer was introduced but has never had any effect.
- All existing v2 model tests (`test_timexer_v2`, `test_dlinear_v2`) continue to pass with the fix applied since they use the default `add_target_scales=False` path.
- Fix is fully contained in `pytorch_forecasting/data/_tslib_data_module.py`.

**Versions**

<details>

```
System:
    python: 3.11.13 | packaged by Anaconda, Inc. | (main, Jun  5 2025, 13:03:15) [MSC v.1929 64 bit (AMD64)]
executable: E:\Anaconda\envs\vlm_env\python.exe
   machine: Windows-10-10.0.26100-SP0
Python dependencies:
          pip: 25.2
pytorch-forecasting: 1.6.1
        torch: 2.6.0+cu126
    lightning: 2.6.1
        numpy: 2.3.3
        scipy: 1.16.2
       pandas: 2.3.3
      cpflows: None
   matplotlib: 3.10.7
       optuna: None
optuna-integration: None
pytorch_optimizer: None
 scikit-learn: 1.7.2
  scikit-base: 0.13.1
  statsmodels: None
```

</details>

phantom-712:
#### Reference Issues/PRs
 
Fixes #2164
 
`TslibDataModule` accepted `target_normalizer`, `add_target_scales`, and `scalers` as public API parameters but never applied them. `_preprocess_data()` always returned `target_scale = {}` regardless of configuration, meaning targets were never normalized and `x["target_scale"]` was never populated in any batch.
 
This PR wires up the full normalization pipeline:
 
- Adds `_extract_scale_params()` to extract center/scale from fitted sklearn scalers (handles both `RobustScaler` and `StandardScaler`)
- `setup()` now fits the `target_normalizer` on training data only before dataset creation, avoiding val/test data leakage
- `_preprocess_data()` now applies the fitted normalizer and populates `target_scale = {"center": tensor, "scale": tensor}`
- `collate_fn()` now correctly stacks dict-valued `target_scale` across the batch into `(batch_size, 1)` tensors instead of calling `torch.stack` on a dict, which would raise `TypeError`
 
- The fitting logic in `setup()` -  specifically that the normalizer is fit on training indices only and the fitted parameters are cached correctly
- The `collate_fn` change -  `target_scale` is now a dict so `center` and `scale` are stacked separately to produce `(batch_size, 1)` tensors
- The triple guard in `_preprocess_data()` - ensuring the default path (`add_target_scales=False`) is completely unaffected
 
Existing tests (`test_timexer_v2`, `test_dlinear_v2`) pass unchanged, confirming the default path is unaffected. Smoke tests were run manually verifying both the default path (no `target_scale` in batch) and the scaling path (`target_scale` dict with correct `center`/`scale` values present in batch).
 
#### Any other comments?
 
N/A
 
#### PR checklist
 
- [x] The PR title starts with [BUG]
- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/data/_tslib_data_module.py
+++ b/pytorch_forecasting/data/_tslib_data_module.py
@@ -344,6 +344,9 @@ class TslibDataModule(LightningDataModule):
         self.scalers = scalers or {}
         self.shuffle = shuffle
 
+        self._normalizer_fitted = False
+        self._fitted_target_scale = None
+
         self.continuous_indices = []
         self.categorical_indices = []
 
@@ -364,7 +367,38 @@ class TslibDataModule(LightningDataModule):
 
         self._validate_indices()
 
-    def _validate_indices(self):
+    def _extract_scale_params(self, normalizer: Any) -> dict[str, torch.Tensor]:
+        """
+        Extract center and scale parameters from a fitted sklearn-like scaler.
+
+        Returns a dict with keys "center" and "scale" as torch.float32 tensors.
+        """
+
+        center = None
+        if hasattr(normalizer, "center_"):
+            center = normalizer.center_
+        elif hasattr(normalizer, "mean_"):
+            center = normalizer.mean_
+
+        scale = None
+        if hasattr(normalizer, "scale_"):
+            scale = normalizer.scale_
+
+        if center is None or scale is None:
+            raise ValueError(
+                "Cannot extract target scaling params. Expected scaler with "
+                "center_/scale_ (RobustScaler) or mean_/scale_ (StandardScaler)."
+            )
+
+        center_np = np.asarray(center, dtype=np.float32).reshape(-1)
+        scale_np = np.asarray(scale, dtype=np.float32).reshape(-1)
+
+        return {
+            "center": torch.tensor(center_np, dtype=torch.float32),
+            "scale": torch.tensor(scale_np, dtype=torch.float32),
+        }
+
+    def _validate_indices(self) -> None:
         """
         Validate that we have meaningful features for training.
         Raises warnings for missing features or indices.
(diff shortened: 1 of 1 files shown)


StrikerEureka34:
Thanks for fixing this! I’ve been wrestling with [`TslibDataModule`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/data/_tslib_data_module.py) too, so this is super helpful.

Just one flag: loading the entire dataset's targets into memory for `np.concatenate` in [`setup()`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/data/_tslib_data_module.py#L673-L759) will likely cause OOM on large datasets (I experienced the same while running COSS series, have discussed the results in #2051). 

Maybe we could use `partial_fit()` or require a pre-fitted scaler instead?

PR #2182, opened 2026-03-13
[ENH] add TimesNetModel v2 (BaseModel) with package layer

ShreyanshPagariya-13:
Adds `TimesNet` to v2

- **`_timesnet_layers.py`** — correctness of the `TimesBlock` period-padding
  logic and the `FFT_for_Period` amplitude weighting / aggregation step.
- **`_timesnet_v2.py` `forward()`** — the order of operations, though matches the original
  DSIPTS implementation.
- **`cat_cardinalities` validation** — whether the `ValueError` raised on
  mismatch between `len(cat_cardinalities)` and `metadata["encoder_cat"]` is
  the right approach, or whether silent padding / truncation would be preferable.
- **Output shape** — model returns `(B, H, n_targets)`, reviewers should confirm this is correct for the loss functions
  currently supported in v2.

No dedicated test file has been added yet. The implementation has been manually
verified.

#### Any other comments?

- This PR is part of the ongoing pytorch-forecasting v2 / sktime integration
  effort tracked in issue #1736.

#### PR checklist

- [x] The PR title starts with `[ENH]` — adding new model code.
- [ ] Added/modified tests — manual tests pass; formal test file to be added in a follow-up.
- [ ] Used pre-commit hooks when committing to ensure that code is compliant with hooks.
  Install hooks with `pre-commit install`.
  To run hooks independent of commit, execute `pre-commit run --all-files`.


pytorch_forecasting/layers/_timesnet/_timesnet_layers.py
phoeenniixx, line 1: please add this to `layers` module

pytorch_forecasting/models/timesnet/_timesnet_v2.py
```diff
@@ -0,0 +1,276 @@
...
+        """Return the package-layer class for this model."""
+        from pytorch_forecasting.models.timesnet._timesnet_v2_pkg import (
+            TimesNet_pkg_v2,
+        )
+
+        return TimesNet_pkg_v2
+
+    def __init__(
+        self,
+        metadata: dict,
```
phoeenniixx: i think you should move it after the args with default values?

pytorch_forecasting/models/timesnet/_timesnet_v2_pkg.py
```diff
@@ -0,0 +1,96 @@
...
+        list of dict
+            Three parameter sets covering:
+
+            * minimal architecture (fast, default loss MAE)
+            * medium architecture with explicit MAE loss and custom lengths
+            * small architecture with MAPE loss and short sequences
+        """
+        from pytorch_forecasting.metrics import MAE, MAPE
+
+        params = [
```
phoeenniixx, line 49: add the default param as well `{}`

phoeenniixx (changes requested):
Thanks!
just added few comments

pytorch_forecasting/layers/_timesnet/_timesnet_layers.py
ShreyanshPagariya-13: Thanks for the review and suggestions, added the implementation to the `layers` module.

pytorch_forecasting/models/timesnet/_timesnet_v2.py
ShreyanshPagariya-13: yaa following the convention i ordered them as needed.

pytorch_forecasting/models/timesnet/_timesnet_v2_pkg.py
ShreyanshPagariya-13: Added the default parameter in the package configuration.

Issue #2185, opened 2026-03-13
[ENH] Rework N-HiTS v2 aligned with BaseModel standards


echo-xiao:
**Is your feature request related to a problem? Please describe.**
See https://github.com/sktime/mentoring/blob/main/internships/projects_2026.md: pytorch-forecasting-and-dsipts. It requires rework a pytorch-forecasting model.

reference：https://github.com/sktime/pytorch-forecasting/issues/1992

**Describe the solution you'd like**
I propose adding N-HiTS v2, which fully leverages the v2 framework. Key features of this implementation include:
1. Lightweight Integration: Inherits from the new BaseModel (v2), reducing overhead and improving maintainability.
2. V2 Package Standard: Includes NHiTS_v2_pkg_v2 which conforms to the Base_pkg container requirements and updated _tags metadata.
3. Enhanced Regularization: Added support for backcast_loss_ratio, allowing users to balance forecast accuracy with backcast reconstruction loss.
4. Decoupled Architecture: Core logic is encapsulated in NHiTSModule, separating the neural network architecture from the Lightning-based model wrapper.

**Describe alternatives you've considered**
None

**Additional context**
I have already completed a local implementation that includes:
1. _nhits_v2.py and _nhits_pkg_v2.py.
2. Comprehensive unit tests in test_nhits_v2.py.
3. Verified compatibility with the test_all_estimators_v2 automated test suite.
4. Passed all pre-commit hooks (Ruff linting and formatting).

PR #2186, opened 2026-03-13
[ENH] Add N-HiTS v2 implementation with BaseModel

echo-xiao:

This PR introduces the v2 implementation for the N-HiTS model, aligning with the new framework standards:
1. Core Implementation: Added NHiTS_v2 class inheriting from the new lightweight BaseModel (v2).
2. Package Container: Implemented NHiTS_v2_pkg_v2 with updated _tags (including author, compute complexity, and capabilities).
3. Loss Optimization: Supported backcast_loss_ratio to allow weighted regularization between forecast and backcast losses.
4. Decoupling: Used NHiTSModule for core computation to maintain a clean separation between logic and the framework base class.
5. Test Suite: Added parameter configurations in get_test_train_params covering various pooling modes, interpolation settings, and loss functions.

1. The initialization logic in NHiTS_v2 and how it interacts with the v2 BaseModel.
2. The weighted loss calculation in _compute_loss using backcast_loss_ratio.
3. Completeness of metadata tags in _nhits_pkg_v2.py.

Yes, added a comprehensive test file tests/test_models/test_nhits_v2.py which includes:
1. [x] Output shape validation for forward pass.
2. [x] Gradient flow check during training_step.
3. [x] Stability tests for different backcast_loss_ratio values.
4. [x] Architecture variant tests (n_blocks, hidden_size).

#### Any other comments?
Code has been formatted and linted using pre-commit (Ruff).

#### PR checklist

- [x ] Added/modified tests
- [ x] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.


echo-xiao:
Reference Issues: https://github.com/sktime/pytorch-forecasting/issues/2185

pytorch_forecasting/models/nhits/_nhits_pkg_v2.py
```diff
@@ -0,0 +1,95 @@
+"""NHiTS v2 package container."""
+
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class NHiTS_v2_pkg_v2(Base_pkg):
```
phoeenniixx: I think `NHiTS_pkg_v2` should work now? We added a fix recently, pls give it a try?

pytorch_forecasting/models/nhits/_nhits_v2.py
```diff
@@ -0,0 +1,321 @@
...
+        self._n_stacks = n_stacks
+        self._backcast_loss_ratio = backcast_loss_ratio
+
+        self.model = NHiTSModule(
+            context_length=self.context_length,
+            prediction_length=self.prediction_length,
+            output_size=[1],
+            static_size=0,
+            encoder_covariate_size=0,
+            decoder_covariate_size=0,
```
phoeenniixx: Can you pls clarify how (and why) we are hardcoding these values?

pytorch_forecasting/models/nhits/_nhits_v2.py
```diff
@@ -0,0 +1,321 @@
...
+            static_hidden_size=hidden_size,
+            n_blocks=n_blocks,
+            n_layers=n_layers,
+            hidden_size=n_stacks * [2 * [hidden_size]],
```
phoeenniixx: Here the `hidden_size` is different than the `hidden_size` from `__init__`?

pytorch_forecasting/models/nhits/_nhits_v2.py
```diff
@@ -0,0 +1,321 @@
...
+            loss,
+            on_step=(prefix == "train"),
+            on_epoch=True,
+            prog_bar=True,
+            logger=True,
+        )
+        self.log_metrics(out["prediction"], y, prefix=prefix)
+        return loss, out
+
+    def training_step(self, batch, batch_idx):
```
phoeenniixx, line 376: I think the `BaseModel` should be able to handle this, you dont need to make changes here?

phoeenniixx (changes requested):
Thanks a lot! This looks good!
Added some suggestions and doubts :)

pytorch_forecasting/models/nhits/_nhits_v2.py
echo-xiao: The v2 datamodule doesn't support covariates or static features, so these they are always None in the forward pass. But None is invalid, so 0 is the only way to indicate "not used".

pytorch_forecasting/models/nhits/_nhits_v2.py
echo-xiao: Yes, it is differnt. I will use another name instead.

pytorch_forecasting/models/nhits/_nhits_v2.py
echo-xiao: BaseModel use prediction and y to calculate loss, however, NHits requires target_past to calculate backcast loss. So the changes here is needed.

pytorch_forecasting/models/nhits/_nhits_v2.py
phoeenniixx: I think still hardcording values is not good. We already have warnings saying it is not stable to use v2. But once we have a categorical support, we would need  to make changes here as well, while now, the user can simply themselves add these values and the problem is solved

echo-xiao: I see. You are right. I've updated to derive these values from DataModule metadata. When v2 DataModule adds covariate or static feature support, this will automatically reflect the actual datasets configuration.

pytorch_forecasting/models/nhits/_nhits_v2.py
phoeenniixx: If that is the case, then I think we should have `test_step` etc as well?

pytorch_forecasting/models/nhits/_nhits_pkg_v2.py
```diff
@@ -0,0 +1,95 @@
...
+                n_blocks=[1, 1, 1],
+                n_layers=2,
+                hidden_size=256,
+                pooling_mode="max",
+                interpolation_mode="linear",
+            ),
+            dict(
+                n_blocks=[1, 1],
+                hidden_size=128,
+                backcast_loss_ratio=0.1,
```
phoeenniixx, line 75: question: does this model support only point prediction losses?
I think we should add a 1-2 more fixtures where we pass the losses to `loss` as well (you can add the losses to the existing fixtures itself), not just `logging_metrics`.

echo-xiao: - the nhits architecture supports both point losses and quantileloss foor probabilitic/interval forecasts. this v2 implementation currently wires up point prediction only. i'll add quantilelosss later. 
- also, i have added loss=. to more of the test params

echo-xiao: and it supports quantileloss now.

pytorch_forecasting/models/nhits/_nhits_pkg_v2.py
```diff
@@ -0,0 +1,109 @@
...
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class NHiTS_pkg_v2(Base_pkg):
+    """NHiTS v2 package container."""
+
+    _tags = {
+        "info:name": "NHiTS_v2",
+        "info:compute": 2,
+        "authors": ["echo-xiao"],
```
phoeenniixx: please add the authors of v1 model as well if the code has v1 implementation as well

pytorch_forecasting/models/nhits/_nhits_v2.py
```diff
@@ -0,0 +1,425 @@
...
+        NHiTS_pkg_v2 : type
+            Package class associated with this model.
+        """
+        from pytorch_forecasting.models.nhits._nhits_pkg_v2 import NHiTS_pkg_v2
+
+        return NHiTS_pkg_v2
+
+    def __init__(
+        self,
+        *,
```
phoeenniixx: do we really need to enforce "keyword only" here? people do pass keywords anyway, but we dont enforce this usually ourselves

phoeenniixx (changes requested):
Thanks I think it is almost ready!

I think we have redundant tests  in the test file... `forward` and `fit`, `predict` contract is already checked by the test framework. Can we de-duplicate some tests here?

Few smaller non-blocking comments as well (see below)

pytorch_forecasting/models/nhits/_nhits_v2.py
echo-xiao: yes, deleted

pytorch_forecasting/models/nhits/_nhits_v2.py
```diff
@@ -0,0 +1,424 @@
...
+        # a quantile loss needs one output per quantile, point losses need one.
+        if isinstance(loss, QuantileLoss):
+            n_outputs_per_target = len(loss.quantiles)
+        else:
+            n_outputs_per_target = 1
+        output_size = [n_outputs_per_target] * n_targets
+
+        # The backcast is single-valued per target and cannot be scored by a
+        # quantile loss, so mixing it with backcast regularization is disallowed
+        # (mirrors the v1 NHiTS constraint).
```
phoeenniixx, line 153: we can remove this comment

tests/test_models/test_nhits_v2.py
```diff
@@ -0,0 +1,420 @@
...
+        time_series_dataset=ts,
+        max_encoder_length=CONTEXT_LENGTH,
+        max_prediction_length=PREDICTION_LENGTH,
+        batch_size=BATCH_SIZE,
+    )
+    dm.setup("fit")
+    return dm
+
+
+def test_nhits_v2_backcast_loss_ratio_rejects_quantile_loss(sample_datamodule):
```
phoeenniixx, line 69: we dont need this test

tests/test_models/test_nhits_v2.py
```diff
@@ -0,0 +1,420 @@
...
+
+    with pytest.raises(ValueError, match="backcast_loss_ratio"):
+        NHiTS_v2(
+            loss=QuantileLoss(),
+            metadata=metadata,
+            backcast_loss_ratio=0.1,
+        )
+
+
+def test_nhits_v2_default_loss(sample_datamodule):
```
phoeenniixx, line 91: we dont need this test

tests/test_models/test_nhits_v2.py
```diff
@@ -0,0 +1,420 @@
...
+    if batch_x_2d_mask["encoder_mask"].dim() == 3:
+        batch_x_2d_mask["encoder_mask"] = batch_x_2d_mask["encoder_mask"].squeeze(-1)
+
+    with torch.no_grad():
+        out = model(batch_x_2d_mask)
+
+    assert out["prediction"].shape[1] == PREDICTION_LENGTH
+
+
+def test_nhits_v2_forward_with_3d_mask(sample_datamodule):
```
phoeenniixx, line 129: we can merge this and 2d mask test? Although do we need it? i mean this is implemented and there is little to 0 prob that this will cause issue as you have already covered this in `forward` quite explicitly?

phoeenniixx (changes requested):
Very Nice! I think it is ready!
Please resolve the conflicts and plz review the tests as well - a lot of them just test "if this thing is implemented or not" - like checking if the default loss is `MASE`. We dont need such tests as this is for sure that will be passed and there is almost no possiblity that there can be a mistake here as you made the default `MASE` yourself.

We should make sure instead that the complete implementation works or not. The tests that compares v1 and v2 implementations are the best eg that makes sure that the implementation is correct and completely close to the v1. We should add similar tests - just checking if the feat we implemented work as intended.

echo-xiao:
> Very Nice! I think it is ready! Please resolve the conflicts and plz review the tests as well - a lot of them just test "if this thing is implemented or not" - like checking if the default loss is `MASE`. We dont need such tests as this is for sure that will be passed and there is almost no possiblity that there can be a mistake here as you made the default `MASE` yourself.
> 
> We should make sure instead that the complete implementation works or not. The tests that compares v1 and v2 implementations are the best eg that makes sure that the implementation is correct and completely close to the v1. We should add similar tests - just checking if the feat we implemented work as intended.

agreed. i have already removegd the three that only asserted a feature exists. 
- the default MASE loss
- the backcast_loss_ratio guard
- the two mask tests that existed to cover the dim()==3 branch

current tests:
| Test | What it catches |
|---|---|
| `test_backcast_is_reconstruction_not_residual` | the wrapper passing the raw residual through |
| `test_backcast_weighting_matches_v1_formula` | a wrong backcast convention or weight |
| `test_backcast_loss_ratio_zero_is_forecast_only` | the override leaking into `ratio=0` |
| `test_train_val_test_steps_share_weighted_loss` | one step falling back to the `BaseModel` loss |
| `test_combined_loss_gradient_is_weighted_sum` | a detached backcast term |
| `test_masked_positions_reconstruct_the_input_exactly` | the mask not reaching the residual |
| `test_2d_and_3d_encoder_mask_are_equivalent` | a missing squeeze, which broadcasts at the wrong rank silently |
| `test_nhits_v1_v2_wrapper_outputs_match` `[point, quantile]` | any divergence between the two wrappers |
| `test_nhits_v1_v2_weighted_loss_matches` `[mae, mase]` | a wrong MASE scaling series, invisible under MAE |
| `test_nhits_v1_v2_train_loss_comparable` | a systematic regression in pipeline or loss wiring |

Issue #2187, opened 2026-03-14
[BUG] get_model_cls() called but never defined in _BasePtForecaster_Common


Quant-Code-Hacker:
[BUG] get_model_cls() called but never defined in _BasePtForecaster_Common

**Describe the bug**

`_BasePtForecaster_Common` calls `cls.get_model_cls()` in three places, but this method is never defined anywhere in the class or its hierarchy. The correct method which is defined on line 15 of the same file is `get_cls()`.

Affected file: `pytorch_forecasting/models/base/_base_object.py`

The three call sites:
- Line 56 — `create_test_instance()`: `return cls.get_model_cls()(**params)`
- Line 96 — `create_test_instances_and_names()`: `objs += [cls.get_model_cls()(**params)]`
- Line 24 — `name()` fallback branch: `name = cls.get_model_cls().__name__`

Why no error is raised as:
- Line 24: all current v2 packages set the `info:name` tag, so the `if name is None` branch is never reached.
- Lines 56 and 96: these methods are never called in the current v2 test suite, so the bug is silently dormant.

Any new pkg class that either (a) omits the `info:name` tag, or (b) calls `create_test_instance()` / `create_test_instances_and_names()`, will immediately raise `AttributeError`.

**To Reproduce**

```python
from pytorch_forecasting.models.dlinear._dlinear_pkg_v2 import DLinear_pkg_v2

# Bug 1 : create_test_instance() line 56
DLinear_pkg_v2.create_test_instance()
# AttributeError: type object 'DLinear_pkg_v2' has no attribute 'get_model_cls'

# Bug 2 : create_test_instances_and_names() line 96
DLinear_pkg_v2.create_test_instances_and_names()
# AttributeError: type object 'DLinear_pkg_v2' has no attribute 'get_model_cls'
```

**Expected behavior**

`get_model_cls()` on lines 24, 56, and 96 should be replaced with `get_cls()`, which is already defined in `_BasePtForecaster_Common` (line 15). After replacing the AttributeError will no longer be raised.

Fix:
- Line 24: `name = cls.get_cls().__name__`
- Line 56: `return cls.get_cls()(**params)`
- Line 96: `objs += [cls.get_cls()(**params)]`

Happy to open a PR with the fix — it is a three-line change.

**Versions**
<details>

System:
    python: 3.13.5 (tags/v3.13.5:6cb20a2, Jun 11 2025, 16:15:46) [MSC v.1943 64 bit (AMD64)]
[...]

PR #2188, opened 2026-03-14
[BUG] fix: replace undefined get_model_cls() with get_cls()

saschabuehrle:
## Bug
[#2187](https://github.com/sktime/pytorch-forecasting/issues/2187) — `get_model_cls()` is called but never defined in `_BasePtForecaster_Common`.

## Fix
Replaced all three calls to the undefined `get_model_cls()` with `get_cls()`, which is already defined in the same class (line 15):

- Line 24: `name = cls.get_model_cls().__name__` → `name = cls.get_cls().__name__`
- Line 56: `return cls.get_model_cls()(**params)` → `return cls.get_cls()(**params)`
- Line 96: `objs += [cls.get_model_cls()(**params)]` → `objs += [cls.get_cls()(**params)]`

## Testing
Before this fix:
```python
from pytorch_forecasting.models.dlinear._dlinear_pkg_v2 import DLinear_pkg_v2
DLinear_pkg_v2.create_test_instance()
# AttributeError: type object 'DLinear_pkg_v2' has no attribute 'get_model_cls'
```

After this fix, `create_test_instance()`, `create_test_instances_and_names()`, and the `name()` fallback branch will work correctly.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/pytorch_forecasting/models/base/_base_object.py
+++ b/pytorch_forecasting/models/base/_base_object.py
@@ -21,7 +21,7 @@ class _BasePtForecaster_Common(_BaseObject):
         """Get model name."""
         name = cls.get_class_tags().get("info:name", None)
         if name is None:
-            name = cls.get_model_cls().__name__
+            name = cls.get_cls().__name__
         return name
 
     @classmethod
@@ -53,7 +53,7 @@ class _BasePtForecaster_Common(_BaseObject):
                 "get_test_params should either return a dict or list of dict."
             )
 
-        return cls.get_model_cls()(**params)
+        return cls.get_cls()(**params)
 
     @classmethod
     def create_test_instances_and_names(cls, parameter_set="default"):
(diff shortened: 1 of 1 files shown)


phoeenniixx:
Thanks a lot! nice catch!
I think we forgot to make changes here, when we updated other parts of the codebase.
FYI @fkiraly @PranavBhatP, I am surprised why it was not caught by the CI?

phoeenniixx:
Or is it calling an internal method from `skbase`?

saschabuehrle:
Thanks for checking this path in detail.

It is not calling an internal skbase method here — this code path uses `_BasePtForecaster_Common`, where `get_model_cls()` is undefined and raises an `AttributeError` in the test helpers. Replacing it with `get_cls()` matches the current public helper used elsewhere and fixes the failing test-instance constructors.

Greetings, saschabuehrle

pawamoy:
Sorry to interrupt, I wanted to let you know that this PR is most probably entirely AI-generated, including replies to your comments. Just see the batch of PRs they opened recently to form your own opinion.

PR #2190, opened 2026-03-14
[BUG] Remove outdated strict naming convention test in test_all_estim…

### LINKED ISSUE #2183: [BUG] The old test for naming convention is still present in `test_all_estimators_v2`
**Describe the bug**

The old test to check naming convention is still present in `test_all_estimators_v2`. See [here](https://github.com/sktime/pytorch-forecasting/blob/1a1927903fc36a4155627eb92e4b064d77cbdb1d/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py#L119-L125)

It was meant to be removed and updated by the new "relaxed" naming convention test. See #2080. 

**Suggested Solution**
Remove this block 
https://github.com/sktime/pytorch-forecasting/blob/1a1927903fc36a4155627eb92e4b064d77cbdb1d/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py#L119-L125

from the tests

**Additional context**

See #2080 #2065

AyushDineshRathi:
Closes #2183

The strict naming convention test using `object_pkg.name()` was still present in `test_all_estimators_v2`.

This check enforced that the package name must exactly match the estimator class name. However, the naming convention tests were relaxed in #2080 to allow multiple valid naming patterns.

Because of this update, the old assertion is redundant and may incorrectly enforce the previous strict convention.

This PR removes the obsolete assertion block.

### DIFF
--- a/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py
+++ b/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py
@@ -115,14 +115,6 @@ class TestAllPtForecastersV2(EstimatorPackageConfig, EstimatorFixtureGenerator):
 
     def test_pkg_linkage(self, object_pkg, object_class):
         """Test that the package is linked correctly."""
-        # check name method
-        msg = (
-            f"Package {object_pkg}.name() does not match class "
-            f"name {object_class.__name__}. "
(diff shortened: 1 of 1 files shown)


phoeenniixx (approved):
thanks!

PR #2203, opened 2026-03-17
[BUG] Fix logging_metrics not registered as submodules in v2 BaseModel (#2197)

### LINKED ISSUE #2197: [BUG] v2 BaseModel stores logging_metrics as plain list instead of nn.ModuleList (breaks GPU training)
Describe the bug

The v2 BaseModel stores logging_metrics as a plain Python list instead of wrapping it in nn.ModuleList. This means PyTorch doesn't register the metrics as submodules, so they don't move to GPU when the model does. On GPU training, this causes a device mismatch crash when `log_metrics()` runs.

v1 does this correctly in [_base_model.py](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model.py#L554) in nn.ModuleList. But v2 at [_base_model_v2.py](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model_v2.py#L56)  just assigns the raw list:

```python
# v2 (broken)
self.logging_metrics = logging_metrics if logging_metrics is not None else []

# v1 (correct)
self.logging_metrics = nn.ModuleList([...])
```

The metrics themselves (SMAPE, MAE, MAPE, etc.) all inherit from MultiHorizonMetric, which registers internal state buffers via [add_state()](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/metrics/base_metrics/_base_metrics.py#L829-L838) . These buffers stay on CPU when the model moves to GPU, because the metrics aren't registered submodules.

During a training/validation step, `log_metrics()` calls `metric(y_hat, y)`, which triggers `update()` -> `_update_losses_and_lengths()`, hitting:

```python
self.losses = self.losses + losses  # CPU buffer + GPU tensor -> RuntimeError
```

This affects all 5 existing v2 models (DLinear, TiDE, TimeXer, SAMformer, TFT) since they all inherit from BaseModel.

&ensp;
To Reproduce

```python
import torch
from pytorch_forecasting.metrics import MAE, SMAPE

from pytorch_forecasting.models.dlinear import DLinear

# Create a model with logging_metrics on GPU
model = DLinear(
    loss=MAE(),
    logging_metrics=[SMAPE(), MAE()],
    metadata={...},  # appropriate metadata
)
model = model.to("cuda")

# Check: logging_metrics did NOT move to GPU
for m in model.logging_metrics:
    for name, buf in m._buffers.items():
        print(f"{name}: {buf.device}")  # prints "cpu" -- should be "cuda"

# This will crash during training when log_metrics() is called
# because y_hat (cuda) and metric state buffers (cpu) are on different devices
```

The existing tests don't catch this because:

- The v2 integration suite (test_all_estimators_v2.py) only tests `fit()` and `predict()` at the pkg level, never calling `training_step()` directly with logging_metrics active
- Unit tests that pass ogging_metrics ([test_dlinear_v2.py](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_models/test_dlinear_v2.py#L118), [test_timexer_v2.py](https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_models/test_timexer_v2.py#L208) only run `forward()` under `torch.no_grad()`, never `training_step()`
- CI runs on CPU, so the device mismatch never triggers

Expected behavior

logging_metrics should be stored as nn.ModuleList so that metrics are properly registered as submodules, moved to the correct device with the model, and included in `state_dict()` for checkpointing. This is how v1 handles it.

&ensp;
Suggested fix

In [_base_model_v2.py](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model_v2.py#L56):

```python
# before
self.logging_metrics = logging_metrics if logging_metrics is not None else []

# after
self.logging_metrics = nn.ModuleList(logging_metrics) if logging_metrics else nn.ModuleList()
```

And add a test that actually exercises `log_metrics()` during a training step with `ogging_metrics`passed, to prevent regression.

&ensp;

Additional context

Came across this while studying the v2 base layer for foundation model integration work #1959. 
FMs will train on GPU and need metric logging during fine-tuning, so this is directly in the path. 
Related to #1700 (logging redesign for v2) and #1754 (v1 device propagation for loss metrics), though neither covers this specific issue.

Gitanaskhan26:
#### Reference Issues/PRs

Fixes #2197.

`BaseModel.__init__()` in `_base_model_v2.py` stored `logging_metrics` as a plain Python list instead of `nn.ModuleList`. PyTorch only tracks submodules assigned as `nn.Module`, `nn.ModuleList`, or `nn.ModuleDict` attributes. Because metrics were in a plain list, they were not registered as submodules, so when `model.to("cuda")` was called, metric internal state buffers (`losses`, `lengths` registered via `add_state()`) stayed on CPU. This caused a `RuntimeError: Expected all tensors to be on the same device` during `training_step` / `validation_step` / `test_step`.

All 5 existing v2 models are affected (DLinear, TimeXer, SAMformer, TFT, TiDE).

The fix wraps `logging_metrics` in `nn.ModuleList`, matching the v1 implementation at `_base_model.py:553-558`.

- The one-line production fix in `_base_model_v2.py`
- The three regression tests, in particular whether iterating `metric._defaults` to reach torchmetrics state tensors is acceptable or if there is a cleaner public API

Yes, three tests appended to `tests/test_models/test_dlinear_v2.py`:

- `test_logging_metrics_is_module_list` — asserts the container type and that metrics appear in `model.modules()`
- `test_empty_logging_metrics_is_module_list` — asserts empty case also uses `nn.ModuleList`
- `test_logging_metrics_device_propagation` — moves model to `torch.device("meta")` and checks that metric state tensors follow. This catches the bug on CPU-only CI since `.to("meta")` is a real device move unlike `.to("cpu")` which is a no-op when already on CPU.

#### Any other comments?

Tests reuse the existing `sample_dataset` fixture in `test_dlinear_v2.py`.

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. → [BUG]
- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/base/_base_model_v2.py
+++ b/pytorch_forecasting/models/base/_base_model_v2.py
@@ -53,7 +53,9 @@ class BaseModel(LightningModule):
     ):
         super().__init__()
         self.loss = loss
-        self.logging_metrics = logging_metrics if logging_metrics is not None else []
+        self.logging_metrics = (
+            nn.ModuleList(logging_metrics) if logging_metrics else nn.ModuleList()
+        )
         self.optimizer = optimizer
         self.optimizer_params = optimizer_params if optimizer_params is not None else {}
         self.lr_scheduler = lr_scheduler
--- a/tests/test_models/test_dlinear_v2.py
+++ b/tests/test_models/test_dlinear_v2.py
@@ -166,3 +166,56 @@ def test_univariate_forecast():
     assert "prediction" in output
     assert output["prediction"].shape[0] == dm.batch_size
     assert output["prediction"].shape[1] == metadata["prediction_length"]
+
+
+def test_logging_metrics_is_module_list(sample_dataset):
+    """Test that logging_metrics are stored as nn.ModuleList (#2197)."""
+    dm = sample_dataset["data_module"]
+    metadata = dm.metadata
+
+    metrics = [SMAPE(), MAE()]
+    model = DLinear(
+        loss=MAE(),
+        moving_avg=5,
+        individual=False,
+        logging_metrics=metrics,
+        metadata=metadata,
+    )
+
+    assert isinstance(model.logging_metrics, nn.ModuleList)
+
+    submodule_list = list(model.modules())
+    for m in metrics:
+        assert m in submodule_list
+
+
+def test_empty_logging_metrics_is_module_list(sample_dataset):
+    """Test that empty logging_metrics is also nn.ModuleList (#2197)."""
+    dm = sample_dataset["data_module"]
+    metadata = dm.metadata
+
+    model = DLinear(loss=MAE(), moving_avg=5, individual=False, metadata=metadata)
+
+    assert isinstance(model.logging_metrics, nn.ModuleList)
+    assert len(model.logging_metrics) == 0
+
+
+def test_logging_metrics_device_propagation(sample_dataset):
+    """Test that metric state tensors follow model device moves (#2197)."""
+    dm = sample_dataset["data_module"]
+    metadata = dm.metadata
+
+    model = DLinear(
+        loss=MAE(),
+        moving_avg=5,
+        individual=False,
+        logging_metrics=[SMAPE(), MAE()],
+        metadata=metadata,
+    )
+
+    model.to("meta")
+    for metric in model.logging_metrics:
+        for state_name in metric._defaults:
+            val = getattr(metric, state_name)
+            if isinstance(val, torch.Tensor):
+                assert val.device.type == "meta"


StrikerEureka34:
I request you to remove/close this PR as the scope and nature of the reference issue has not been ascertained yet and I am already working on it.
Appreciate the help though, thanks!

Gitanaskhan26:
Hi @StrikerEureka34, thanks for reporting the issue. The reproduction script was very helpful in tracking this down. I didn't see an open PR or assignment on the issue, so I went ahead with a fix. I'm happy to let the maintainers decide how they'd like to proceed. If there's anything you'd like me to adjust in this implementation, just let me know!

StrikerEureka34:
> I didn't see an open PR or assignment on the issue, so I went ahead with a fix. I'm happy to let the maintainers decide how they'd like to proceed. If there's anything you'd like me to adjust in this implementation, just let me know!

I get your point, but the issue was still being investigated and it's premature to raise a PR.
And also before raising a PR please ask if the author of the issue was working on it or not, which in this case I was.

Gitanaskhan26:
>  please ask if the author of the issue was working on it or not, which in this case I was.

Understood, and I apologize if it felt like I was stepping on your toes! Since the issue was unassigned and didn't have a comment claiming a WIP, I just jumped in when I saw the fix. I will definitely make sure to ask beforehand on future issues.

I'll leave this open for the maintainers to evaluate whenever they have bandwidth. If your investigation led to a different architectural approach, I’m more than happy to collaborate on this or adjust my implementation!

StrikerEureka34:
> Understood, and I apologize if it felt like I was stepping on your toes! Since the issue was unassigned and didn't have a comment claiming a WIP, I just jumped in when I saw the fix. I will definitely make sure to ask beforehand on future issues.

No worries, my only concern was that it leads to duplication of efforts and wastes a lot of time.

> I’m more than happy to collaborate on this

I'll raise a draft PR soon if this bug checks out, looking forward for your thoughts there, thanks!

fnhirwa:
fixed in https://github.com/sktime/pytorch-forecasting/commit/81d1554bc6c8fae1e1cbc6c26be77e225c62a764

PR #2205, opened 2026-03-17
[BUG] Fix logging_metrics device mismatch in BaseModelV2

### LINKED ISSUE #2197: [BUG] v2 BaseModel stores logging_metrics as plain list instead of nn.ModuleList (breaks GPU training)

StrikerEureka34:
#### Reference Issues/PRs

Fixes #2197

Wraps logging_metrics in nn.ModuleList in `BaseModelV2.__init__` so metrics are registered as submodules and move with the model on `.to()` calls.

&ensp;

-The two regression tests in test_base_model_v2.py,  
-Whether iterating `metric._defaults` to reach torchmetrics state tensors is acceptable, or if there's a cleaner public API to assert device propagation without a GPU?

Yes, `tests/test_models/test_base_model_v2.py` (new file):

- `test_logging_metrics_is_module_list`  this asserts the container type directly
- `test_logging_metrics_device_propagation` this moves the model to `torch.device("meta")` and checks that metric state tensors follow. This catches the bug on CPU-only CI since `.to("meta")` is a real device move unlike `.to("cpu")` which is a no-op when already on CPU.

&ensp;

#### Any other comments?

Tests use DLinear with a minimal metadata dict, no dataloader or trainer needed.

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. → [BUG]
- [x] Added/modified tests


fnhirwa (dismissed):
I believe this will work on GPU as well,
any comment @phoeenniixx ?

phoeenniixx:
> I believe this will work on GPU as well, any comment @phoeenniixx ?

Yes, this should work!
Thanks for catching this!!
Although I have a doubt: would it be better to directly add some fixtures to `pkg` classes rather than a standalone test? As there it would be easy to test with more models rather than just one?

StrikerEureka34:
>Although I have a doubt: would it be better to directly add some fixtures to `pkg` classes rather than a standalone test? As there it would be easy to test with more models rather than just one?

Thanks for the feedback!
Makes sense in general, but the fix lives entirely in BaseModelV2.__init__ and all v2 models just pass logging_metrics straight to super().__init__() without touching it again. 
I've checked DLinear, TimeXer, TFT, TiDE v2, and all follow the same pattern.  So testing with one model is genuinely sufficient; there's no model-specific divergence possible.

Still happy to wire up a parameterized conftest fixture across all v2 models if you'd prefer that.

phoeenniixx:
Actually what I mean is do we really need a new test file? Its quite a straightforward test, no? It would be easier to add just some fixtures in few models?
What do you think?

StrikerEureka34:
> Actually what I mean is do we really need a new test file? Its quite a straightforward test, no? It would be easier to add just some fixtures in few models? What do you think?

Fair point, happy to drop the new file. Would move both tests into test_dlinear_v2.py, that already has a logging_metrics test so it's a natural home, Does that work?

phoeenniixx (approved):
Thanks! LGTM

Hi @fnhirwa, do you have any comments?

StrikerEureka34:
I've updated the branch after the recent commit.

Issue #2184, opened 2026-03-13
[ENH] Architectural Proposal: Data Pipeline Subclassing for Foundation Models (TTM Case Study)


StrikerEureka34:
**Related:** [#1959](https://github.com/sktime/pytorch-forecasting/issues/1959) [#2051](https://github.com/sktime/pytorch-forecasting/issues/2051)

### Is your feature request related to a problem? Please describe.

I've been looking into what it would take to wire IBM's [TinyTimeMixer (TTM)](https://github.com/ibm-granite/granite-tsfm/blob/main/tsfm_public/models/tinytimemixer/modeling_tinytimemixer.py) into the v2 pipeline. As part of that, I did a fairly thorough sweep, of `_tslib_data_module.py` and `_base_model_v2.py` against TTM's actual `forward()` signature. Here's what I found.
&ensp;

#### Architecture Sweep Findings

**Finding 1: Exogenous features work differently in TTM, and target may get double-counted**

This is the big one. TTM has **no `exogenous_values` parameter** in its [forward()](https://github.com/ibm-granite/granite-tsfm/blob/main/tsfm_public/models/tinytimemixer/modeling_tinytimemixer.py). Instead, exogenous features go into `past_values` as additional channels. The config uses [prediction_channel_indices](https://github.com/ibm-granite/granite-tsfm/blob/main/tsfm_public/models/tinytimemixer/modeling_tinytimemixer.py) and [exogenous_channel_indices](https://github.com/ibm-granite/granite-tsfm/blob/main/tsfm_public/models/tinytimemixer/modeling_tinytimemixer.py) to tell the model which channels are which.

This maps directly to the channel-ordering requirements from #2051 [(thread here)](https://github.com/sktime/pytorch-forecasting/issues/2051#issuecomment-4053508328) @agobbifbk's DummyDataset example stacks `[target, known_future_covs, past_only_covs]` into a single tensor with strict ordering:
1. Targets first
2. Past-only covariates next
3. Known-future covariates last and their order must match the order in `future_values`

The `_TTMDataset.__getitem__` I'm proposing enforces exactly this ordering.

Two sub-problems:
[...]

Issue #1959, opened 2025-08-28
[ENH] add foundational models into `PTF` for v2.


PranavBhatP:
**Is your feature request related to a problem? Please describe.**
There has been a clear demand for integrating foundational models into `pytorch-forecasting` especially for v2. These foundational models are fundamentally zero shot (and some cases few shot) and their weights are loaded from huggingface. These can be implemented as a child class extension of the v2 `BaseModel`. This specific class would handled everything from loading the weights, configuring the model and generating prediction. The final prediction API contract would be dictated by the new design for predict in v2.

Some promising candidates for integrating into the the library.

1) IBMs TinyTimeMixer
2) Amazon's Chronos.

This issue can serve as a discussion thread for future steps taken in this direction.

Nidhicodes:
This makes a lot of sense directionally, especially with the growing interest in zero-shot / few-shot forecasting models.

A question before jumping into implementation: would it make sense to first agree on how foundation models fit into the v2 abstractions, especially around the predict() contract?

A few design points that might be worth aligning on early:
-Base class vs parallel hierarchy: should these models inherit from BaseModel directly, or should there be a lightweight FoundationModelBase that bypasses training-specific assumptions?
-Predict semantics: for zero-shot models, would predict() operate directly on raw time series (e.g. pandas / tensors), or still require a
-TimeSeriesDataSet: Some foundation models don’t map cleanly onto encoder/decoder concepts.
-Training & fine-tuning: should v2 treat these as inference-only by default, with optional fine-tuning hooks, or explicitly disable fit()?
-Dependency boundaries: would loading weights from Hugging Face be an optional extra (to avoid pulling transformers into the core), with model-specific adapters?

For a first step, it might be useful to pick one pilot model (e.g. Chronos or TinyTimeMixer) and use it to validate the API shape before generalizing. That could help avoid locking in abstractions too early.

Happy to help sketch an interface proposal or compare integration requirements across candidate models if that’s useful.

ShreyasN707:
Hi @PranavBhatP, @phoeenniixx,

I’m interested in contributing to the first batch of foundation model integrations for PTF v2.

Before starting, I wanted to clarify a few design points:

- Should foundation models inherit directly from BaseModel, or would a lightweight FoundationModelBase make more sense given their zero-shot/inference-first nature?

- For zero-shot models like Chronos, should predict() still require a TimeSeriesDataSet, or support raw series input?

[...]

PR #2206, opened 2026-03-17
[WIP][ENH] TTM data pipeline: _dataset_class hook + TTMDataModule

StrikerEureka34:
#### Reference Issues/PRs
WIP towards #2184.  Related: #1959 #2051

WIP draft for #2184, that is integrating IBM's TinyTimeMixer (TTM) into pytorch-forecasting's v2 pipeline. 
TTM is a pretrained foundation model for zero-shot and fine-tuned time series forecasting, and it expects a specific input format that our current v2 pipeline doesn't produce.

This draft/WIP PR covers the data layer only. No model weights, no `tsfm_public` dependency yet, as the main aim of it was to just getting the pipeline to speak TTM's language.

&ensp;

#### Issues I ran into

`TslibDataModule.setup()` had four hardcoded `_TslibDataset(...)` calls. 
So there was no clean way to swap in a different dataset class without duplicating the entire method.

&ensp;

#### Changes made

Added `_dataset_class = _TslibDataset` to TslibDataModule and replaced those four calls with `self._dataset_class(...)`. 
Which gives a zero behavioral change for existing code.

On top of that hook:

- _TTMDataset  now overrides `__getitem__` to produce TTM-format batches:
  past_values, past_observed_mask, future_values, prediction_channel_indices
- TTMDataModule now sets _dataset_class = _TTMDataset, derives channel indices
  from metadata, and provides a `collate_fn` for the mixed Tensor/non-Tensor batch

So the channel ordering follows TTM's spec:
`[targets | past-only covariates | known-future covariates]`

&ensp;

#### Does it work?

Yes, with a `MockTTM` standing in for the real model:

```
test_ttm_data_module_uses_ttm_dataset  PASSED
test_ttm_dataset_captures_stage        PASSED
test_channel_indices_derived           PASSED
test_batch_shapes                      PASSED
test_no_future_values_in_predict       PASSED
test_mock_ttm_forward                  PASSED
```

One pre-existing failure in test_multivariate_target ('list' object has no attribute 'shape') which was reproducible on `main`, so unrelated to this PR and the changes made by it, worth a separate issue if not already tracked?

&ensp;

#### What's left

- **Caching**: `_preprocess_data` runs per `__getitem__`  so it needs a per-series cache before this is usable at scale (I mentioned it as a TODO in the code and using a short term fix for current scope)
- **Model wrapper**: thin wrapper around `tsfm_public.models.tinytimemixer`, as an optional dependency.
- **Training glue**: loss, forward, predict steps wired through `TslibBaseModel`

&ensp;

The _dataset_class hook is intentionally generic, with TTMDataModule as the first user, but the same pattern should work for Chronos, Moirai, and others with their own `__getitem__` overrides.
More discussion and testing is needed for that I believe, but initial findings seem promising.

Looking for early feedback on this hook design and TTM channel format assumptions before going further, thanks!
@phoeenniixx @PranavBhatP @agobbifbk

### DIFF
--- a/pytorch_forecasting/data/_fm_data_module.py
+++ b/pytorch_forecasting/data/_fm_data_module.py
@@ -2,7 +2,7 @@
 Foundation model data modules for pytorch-forecasting.
 
 Provides ``TTMDataModule`` for IBM's TinyTimeMixer (TTM) and the backing
-``_TTMDataset``.  The pattern — subclassing ``TslibDataModule`` and setting
+``_TTMDataset``.  The pattern subclassing ``TslibDataModule`` and setting
 ``_dataset_class`` — is designed to be reused for other foundation models
 (Chronos, Moirai, etc.) with their own ``__getitem__`` overrides.
 """
@@ -76,7 +76,7 @@ class _TTMDataset(_TslibDataset):
         history_target = processed_data["target"][history_indices]
         future_target = processed_data["target"][future_indices]
 
-        # Continuous features: (total_len, n_cont) — targets already excluded by base
+        # Continuous features: (total_len, n_cont) targets already excluded by base
         cont_features = processed_data["features"]["continuous"]
         cont_names = self.data_module.metadata["feature_names"]["continuous"]
         cont_name_to_idx = {name: i for i, name in enumerate(cont_names)}
@@ -123,11 +123,11 @@ class _TTMDataset(_TslibDataset):
         }
 
         # Stage-aware: omit future_values during predict; include for all other stages
-        # (None is treated identically to "fit" — include future_values)
+        # (None is treated identically to "fit" and include future_values)
         if self._stage != "predict":
             x["future_values"] = future_known
 
-        # y — squeeze to (prediction_length,) for single target
+        # y, squeeze to (prediction_length,) for single target
         if self.data_module.n_targets > 1:
             y = [future_target[:, i] for i in range(self.data_module.n_targets)]
         else:
@@ -186,7 +186,7 @@ class TTMDataModule(TslibDataModule):
         n_targets = self.metadata["n_features"]["target"]
 
         # Ordering must match _TTMDataset.__getitem__ concatenation:
-        #   targets → past-only (cont_names order) → known-future (cont_names order)
+        #   targets -> past-only (cont_names order) -> known-future (cont_names order)
         past_only_cont = [n for n in cont_names if n not in known_names]
         known_future_cont = [n for n in cont_names if n in known_names]
 
--- a/pytorch_forecasting/data/_fm_data_module.py
+++ b/pytorch_forecasting/data/_fm_data_module.py
@@ -1,0 +1,238 @@
+"""
+Foundation model data modules for pytorch-forecasting.
+
+Provides ``TTMDataModule`` for IBM's TinyTimeMixer (TTM) and the backing
+``_TTMDataset``.  The pattern — subclassing ``TslibDataModule`` and setting
+``_dataset_class`` — is designed to be reused for other foundation models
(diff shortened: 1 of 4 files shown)


agobbifbk:
Are you sure that we need to dedicate a custom data layer for a single model? Now you are aware on what the model expects as input. I would like to see something like I posted [here](https://github.com/sktime/pytorch-forecasting/issues/2051#issuecomment-4053508328) where you try to figure out if the pieces (in terms of tensors, past, future, categorical, numerical) we have in the current data module can be rearranged in order to build the batch that TTM expects. If we can move the logic to the forward loop we can avoid custom data layer. If you see an high incompatibility I would like to see why you think it is not possible. THX

StrikerEureka34:
> Are you sure that we need to dedicate a custom data layer for a single model? Now you are aware on what the model expects as input. I would like to see something like I posted [here](https://github.com/sktime/pytorch-forecasting/issues/2051#issuecomment-4053508328) where you try to figure out if the pieces (in terms of tensors, past, future, categorical, numerical) we have in the current data module can be rearranged in order to build the batch that TTM expects. If we can move the logic to the forward loop we can avoid custom data layer. If you see an high incompatibility I would like to see why you think it is not possible. THX

Yeah this makes a lot of sense. Two wins if we do it this way:
1. No need for a dedicated data layer for TTM as _TslibDataset already carries everything.
2. cats can go straight through nn.Embedding inside forward, which honestly wouldn't have worked in a worker-side dataset anyway (embeddings live on the main process). 

That second point I'd missed completely, so thanks for pushing back, I hope this is what you meant?

Untitled Diagram (1)

I've added a gist here with the test suite to check if the rearrangement works and if cats can go in as embeddings, while keeping the pointers and diagram you mentioned in #2051 in mind: [here](https://gist.github.com/StrikerEureka34/160d985257d1147b4c95013dd4158080). 

What the results mean:-
- `Part 1 max|forward - dataset|=0.00e+00`: 
Forward-loop is bit-identical to `_TTMDataset`. So nothing will be lost if we move it out.
- `Part 2 (20, 11) vs (20, 3) (diff = 8 cat-emb channels)`: 
The cats are carried by forward and silently dropped by `_TTMDataset`. 
min_emb_grad_abs_sum=80 confirms embeddings are learnable only when they sit on the model.
I had missed this completely, and its all the more reason why we don't need the separate data module here.
- `R1/R2/R3 all return True for both cont and cat..`

Some of the implications that I can see for this pr after the analysis are:-
- `_TTMDataset` + custom collate go.
- TTMDataModule goes too. TslibDataModule is used as-is; its metadata["feature_names"] already has continuous / categorical / known.
- `TTMForecaster.__init__` derives past_only_* / known_future_* splits from metadata once. forward uses them for rearrangement + nn.Embedding.
- `_dataset_class` hook stays available for Chronos/Moirai if they ever need dataset-side work just nobody overrides it for TTM, should we keep this for flexibility or remove it entirely? 

So what this would finally look like is:-

Untitled Diagram drawio (10)



Please let me know if this provides a clearer picture, or there are more options that I might've missed? 
Thanks! @agobbifbk @phoeenniixx

agobbifbk:
Much better now, so you idea is that we can build the batch to be compliant with the TTM model, but there are some work to do:
1- ensure the order of the targets
2- have a clear idea if, as for chronos2, the future covariates needs to be also past covariates, in case this is mandatory and the user did not provide the correct metadata raise a warning and fix the past, or raise an error
3- if 2 implies that if you want to use the future covariates they should be also in the past, be sure that the order is correct

The good news is that the reordering part can be done BY DEFAULT since this has not implication in the other models (or at least it should). The subsetting (know should be a subset of the past) can be checked after model initialization using the metadata, isn't it?

If you have all the pieces you can start implementing TTM or raise an issue for doing it!
THX so much!

PR #2208, opened 2026-03-17
[BUG] remove erroneous .detach() in TiDE multivariate prediction path

haoyu-haoyu:
## Summary

In the TiDE model's multivariate target code path (line 333), prediction tensors are processed with:

```python
prediction = [i.clone().detach().requires_grad_(True) for i in prediction]
```

This is incorrect because:

1. `.detach()` disconnects the tensor from the computational graph
2. `.requires_grad_(True)` on a detached tensor creates a **new leaf** — it does NOT reconnect to the encoder/decoder graph
3. Gradients cannot flow back through the prediction during backpropagation
4. The model effectively **cannot learn** for multivariate targets

### Fix

```diff
- prediction = [i.clone().detach().requires_grad_(True) for i in prediction]
+ prediction = [i.clone().requires_grad_(True) for i in prediction]
```

`.clone()` creates a copy (preserving gradient connection), and `.requires_grad_(True)` ensures gradient tracking — without the graph-breaking `.detach()`.

## Test plan

- [ ] TiDE model training converges for multivariate targets
- [ ] `pytest tests/ -k tide` passes
- [ ] Gradient flow verified: `loss.backward()` propagates to encoder parameters

### DIFF
--- a/pytorch_forecasting/models/tide/_tide.py
+++ b/pytorch_forecasting/models/tide/_tide.py
@@ -330,7 +330,7 @@ class TiDEModel(BaseModelWithCovariates):
             # from (batch size, seq len, output_dim) to
             # (output_dim, batch size, seq len)
             prediction = prediction.permute(2, 0, 1)
-            prediction = [i.clone().detach().requires_grad_(True) for i in prediction]
+            prediction = list(prediction)
 
         # rescale predictions into target space
         prediction = self.transform_output(prediction, target_scale=x["target_scale"])


haoyu-haoyu:
Thanks for the approval @phoeenniixx! Good question about `.clone()`.

Looking at the code flow:
- `prediction.permute(2, 0, 1)` creates a **view** (not a copy)
- `for i in prediction` yields slices that are also **views** into the same memory
- Without `.clone()`, these views share the underlying storage

Whether `.clone()` is needed depends on whether `transform_output()` does any in-place operations on the prediction tensors. If it does, modifying one slice could corrupt another since they share memory.

**My recommendation:** keep `.clone()` for safety since we can't easily guarantee no in-place ops downstream. The memory overhead is minimal (one extra copy of the prediction tensor), and it prevents subtle bugs if the code is later modified.

That said, if you'd prefer to remove it for memory efficiency, the minimal change would be:
```python
prediction = [i.requires_grad_(True) for i in prediction]
```
This keeps the views (no copy) and just enables gradient tracking. Happy to update either way!

haoyu-haoyu:
After tracing the full call chain more carefully, I think you're right — `.clone()` is NOT needed here:

1. `transform_output()` → `loss.rescale_parameters()` → `encoder()` all create **new tensors** (de-normalization is not in-place), so views are safe
2. `.requires_grad_(True)` is also redundant — the views already inherit `requires_grad=True` from the autograd graph through `permute()`

The simplest correct code is just:
```python
prediction = list(prediction)
```

Want me to update the PR to remove both `.clone()` and `.requires_grad_(True)`?

haoyu-haoyu:
Updated — simplified to `prediction = list(prediction)` per your suggestion.

The `.clone()` and `.requires_grad_(True)` were both unnecessary:
- Views from `permute()` already share the autograd graph
- `transform_output()` → `rescale_parameters()` creates new tensors (no in-place ops)
- Gradient tracking is inherited from the original model output

One line, zero overhead.

Issue #2212, opened 2026-03-19
Runtime crash in Samformer initialization when metadata is None


lomesh2312:

fnhirwa:
I think we need to revise the docstring of this model as they're too ambiguous to know how the metadata and other arguments are handled. Could you include that in your PR as well?

lomesh2312:
@fnhirwa 
Thanks for the feedback!

Yes, that makes sense. I’ll update the docstring to clearly explain how metadata is expected to be provided and used in the model.

I’ll push the changes shortly.

lomesh2312:
@fnhirwa 
I've pushed the updates including the improved docstring and a small unit test for metadata validation. Please let me know if anything else needs to be adjusted.

PR #2213, opened 2026-03-19
[BUG] Add metadata validation in Samformer initialization

lomesh2312:
This PR addresses Issue #2212 where the `Samformer` model crashes upon initialization if `metadata` is `None`.

### Changes Made:
- Added a `ValueError` validation check in `__init__` for the `Samformer` model to immediately raise an appropriate error message if `metadata` falls back to `None`.
- Formatted code to adhere to `ruff` 88 character line-length limits.

All local tests and linters pass successfully.

### DIFF
--- a/tests/models/test_samformer.py
+++ b/tests/models/test_samformer.py
@@ -1,0 +1,10 @@
+import pytest
+import torch.nn as nn
+
+from pytorch_forecasting.models.samformer._samformer_v2 import Samformer
+
+
+def test_samformer_metadata_validation():
+    """Verify that Samformer raises ValueError when initialized with metadata=None."""
+    with pytest.raises(ValueError, match="Samformer requires 'metadata' dictionary"):
+        Samformer(loss=nn.MSELoss(), hidden_size=512, use_revin=True, metadata=None)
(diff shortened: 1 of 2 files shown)


fnhirwa (commented):
this looks goof but I don't understand why you changed some optional arguments to be required in the docstring.

also let's see what @phoeenniixx thinks about this

lomesh2312:
@fnhirwa 
Thanks for pointing that out.

You're right, I unintentionally removed the "optional" designation for some arguments like `hidden_size` and `use_revin` in the docstring, even though they still have default values and are not required.

My intention was only to clarify that `metadata` is required for proper initialization, but I see how this introduced inconsistency in the documentation.

I'll update the docstring to restore the optional arguments correctly and keep only `metadata` clearly documented as required.

Thanks again for catching this.

tests/models/test_samformer.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+import torch.nn as nn
+
+from pytorch_forecasting.models.samformer._samformer_v2 import Samformer
+
+
+def test_samformer_metadata_validation():
```
phoeenniixx, line 7: do we really need to add a standalone test for this validation? I mean this is not that hard to catch?
What do you think @lomesh2312 @fnhirwa

phoeenniixx:
Thanks! LGTM, just not sure if we should add a special file with a standalone test for this specific issue...

lomesh2312:
@phoeenniixx  Good point, that makes sense.

I added the standalone test mainly to explicitly cover the validation case, but I agree it might be unnecessary if it's already implicitly covered or considered too trivial.

I'm happy to remove or integrate it into existing tests if that's preferred. Please let me know what you'd suggest.

PR #2220, opened 2026-03-20
[ENH] Add pretrain/fine-tune lifecycle to BaseModel v2

### LINKED ISSUE #2105: [ENH] Extend TFT v2 with pretrain/fine-tune lifecycle as proof-of-concept for #41
## Summary

This issue proposes extending the existing experimental TFT v2 implementation 
(introduced in #1812) to support a `pretrain` / `fine-tune` lifecycle, serving 
as a concrete proof-of-concept for the API design being discussed in 
[STEP #41](https://github.com/sktime/enhancement-proposals/pull/41) and sktime/sktime#6997.

## Motivation

[STEP #41](https://github.com/sktime/enhancement-proposals/pull/41) proposes a `pretrain → fit → predict` lifecycle for global and 
foundation-model-capable forecasters. The design is currently at the proposal 
stage with no concrete implementation to validate it against. 

TFT is the ideal candidate to prototype this for three reasons:

1. **It is the flagship model of pytorch-forecasting** — validating the API 
   against TFT stress-tests it against one of the most complex architectures 
   in the library (gating layers, variable selection networks, multi-head 
   attention, static enrichment).

2. **TFT is natively a global forecaster** — it was designed for multi-series 
   training with static and time-varying covariates, making it directly relevant 
   to the "global learning" part of Project 3.1, not just the pretrain/fine-tune 
   part.

3. **An experimental v2 TFT already exists** (#1812) — this issue builds on 
   that foundation rather than starting from scratch, keeping the scope 
   manageable and the PR reviewable.

---

## Current State

The experimental TFT v2 from #1812 uses the new D1/D2 layer and pkg wrapper 
architecture. However it:

- Has no `pretrain()` method or pretrained weight loading support
- Has no `is_pretrained` state tracking
- Has no `fit_strategy` routing (fine-tune vs train from scratch)
- Cannot accept a `pretrained_weights` path/repo-id at `__init__` time
- Has no `capability:pretraining` or `capability:global_forecasting` tag 
  registered with skbase

---

### 1. Add `pretrained_weights` as an `__init__` parameter
```python
class TemporalFusionTransformerV2(BaseModel):
    def __init__(
        self,
        hidden_size: int = 16,
        lstm_layers: int = 1,
        dropout: float = 0.1,
        # ... existing params ...
        pretrained_weights: str | None = None,  # NEW
        # local path, URL, or HuggingFace repo ID
    ):
```

### 2. Add `pretrain()` method to the v2 model layer
```python
def pretrain(self, datamodule, trainer_params=None):
    """
    Pre-train on a large corpus DataModule.
    Sets self.is_pretrained_ = True after completion.
    Does NOT set is_fitted = True.
    Weights accessible via self.network.state_dict().
    """
```

### 3. Add fine-tune routing in `fit()`
```python
def fit(self, datamodule, trainer_params=None):
    """
    If is_pretrained_ is True, load weights and fine-tune 
    (frozen backbone, trainable head by default).
    Otherwise, train from scratch.
    """
    if getattr(self, "is_pretrained_", False):
        self._load_pretrained_weights()
        self._freeze_backbone()
    # ... proceed with training
```

### 4. Register skbase capability tags
```python
_tags = {
    "capability:global_forecasting": True,
    "capability:pretraining": True,      # new tag per STEP #41
}
```

### 5. Update the pkg layer wrapper

The `pkg` wrapper (sktime-facing layer) should expose:
- `pretrain(y, X=None)` that delegates to the model's `pretrain()`
- `pretrained_weights` as a passthrough `__init__` parameter

---

## Implementation Plan

- [ ] Study existing TFT v2 code from #1812 and D1/D2 architecture from #1811
- [ ] Add `pretrained_weights` param and weight loading utilities 
      (local path + HuggingFace `hf://` prefix)
- [ ] Implement `pretrain()` on model layer
- [ ] Implement fine-tune routing in `fit()`
- [ ] Add `capability:pretraining` tag and ensure skbase test contracts 
      run for it
- [ ] Update the pkg layer wrapper
- [ ] Add unit tests: pretrain → fit → predict lifecycle; cold-start 
      (pretrained_weights only, no pretrain call); train from scratch 
      (backward compat)
- [ ] Add a minimal notebook/docstring example showing the three usage modes

---

Before raising a draft PR for this I had some doubts, how should I approach the following problems?

1. Should `pretrain()` live on the `Model` layer (ptf-v2 internal) or only 
   on the `pkg` wrapper (sktime-facing)? I am currently leaning toward the 
   Model layer so it can be used independently of sktime, with the pkg layer 
   being a thin delegation.

2. For weight freezing during fine-tune: should the default be 
   (a) freeze backbone, train head only, or (b) fine-tune all layers with 
   a lower LR? Or should we expose a `fine_tune_strategy` parameter?

cc @phoeenniixx @agobbifbk @fkiraly

echo-xiao:
#### Reference Issues/PRs
close #2105 

Implements the pretrain -> fine-tune -> predict lifecycle for basemodel v2. 

#### Changes:

_base_model_v2.py
- pretrained_weights init param 
- fine_tune_strategy init param
- pretrain(): pre-trian on gloabl/panel data
- _pretrain(): hook for subclasses to override
- _post_init_laod_pretrained(): subclaases call
- load_ptrtrained_weights(): load HuggingFace path
- _freeze_backbone() / _unfreeze_backbone(): freeze_utilities
- fine_tune_strategy param
- three mode docstring examples (pretrain->finetune, cold-start, train from scratch)

1. design of _pretrain() hook: most important thing
2. find_tune_strategy default value
3. laod_pretrained_weights(), is this pattern acceptable for all v2 subclasses?
4. post_init_load_pretrain() pattern, is this pattern acceptable for all v2 subclasses?
5. if all of these changes match https://github.com/sktime/enhancement-proposals/pull/41

tests/test_models/test_basemodel_pretrain.py
1. pretrain() sets and returns self
2. _pretrain() subclass hook
3. weights change after pretraining
4. load_pretrained_weights() with difference format

#### Any other comments?

NHiTS_v2 POC integration will be added as follow-up after #2186 is merged.

#### PR checklist

- [x] Added/modified tests


pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -343,6 +386,221 @@ def _get_scheduler(
...
+    def _unfreeze_backbone(self) -> None:
+        """Unfreeze all model parameters.
+
+        Call this after fine-tuning if you want to continue training all
+        layers (e.g., for a second fine-tuning phase at a lower LR).
+        """
+        for param in self.parameters():
+            param.requires_grad_(True)
+
+    def _post_init_load_pretrained(self) -> None:
```
phoeenniixx, line 508: Where (and how) is this method used? I mean why not directly use `load_pretrained_weights`?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -40,6 +40,45 @@ class BaseModel(LightningModule):
...
+
+    Examples
+    --------
+    **Mode 1 — pretrain on large corpus, then fine-tune on target dataset:**
+
+    >>> model = MyModel(loss=MAE(), metadata=dm_large.metadata)  # doctest: +SKIP
+    >>> model.pretrain(dm_large, trainer_kwargs={"max_epochs": 20})  # doctest: +SKIP
+    >>> from lightning import Trainer  # doctest: +SKIP
+    >>> trainer = Trainer(max_epochs=5)  # doctest: +SKIP
+    >>> trainer.fit(model, dm_target)  # backbone frozen automatically  # doctest: +SKIP
```
phoeenniixx, line 65: Would it be better to use `Basepkg` here?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -40,6 +40,45 @@ class BaseModel(LightningModule):
...
+
+    **Mode 2 — cold-start: load community pretrained weights, skip pretrain():**
+
+    >>> model = MyModel(  # doctest: +SKIP
+    ...     loss=MAE(),
+    ...     metadata=dm_target.metadata,
+    ...     pretrained_weights="hf://org/my-pretrained-model/weights.pt",
+    ... )
+    >>> trainer = Trainer(max_epochs=5)  # doctest: +SKIP
+    >>> trainer.fit(model, dm_target)  # fine-tunes from pretrained  # doctest: +SKIP
```
phoeenniixx, line 75: Would it be better to use `Basepkg` here?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -343,6 +386,221 @@ def _get_scheduler(
         else:
             raise ValueError(f"Scheduler {self.lr_scheduler} not supported.")
 
+    def pretrain(
```
phoeenniixx, line 389: should it be in `Basepkg` ?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -343,6 +386,221 @@ def _get_scheduler(
...
+
+        Default behaviour: freeze all parameters in ``self.model`` if that
+        attribute exists; otherwise freeze every parameter in the module.
+
+        Subclasses with a clear backbone / head split should override this
+        method to freeze only the backbone, keeping the head trainable.
+        """
+        target = getattr(self, "model", self)
+        for param in target.parameters():
+            param.requires_grad_(False)
```
phoeenniixx, line 497: What if I just want to freeze just a part of the model? Like first few layers?

phoeenniixx (changes requested):
Thanks! This looks good! 
Sorry, I don't have the complete mental model ready of how pretraining should look like... But I have added some doubts (some of which may feel redundant :))
- How the inference mode would look here? After freezing the backbone? 
- Also, if we load weights from hf, how should we finetune the model? I think there can be multiple strats?
- I think we should think of how to integrate this with the FMs, I am not sure if the current `BaseModel` is a good place for this? 
- Should the `pretrain` etc methods be in `Basepkg` similar to the way `fit`, `predict` is?

FYI @agobbifbk, @PranavBhatP, @fkiraly  what are your thoughts on this?

Also, please dont share the GSoC proposal here, I think you should follow the official channels and apply at the GSoC portal and the `sktime` form...

pytorch_forecasting/models/base/_base_model_v2.py
echo-xiao: I see, you are right. pretrain() and _pretrain() should move to Base pkg to be consistant with how fit() and train() is constructed. Basemodel should only keep the lifecycle state and the freeze/ewight-loading hooks.

pytorch_forecasting/models/base/_base_model_v2.py
echo-xiao: Subclasses can override it.

pytorch_forecasting/models/base/_base_model_v2.py
echo-xiao: When Basemodel.__init__ runs, self.model doesen't exists yet. _post_init_load_pretrained() is called after the model is fully built.

echo-xiao:
Thanks for @phoeenniixx 's feedback. For the problems below:

1. How the inference mode would look here? After freezing the backbone?
-- Inference works as normal. Frozen parameters still participate in the forward pass. Training only updates unfrozen parts. Frozen parameters are not updated during training.


2. Also, if we load weights from hf, how should we finetune the model? I think there can be multiple strats?
-- Currently supports freeze_backbone and full. Other strategies can be added as follow-ups.   


3. I think we should think of how to integrate this with the FMs, I am not sure if the current BaseModel is a good place for this?
-- This PR provides the base infrastructure (load_pretrained_weights, is_pretrained_, freeze_backbone) for FM integration to build on. Happy to align with the FM project contributors on interface design.


4. Should the pretrain etc methods be in Basepkg similar to the way fit, predict is?
-- Agreed

echo-xiao:
@phoeenniixx @fkiraly 

I recieved the email about my application is not completed. I have just completed and sumitted it. I kindly ask if you could reconsider my application given the forms are now complete. 

I am a Data Scientist who has logn relied on sktime in my professional work. And my proposal outlines my deep architectural vision for the project. Please give my proposal a chance if you could.

PR #2226, opened 2026-03-23
[BUG] Move prediction tensors to CPU between batches in PredictCallback

### LINKED ISSUE #2224: [BUG] Prediction tensors not moved to CPU between batches, causing VRAM accumulation
**Describe the bug**

When running `model.predict()` on a CUDA device, `PredictCallback.on_predict_batch_end` collects each batch output into self.predictions without calling `.detach()` or `.cpu()`:

[Bug Lives Here:](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/callbacks/predict.py#L69)

```python
```

Every prediction tensor stays pinned in VRAM until `on_predict_epoch_end` runs at the very end of inference. The same issue affects `return_info` tensors at [here](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/callbacks/predict.py#L73-L79).

While reviewing I found that, PredictCallback was introduced in [#1984](https://github.com/sktime/pytorch-forecasting/pull/1984) as the v2 predict path and the v1 base model had an explicit guard at [here](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model.py#L310-L311) that moved outputs off-device after every batch:

```python
x   = move_to_device(detach(x),   "cpu")
```

But this pattern was not carried over into the v2 callback.

&nbsp;

**To Reproduce**

Full repro script (any CUDA GPU): [Gist here](https://gist.github.com/StrikerEureka34/19f78514c1a3fbb39b7f353d67599f37)
Result I got while running it locally are:-

```python
# device: cuda
# VRAM before loop: 0.0 MB
# batches collected : 127
# tensor shape      : torch.Size([64, 24])
# tensor device     : cuda:0
# VRAM held by 127 prediction tensors: 9.44 MB
# Expected: ~0 MB (tensors should be on CPU). Got: 9.44 MB on GPU.
# With a real FM dataset (1000+ series, multivariate), this accumulates to OOM.
```

&ensp;

**Expected behavior**

So here `cb.predictions[0].device` should be cpu after each batch. But right now it is `cuda:0`, which means every batch tensor accumulates in VRAM for the entire inference run.

&ensp;

**Proposed Fix**

Add '.detach().cpu()' before each append at [here](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/callbacks/predict.py#L69) for predictions, and at [here](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/callbacks/predict.py#L73-L79) for return_info tensors (y, decoder_lengths) and a dict comprehension over x since it contains mixed tensor and non-tensor values.

&ensp;

**Additional context**

For FM-scale workloads(#1959), inference over thousands of series runs hundreds of batches per call. 
This accumulation is linear, for example 127 batches on this test already held 9.44 MB above model weight allocations. 

What is worse is that on a smaller GPU or with multivariate targets this quietly exhausts VRAM mid-inference with no message pointing to PredictCallback as the cause, which will cause a lot of confusion as it did to me while working with a bigger timeseries data.

If this bug is verified, from the reproducible script as attached, I'm happy to submit a PR for the same, thanks!
@phoeenniixx @fnhirwa @PranavBhatP

FarahAbdo:
Fixes #2224

## Problem
`PredictCallback.on_predict_batch_end` collects batch outputs into
`self.predictions` and `self.info` without moving tensors off the GPU.
Every tensor stays pinned in VRAM until `on_predict_epoch_end`, causing
linear VRAM accumulation across batches.

The v1 base model had an explicit guard:
```python
x   = move_to_device(detach(x),   "cpu")
```

This pattern was not carried over into the v2 PredictCallback.

Fix
Added a recursive _move_to_cpu static method that handles tensors,
dicts (like x which contains mixed tensor/non-tensor values),
lists, and tuples.

Applied _move_to_cpu to all appends in on_predict_batch_end:
predictions, x, y, index, and decoder_lengths.

No changes needed in on_predict_epoch_end since tensors are
already on CPU by the time collation runs.

Testing
Validated against the reproduction script from #2224.
After fix: cb.predictions[0].device returns cpu and VRAM stays
at ~0 MB between batches.

cc @phoeenniixx @fnhirwa @PranavBhatP

### DIFF
--- a/pytorch_forecasting/callbacks/predict.py
+++ b/pytorch_forecasting/callbacks/predict.py
@@ -1,12 +1,10 @@
 from typing import Any, Optional
 from warnings import warn
 
-
 from lightning import Trainer
 from lightning.pytorch import LightningModule
 from lightning.pytorch.callbacks import BasePredictionWriter
 import torch
-
 
 
 class PredictCallback(BasePredictionWriter):
@@ -30,7 +28,6 @@ class PredictCallback(BasePredictionWriter):
         Additional keyword arguments for `to_prediction` or `to_quantiles`.
     """
 
-
     def __init__(
         self,
         mode: str = "prediction",
@@ -43,14 +40,12 @@ class PredictCallback(BasePredictionWriter):
         self.mode_kwargs = mode_kwargs or {}
         self._reset_data()
 
-
     def _reset_data(self, result: bool = True):
         """Clear collected data for a new prediction run."""
         self.predictions = []
         self.info = {key: [] for key in self.return_info}
         if result:
             self._result = None
-
 
     @staticmethod
     def _move_to_cpu(obj):
@@ -73,7 +68,6 @@ class PredictCallback(BasePredictionWriter):
             return type(obj)(PredictCallback._move_to_cpu(v) for v in obj)
         return obj
 
-
     def on_predict_batch_end(
         self,
         trainer: Trainer,
@@ -86,7 +80,6 @@ class PredictCallback(BasePredictionWriter):
         """Process and store predictions for a single batch."""
         x, y = batch
 
-
         if self.mode == "raw":
             processed_output = outputs
         elif self.mode == "prediction":
@@ -96,9 +89,7 @@ class PredictCallback(BasePredictionWriter):
         else:
             raise ValueError(f"Invalid prediction mode: {self.mode}")
 
-
         self.predictions.append(self._move_to_cpu(processed_output))
-
 
         for key in self.return_info:
             if key == "x":
@@ -112,7 +103,6 @@ class PredictCallback(BasePredictionWriter):
             else:
                 warn(f"Unknown return_info key: {key}")
 
-
     def on_predict_epoch_end(self, trainer: Trainer, pl_module: LightningModule):
         """Collate all batch results into final tensors."""
         if self.mode == "raw" and isinstance(self.predictions[0], dict):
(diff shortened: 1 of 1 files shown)


StrikerEureka34:
Hi @FarahAbdo , thankyou for this, actually I was already working on this and the Bug still needs validation from the maintainers, so raising a PR seems premature at this stage.

In the meantime you can work on other active issues, thanks!

FarahAbdo:
Totally understand @StrikerEureka34 ! I saw the issue was unassigned and in "Needs triage & validation" so I went ahead. Happy to close this PR if you'd prefer to submit one yourself, or the maintainers can decide which approach they prefer. Either way, the fix is there if it's useful 🙂

StrikerEureka34:
I see, thanks for verifying the bug and for the detailed write-up on the fix, really appreciate it! 
I actually stumbled upon this yesterday while working with a larger timeseries dataset, so the timing overlapped. I have a draft PR in progress but still ironing out a few things before I push it and also waiting on for maintainer's validation. 
Would love to get your thoughts on it once it's up! 🙂

One thing that might help in future, a quick comment on the issue asking if anyone's already working on it can save a lot of overlap. Just a thought! 😊

fnhirwa:
closing in favor of https://github.com/sktime/pytorch-forecasting/commit/af9b14f34cacd3ad37ded9034f6d3f66db9aee6c

PR #2228, opened 2026-03-23
[BUG] Move prediction tensors to CPU between batches in PredictCallback

### LINKED ISSUE #2224: [BUG] Prediction tensors not moved to CPU between batches, causing VRAM accumulation

StrikerEureka34:
#### Reference Issues/PRs
Fixes #2224

As described in #2224, tensors collected in `on_predict_batch_end` were staying on the GPU between batches causing VRAM to grow linearly across the inference run.

Fixed by applying `move_to_device(detach(...), "cpu")` at each append site using
utilities already in pytorch_forecasting.utils, the same pattern already used in the v1 base model.  `x_cpu` is computed once and reused for both x and decoder_lengths to avoid reading from the GPU dict twice.
&nbsp;
Just the changes in `on_predict_batch_end` in `pytorch_forecasting/callbacks/predict.py`.
Nothing else was touched.
&nbsp;
Yes, I've added `tests/test_callbacks.py` with 4 tests covering prediction mode, raw mode dict output, the x dict, and y/decoder_lengths.

There was no existing test file for callbacks so created a standalone `test_callbacks.py` directly under `tests/`, following the same pattern as `test_metrics.py` rather than adding a subdirectory for a single file.

Since these run on CPU (no CUDA needed in CI), the device check alone would pass trivially even without the fix. 
To get around this, I've created tensors as non-leaf (`requires_grad=True` + an addition op) so they have a real `grad_fn`. 
The `assert grad_fn is None` then only passes if `.detach()` actually ran, making these tests meaningful on CPU too.
&nbsp;
#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/callbacks/predict.py
+++ b/pytorch_forecasting/callbacks/predict.py
@@ -6,6 +6,8 @@ from lightning.pytorch import LightningModule
 from lightning.pytorch.callbacks import BasePredictionWriter
 import torch
 
+from pytorch_forecasting.utils import detach, move_to_device
+
 
 class PredictCallback(BasePredictionWriter):
     """
@@ -66,17 +68,20 @@ class PredictCallback(BasePredictionWriter):
         else:
             raise ValueError(f"Invalid prediction mode: {self.mode}")
 
-        self.predictions.append(processed_output)
+        self.predictions.append(move_to_device(detach(processed_output), "cpu"))
+
+        # detach(x) copies the dict; move_to_device won't mutate the original batch
+        x_cpu = move_to_device(detach(x), "cpu")
 
         for key in self.return_info:
             if key == "x":
-                self.info[key].append(x)
+                self.info[key].append(x_cpu)
             elif key == "y":
-                self.info[key].append(y[0])
+                self.info[key].append(move_to_device(detach(y[0]), "cpu"))
             elif key == "index":
-                self.info[key].append(y[1])
+                self.info[key].append(move_to_device(detach(y[1]), "cpu"))
             elif key == "decoder_lengths":
-                self.info[key].append(x["decoder_lengths"])
+                self.info[key].append(x_cpu["decoder_lengths"])
             else:
                 warn(f"Unknown return_info key: {key}")
 
--- a/pytorch_forecasting/callbacks/predict.py
+++ b/pytorch_forecasting/callbacks/predict.py
@@ -70,16 +70,19 @@ class PredictCallback(BasePredictionWriter):
 
         self.predictions.append(move_to_device(detach(processed_output), "cpu"))
 
-        # detach(x) copies the dict; move_to_device won't mutate the original batch
-        x_cpu = move_to_device(detach(x), "cpu")
+        # Only pay the detach+copy cost if x or decoder_lengths are actually requested
+        needs_x = any(k in ("x", "decoder_lengths") for k in self.return_info)
+        x_cpu = move_to_device(detach(x), "cpu") if needs_x else None
 
         for key in self.return_info:
             if key == "x":
                 self.info[key].append(x_cpu)
             elif key == "y":
-                self.info[key].append(move_to_device(detach(y[0]), "cpu"))
+                y_cpu = move_to_device(detach(y[0]), "cpu")
+                self.info[key].append(y_cpu)
             elif key == "index":
-                self.info[key].append(move_to_device(detach(y[1]), "cpu"))
+                index_cpu = move_to_device(detach(y[1]), "cpu")
+                self.info[key].append(index_cpu)
             elif key == "decoder_lengths":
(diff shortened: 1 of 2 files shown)


StrikerEureka34:
Hi @fnhirwa @phoeenniixx , I think this should be a complete fix, it uses the existing move_to_device and detach utilities from pytorch_forecasting.utils and includes tests to avoid silent regression.

I reached out to @FarahAbdo to incorporate anything useful from #2226 but haven't heard back yet, ready to do so if they respond.

Would really appreciate your review when you get a chance, thanks!

pytorch_forecasting/callbacks/predict.py
```diff
@@ -66,17 +68,20 @@ def on_predict_batch_end(
...
+        # detach(x) copies the dict; move_to_device won't mutate the original batch
+        x_cpu = move_to_device(detach(x), "cpu")
 
         for key in self.return_info:
             if key == "x":
-                self.info[key].append(x)
+                self.info[key].append(x_cpu)
             elif key == "y":
-                self.info[key].append(y[0])
+                self.info[key].append(move_to_device(detach(y[0]), "cpu"))
```
fnhirwa: Just a simple modification, can we try to isolate the movement to the device from the append function itself?

just something like 
```python
y_cpu = move_to_device(detach(y[0]), "cpu")
# then append
self.info[key].append(y_cpu)
```

StrikerEureka34: > Just a simple modification, can we try to isolate the movement to the device from the append function itself?

Thankyou so much for the review!
Makes sense, will do the same for y_cpu and index_cpu. Also planning to guard x_cpu to only compute when actually needed.

StrikerEureka34:
I've made the changes as requested, please let me know if this implementation holds, thanks! @fnhirwa @phoeenniixx

fnhirwa (approved):
looks good to me

PR #2232, opened 2026-03-25
[ENH] softs_v2 Model added

### LINKED ISSUE #2231: [ENH] Add v2 interface support for `SOFTS (Star Aggregate-Dispatch for Time Series Forecasting)`
Referencing #1992

This issue tracks adding v2 interface support for **SOFTS (Star Aggregate-Dispatch for Time Series Forecasting)**.

**SOFTS** is a state-of-the-art multivariate time series forecasting model introduced at **AAAI 2024** by the THUML group. It proposes a novel Star Aggregate-Dispatch (STAD) mechanism that uses a global star prototype to capture inter-series dependencies efficiently, enabling highly scalable forecasting across large numbers of variates while keeping computational cost linear. Given the `v2 interface's goal of modularity and scalability`, **SOFTS** is a meaningful addition as it bridges channel-wise and full-attention paradigms.

**Plan:**

- Implement the core SOFTS STAD module and encoder blocks in the layers directory.
- Implement the forecasting-specific estimator in the models directory, following the v2 pattern used by `DLinear` and `TFT`.
- Ensure the STAD global prototype mechanism correctly handles arbitrary numbers of variates under the v2 interface constraints.
- Ensure compliance with the v2 requirement for explicit dependency management/interfacing.
- Reference: Han, L., et al. (2024). SOFTS: Efficient Multivariate Time Series Forecasting with Series-Core Fusion. https://arxiv.org/abs/2404.14197

CC: @fkiraly  @PranavBhatP  @phoeenniixx

Muhammad-Rebaal:
Fixes #2231 

Hi @fkiraly , @phoeenniixx, @PranavBhatP !

I have implemented the SOFTS (Star Aggregate-Dispatch for Time Series Forecasting) model within the PyTorch Forecasting v2 architecture.

Could you please review the PR?

- Isolated Layer Abstraction `(pytorch_forecasting/layers/_blocks/_softs_block.py)`: Implemented the core neural network components, specifically the novel STADModule (Star Aggregate-Dispatch mechanism) and the SoftsEncoderLayer, completely isolated from the base estimating logic to strictly adhere to the project's v2 architectural standards.
- Model Construction `(pytorch_forecasting/models/softs/)`: Created the main Softs estimator wrapper which inherits from TslibBaseModel. It effectively handles dynamic feature input alignments (history_cont and history_target), optionally applies RevIN scaling, and routes the forward pass through the STAD-based encoders.
- V2 Package Management `(_softs_pkg_v2.py)`: Established the Softs_pkg_v2 class representing model metadata (capability tags, compute requirements) and defined automated testing configurations (get_test_train_params) tightly integrated with the underlying TslibDataModule.
- Registry & Exports: Updated the relevant `__init__.py` files across the layers and models directories to properly expose Softs, Softs_pkg_v2, and the STAD blocks to the broader PyTorch Forecasting ecosystem.

Thank you!


pytorch_forecasting/layers/_blocks/_softs_block.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Add dispatched info back to original sequence (broadcasting over length)
+        dispatch_out = dispatch_out.unsqueeze(2).repeat(1, 1, L, 1)
+
+        return x + dispatch_out
+
+
+class SoftsEncoderLayer(nn.Module):
+    """
+    Single Encoder layer for SOFTS.
```
phoeenniixx: Can you please add more detailed docstrings for all methods and classes?

pytorch_forecasting/models/softs/_softs_pkg_v2.py
```diff
@@ -0,0 +1,113 @@
...
+        from pytorch_forecasting.models.softs._softs_v2 import Softs
+
+        return Softs
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data._tslib_data_module import TslibDataModule
+
+        return TslibDataModule
```
phoeenniixx: Why are we using this datamodule? I think this data module is mainly for `tslib` models, I have no issue with using this, but pls have a look at `EncoderDecoderDataModule` as well, maybe that would also be helpful.
I have not looked at the architecture, so I have a question: is the model encoder-decoder based model?

pytorch_forecasting/models/softs/_softs_pkg_v2.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @classmethod
+    def _get_test_datamodule_from(cls, trainer_kwargs):
```
phoeenniixx: I dont think we need this anymore

pytorch_forecasting/models/softs/_softs_pkg_v2.py
```diff
@@ -0,0 +1,113 @@
...
+            "val": val_datamodule.val_dataloader(),
+            "test": test_datamodule.test_dataloader(),
+            "data_module": train_datamodule,
+        }
+
+    @classmethod
+    def get_test_train_params(cls):
+        params = [
+            {},
+            dict(hidden_size=128, n_layers=1, use_revin=True),
```
phoeenniixx: Can you try adding some more params to cover more scenarios?

pytorch_forecasting/models/softs/_softs_v2.py
```diff
@@ -0,0 +1,150 @@
...
+
+from pytorch_forecasting.layers._blocks._softs_block import SoftsEncoderLayer
+from pytorch_forecasting.layers._normalization import RevIN
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class Softs(TslibBaseModel):
+    """
+    SOFTS: Efficient Multivariate Time Series Forecasting with Series-Core Fusion.
+    """
```
phoeenniixx, line 45: Add detailed docstrings

phoeenniixx (changes requested):
Thanks!
i would suggest adding detailed docstrings with clear descriptions of all the params, so that it is easy to understand the methods and the architecture in general.

Is this code taken from somewhere or is it completely implemented from scratch? I think if it is taken from somewhere, it would be better to add the original authors to the `authors` tag.

Also, please add the references (like paper, repo etc) to the docstrings  as well of the model.

pytorch_forecasting/models/softs/_softs_pkg_v2.py
Muhammad-Rebaal: We are using `TslibDataModule` because **SOFTS** is a direct-projection MLP model. It only consumes historical input sequences and maps them directly to the forecast window, using `EncoderDecoderModule` felt unnecessary.

From the perspective of data input we can call it encoder-only model.

Code Ref : [Code](https://github.com/Secilia-Cxy/SOFTS/blob/main/layers/Transformer_EncDec.py)

phoeenniixx: I think this can be seen as a encoder-decoder model with decoder being a identity layer?
I think we should use the `TslibDataModule` only if we are interfacing the model from the `tslib` package. Otherwise if it fits encoder-decoder model type, we should use `EncoderDecoderDataModule`

phoeenniixx (changes requested):
I think there are geniune failures, please take a look at that

fkiraly:
@phoeenniixx, I think there are no longer failures. Can you please re-review?

(quick question, do we need to add the model to the docs, or is that automatic now?)

phoeenniixx:
> (quick question, do we need to add the model to the docs, or is that automatic now?)

We still need to add the API reference, but the table generation is automatic now!

pytorch_forecasting/models/softs/_softs_v2.py
```diff
@@ -0,0 +1,177 @@
...
+import torch
+import torch.nn as nn
+from torch.optim import Optimizer
+
+from pytorch_forecasting.layers._blocks._softs_block import SoftsEncoderLayer
+from pytorch_forecasting.layers._normalization import RevIN
+from pytorch_forecasting.models.base._tslib_base_model_v2 import TslibBaseModel
+
+
+class Softs(TslibBaseModel):
```
phoeenniixx: Can the `BaseModel` not work here? I mean we mainly use `TslibBaseModel` for `tslib` models only

pytorch_forecasting/models/softs/_softs_pkg_v2.py
```diff
@@ -0,0 +1,87 @@
...
+            dict(
+                hidden_size=64,
+                n_layers=1,
+                loss=MAE(),
+            ),
+            dict(
+                hidden_size=64,
+                n_layers=1,
+                use_revin=False,
+                loss=MAE(),
```
phoeenniixx, line 77: It would be good if we could add other point prediction losses here as well - just to increase the coverage

phoeenniixx (changes requested):
Thanks! I think this one is almost ready. Just few nitpicks.
Please add the API reference  of the model here: http://github.com/sktime/pytorch-forecasting/blob/main/docs/source/m_layer_v2.rst#api-reference

pytorch_forecasting/models/softs/_softs_pkg_v2.py
Muhammad-Rebaal: Thanks for letting me know I've updated that

pytorch_forecasting/models/softs/_softs_v2.py
Muhammad-Rebaal: The model is actually part of the `tslib model family` that's why I used that. Is there any particular reason to use the base model instead ?

phoeenniixx: I thought `tslib` model family was from here: https://github.com/thuml/Time-Series-Library
But this model was taken from here:  https://github.com/Secilia-Cxy/SOFTS/
That is why I said it is not a `tslib` model. Am I missing something here?

pytorch_forecasting/models/softs/_softs_pkg_v2.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class Softs_pkg_v2(Base_pkg):
+    """
+    SOFTS package container.
+    Reference : https://arxiv.org/abs/2404.14197
+    """
+
+    _tags = {
+        "info:name": "SOFTS",
```
phoeenniixx, line 15: I think the name should be same as the class here: `Softs` in place of `SOFTS`?

Muhammad-Rebaal: I think it should be `SOFTS` all over the place as the actual model name is **SOFTS** not Softs and also the convention also follows that across other models.

pytorch_forecasting/models/softs/_softs_v2.py
Muhammad-Rebaal: Yeah, you're right I just misinterpreted thanks for letting me know. I'll adjust that and use the baseClass.

pytorch_forecasting/models/softs/_softs_pkg_v2.py
phoeenniixx: Then you would have to update the class name. The tag `name` should be exactly same as the class name

Muhammad-Rebaal:
Hi @phoeenniixx, 
I've updated the `class_name` to **SOFTS** and use the `base_class`, also updated the `api_reference`. Kindly have a look at it.
Thanks

pytorch_forecasting/models/softs/_softs_pkg_v2.py
phoeenniixx: You need to update this data module as well then

phoeenniixx (changes requested):
Please use `EncoderDecoderTimeSeriesDataModule` in place of `tslib` one, if we are using the `BaseModel` and not the `tslib` one, then we need to use the corresponding data module as well. Although the tests are passing, but both the data modules represent different ideologies, and with time as we add more feats will diverge from each other, so we shouldn't use them interchangeably

pytorch_forecasting/layers/_blocks/_softs_block.py
```diff
@@ -0,0 +1,114 @@
...
+        x_pooled = x.mean(dim=2)
+        core_node = torch.einsum("bcd,bce->bed", x_pooled, w)
+        core_node = self.channel_mixing(core_node)
+        dispatch_out = torch.einsum("bed,bce->bcd", core_node, w)
+
+        dispatch_out = dispatch_out.unsqueeze(2).repeat(1, 1, L, 1)
+        return x + dispatch_out
+
+
+class SOFTSEncoderLayer(nn.Module):
```
phoeenniixx: should it go to `_encoders`  folder?

pytorch_forecasting/models/softs/_softs_v2.py
```diff
@@ -0,0 +1,224 @@
...
+                out = out[:, :, target_indices, :]
+            else:
+                out = out[:, :, target_indices]
+
+        if "target_scale" in x and hasattr(self, "transform_output"):
+            out = self.transform_output(out, x["target_scale"])
+
+        return {"prediction": out}
+
+    def predict_step(
```
phoeenniixx: why do we need a `predict_step` here? can't the `BaseModel`'s implementation enough?
And if not, why do have not added other step functions like `test_step`?

phoeenniixx (changes requested):
Nice! I think it is  almost ready, just have few doubts (see above)

pytorch_forecasting/layers/_blocks/_softs_block.py
Muhammad-Rebaal: I considered this during implementation. Looking at the current `_encoders/` contents, both `Encoder` and `EncoderLayer` are TimeXer-specific (they take `cross`, `tau`, `delta` params and have global-token logic). 
Whereas `SOFTSEncoderLayer` has a fundamentally different interface it takes a 4D tensor `(B, C, L, D)` and uses STAD instead of attention, so there's no shared contract between them. `_blocks/` currently houses model-specific building blocks like `ResidualBlock` (for DSIPTs), and `SOFTSEncoderLayer` + `STADModule` follow the same pattern self-contained blocks specific to one model. Should I move it to a new `_softs/` subfolder under `layers/` to make the model association clear.

phoeenniixx: I see, but we can have multiple implementations of encoder layers, no? and that is why we created an `_encoders` folder that can host multiple implementations. It is not necessary that this encoder layer is just used by `SOFTS`, what if we see some new model that is derived from `SOFTS`, that could also use this layer. THe name - `SOFTSEncoderLayer` already makes the association pretty clear.

Muhammad-Rebaal: Yes, it can be possible that in the future we'd reuse it. I've adjusted that.

pytorch_forecasting/models/softs/_softs_v2.py
Muhammad-Rebaal: Yes, I thoroughly looked into the implementation. The `BaseModel's` implementation is enough.

Muhammad-Rebaal:
Hi @phoeenniixx,
I've made the following changes as mentioned:
- Moved the softs_encoder block to `_encoder` directory for future resuability.
- Removed the overriding `predict_step` method as **Base_models** implementation is enough.

Kindly have a look

phoeenniixx (approved):
Nice!

Issue #1993, opened 2025-11-28
Roadmap 2026


fkiraly:
From roadmap planning meeting on 2025-11-10

Present: @agobbifbk, @ankurankan, @fkiraly, @julian-fong, @phoeenniixx, @PranavBhatP, @RecreationalMath, @xaheli

### Topics & Prioritization

**pytorch-forecasting v2 API rework** :+1: :+1: :+1::+1: :heavy_check_mark: 
* Addition of `predict` function and full integration of `pkg` class with rest of D1,D2, model layer pipeline
* Addition of preprocessing steps - scaling, transformations etc in D1/D2 layers
* Improve the test coverage. 
* Add temporal splitting
* Migrate the models from v1 to v2 and deprecate v1
* Add extension Templates for new contributors.
* Add adpaters for `nn` losses.

**Extending model zoo** :+1::+1::+1::heavy_check_mark:
* Addition of new models to DSIPTS.

**Maintenance items, CI, tests, avoiding tech debt** :+1::+1: :heavy_check_mark:
* initial review of `pytorch-tabular` API, comparison to `pytorch-forecasting`
* Improve the test coverage in `pytorch-forecasting`
* negative tests and validation tests to be added for PyTorch based models. 
* Add python-3.14 support for ptf (scheduled for the next release).
* urgent maintenance items in `pytorch-tabular` - `pyproject`, python 3.14
* adding `torch` to all extras in `sktime`, or replace `tensorflow` by `torch`

**backend - `sktime` GPU support for `torch` based models** :+1:

* Support for GPU utilisation to be added

**API homogenization of `torch` models within `sktime`** :+1::+1::+1::heavy_check_mark:
* Unify scattered `torch` dataset classes for classification, regression etc.
* multiple `_pytorch.py` exist in sktime repo, they should all be unified into a single class in a single file with a `_pytorch_util.py` containing all the helper functions.
* activation, criterion, optimizer, callbacks initializations currently done crudely. Improvements to be made.

**API homogenization between `sktime`, `pytorch-forecasting`, `pytorch-tabular`, DSIPTS** :+1: :+1: :+1: :+1::heavy_check_mark::heavy_check_mark:

* uniform way of network creation and definition. API design that can be re-used everywhere.
* Create an interface for `pytorch-tabular` and `pytroch-forecasting` much like `prophetverse` and `sktime`
[...]

PR #2236, opened 2026-03-26
[ENH] auto-generate model overview table from registry tags

### LINKED ISSUE #1934: [ENH] automated model overview page
The documentation is outdated and does not list all models available in `pytorch-forecasting`.

We should update the model overview in `docs/source/models.rst` and use that all models are now tagged - so the table can be dynamically generated using `all_objects` from `registry` and no longer needs to be updated manually.

Inspiration can be taken from the `sktime` estimator overview
https://www.sktime.net/en/latest/estimator_overview.html

Though for `pytorch-forecasting`, a simpler approach is perhaps good enough - rendering the table that `all_objects` returns (with relevant tags), and attaching links to the estimator docs page.

IgnazioDS:
## Summary

Replaces the manually maintained CSV comparison table in `models.rst` with an auto-generated table built from model `_tags` metadata via the `all_objects()` registry.

- New Sphinx extension (`_ext/model_overview.py`) with `.. model-overview::` directive
- Queries the registry at docs build time to extract capability tags
- Renders an RST grid table with the same columns as the old manual table
- Adding/updating models in the docs is now automatic — just set the right `_tags`

## Motivation

The current table (lines 24-35 of `models.rst`) is manually maintained and doesn't include newer models like DLinear, TimeXer, Samformer, TFT v2, or TiDE v2. Every new model requires a manual edit to the docs, which is error-prone and easy to forget.

## How It Works

For model overview of v1 models
```rst
.. model-overview-v1::
```

At build time, this directive:
1. Calls `all_objects(return_tags=[...])` to discover all registered models
2. Extracts `capability:exogenous`, `capability:multivariate`, `capability:pred_int`, `info:pred_type`, `info:y_type`, `info:compute`
3. Renders a grid table with cross-references to model classes

## Columns

| Column | Source Tag |
|---|---|
| Name | `info:name` + module path for ``:py:class:`` cross-ref |
| Authors | `authors`|
| Covariates | `capability:exogenous` |
| Multiple targets | `capability:multivariate` |
| Regression | `"numeric" in info:y_type` |
| Classification | `"category" in info:y_type` |
| Probabilistic | `"distr" in info:pred_type` |
| Prediction intervals | `capability:pred_int` |
| Flexible History Length | `capability:flexible_history_length` |
| Cold Start | `capability:cold_start` |
| Compute (1-5) | `info:compute` |

## Test Plan

- [ ] `sphinx-build` completes without errors
- [ ] Generated table includes all registered models (RecurrentNetwork, DecoderMLP, NBeats, NBeatsKAN, NHiTS, DeepAR, TFT, TiDE, xLSTM, DLinear, TimeXer, Samformer, etc.)
- [ ] Cross-references (``:py:class:``) resolve correctly
- [ ] New models added with proper `_tags` automatically appear in the table

Fixes #1934


IgnazioDS:
@fkiraly Hi! This implements #1934 — auto-generating the model overview table from the registry using a custom Sphinx directive. Fixed the ruff lint issue in a follow-up commit. CI should be green now (readthedocs build still pending).

The approach mirrors how sktime handles automated docs generation — queries `all_objects()` with `return_tags` at build time and renders an RST grid table. Happy to adjust if you'd prefer a different format or additional tag columns!

phoeenniixx (changes requested):
Thanks a lot! This looks really nice!
I have one recommendation:
Would it be possible that the table shows:
- `Model_name` if `Model_pkg`
- `Model_v2` if `Model_pkg_v2`

This would be helpful for the users to differentiate between the "stable" v1 models and experimental v2 models. Also, pls donot show `pkg` name in the table, just the model name, i.e, if the name of the `pkg` class is `DecoderMLP_pkg` it should display `DecoderMLP` and if the name is `DLinear_pkg_v2` then it should show `DLinear_v2`.


OR
Would it be better to keep just the model names, but you add a new column where you add the v1, v2 tag. And explain this column below the table?

You could easily differentiate between v1, v2 pkg using the tag `"object_type"`, where v1 models use `["forecaster_pytorch", "forecaster_pytorch_v1"]` and v2 models use `"forecaster_pytorch_v2"`. See [here](https://github.com/sktime/pytorch-forecasting/blob/17c51ba909c7c89ba9eb0d2ce00eb5cae2ecb833/pytorch_forecasting/models/base/_base_object.py#L107-L120)

What are your thoughts? @IgnazioDS 
FYI @fkiraly @PranavBhatP

IgnazioDS:
Addressed your feedback @phoeenniixx:

- Model names now strip `_pkg` and `_pkg_v2` suffixes (e.g., `TFT_pkg` → `TFT`, `DLinear_pkg_v2` → `DLinear_v2`)
- Added a **Version** column using the `object_type` tag to clearly label v1 vs v2 models
- Removed the raw class name column entirely

Thanks for the clear guidance! Let me know if you'd prefer the alternative approach (separate v1/v2 column with footnote).

phoeenniixx (changes requested):
Very Nice! It is much better thanks!
I looked at the new page, it looks perfect!
But the links to the models still open the `pkg` class of the respective model, I think it should lead to the model's implementation...
Also, I think it would be better if you could add a footnote explaining what "version" mean here and that v2 is still experimental and link the issue #1736 and roadmap issue #1993? 

Also, I would suggest to keep your branch updated, that prevents any merge conflicts!

phoeenniixx (approved):
Thanks a lot @IgnazioDS !

phoeenniixx (changes requested):
I feel there are some discrepancies between the current "models" table and this one
- What do you mean by "Prediction intervals" is this same as "flexible history length" from the current table?
- I can't see "cold start" column in the new table. I think we should add that (and if the tags registry doesn't contain this tag, we should add this)

phoeenniixx:
Still need to add the tags to the v2 models to populate the tables correctly

phoeenniixx (dismissed):
Hi @fkiraly, @agobbifbk
I think this PR is ready. Now it creates 2 separate tables for 2 versions - now we can easily put the tables in two different places for both versions.
For now as we dont have a specific rst for v2 models, you wont see the v2 model-overview table, but it would look totally like this (as it uses the same logic) but the models would be the v2 models only

fkiraly (changes requested):
Very nice!

Could you explain though why we are changing some tags of current forecasters?

Maybe the tags should be changed in a separate PR (merged earlier), to ensure we separate tag changes from the documentation feature.

phoeenniixx:
We didnt had tags in the forecasters, they were missing, I just added them so that we could populate the table correctly

I will raise a new PR then for this, these are just few additions anyway

fkiraly (approved):
Looks good to me now!

fkiraly:
Should we also add this for version 2? As a separate table. Though that might be best in a separate PR.

phoeenniixx:
> Should we also add this for version 2? As a separate table. Though that might be best in a separate PR.

Actually I have already added an extension to add v2 table, we just need to add
```rst
.. model-overview-v2 ::
```
to add the table, this would be useful when we create the separate documentation in #2279

PR #2239, opened 2026-03-27
[BUG] Fix TslibBaseModel crash when metadata=None

### LINKED ISSUE #2217: [BUG] `TslibBaseModel` crashes when `metadata=None` (Root cause for Samformer issue #2212)
**Describe the bug**

Hi team! I was digging into the codebase and found a slight inconsistency in `_tslib_base_model_v2.py` that causes an `AttributeError` during initialization if `metadata=None` is passed. 

In the `__init__` method, we safely default empty metadata on line 57 using `self.metadata = metadata or {}`. However, a few lines further down, the code accidentally references the raw `metadata` parameter instead of the safely defaulted `self.metadata` attribute:

```python
# Line 57 — Safely handles None (Great!)
self.metadata = metadata or {}

# Lines 69-70 — Uses self.metadata (Correct)
self.context_length = self.metadata.get("context_length", 0)

# Lines 72, 79, 86, 89 — Uses the raw parameter (BUG)
# If metadata is None, these .get() calls will crash
feature_indices = metadata.get("feature_indices", {})  
feature_dims = metadata.get("n_features", {})            
self.feature_names = metadata.get("feature_names", {})   
self.features = metadata.get("features", "MS")           
```

**To Reproduce**
You can trigger this crash by initializing any v2 tslib model without metadata:
```python
from pytorch_forecasting.models.dlinear._dlinear_v2 import DLinear
from pytorch_forecasting.metrics import MAE

# This raises an AttributeError: 'NoneType' object has no attribute 'get'
model = DLinear(loss=MAE(), metadata=None)
```

**Expected behavior**
The base class should consistently use `self.metadata` for all `.get()` calls so it can safely fall back to the empty dictionary we defined at the top of the method. 

**Additional context**
I bring this up because this is actually the **root cause** of #2212! While #2212 specifically mentions `Samformer` crashing, this bug lives in the shared `TslibBaseModel`, meaning it currently affects **all** v2 subclasses (like `DLinear`, `TimeXer`, etc.). 

PR #2213 currently attempts to patch this for Samformer specifically, but I think a much cleaner, global fix would be to just update those 4 lines in the base class to use `self.metadata`. 

**Versions**
<details>

`pytorch-forecasting`: `main` (latest)

</details>

cc: @phoeenniixx @PranavBhatP — would appreciate any pointers if you have context on `_tslib_base_model_v2.py`

ArchieTyagi:
## What this PR does

Fixes `AttributeError: 'NoneType' object has no attribute 'get'` when initializing any v2 tslib model with `metadata=None`.

## The Bug

In `_tslib_base_model_v2.py`, `self.metadata = metadata or {}` safely defaults to an empty dict on line 57. However, lines 73, 79, 86, and 89 reference the raw `metadata` parameter instead of `self.metadata`, causing a crash when `metadata=None`.

## The Fix

Replaced 4 raw `metadata` references with `self.metadata`:
- `feature_indices = self.metadata.get("feature_indices", {})`
- `feature_dims = self.metadata.get("n_features", {})`
- `self.feature_names = self.metadata.get("feature_names", {})`
- `self.features = self.metadata.get("features", "MS")`

Fixes #2217
Related to #2212


fnhirwa (approved):
Looks good to me will merge once tests pass

PR #2240, opened 2026-03-28
 [ENH] Add gradient_clip_val to BaseModel v2 so clipping is model-owned and checkpoint-portable

### LINKED ISSUE #2237: [ENH] Add gradient_clip_val to BaseModel v2 so clipping is model-owned and checkpoint-portable

So I was training TimeXer on a multivariate dataset with exogenous variables, and I noticed that the loss was quite noisy and the training kept diverging and despite trying everything it didn't improve, so I did a bit of digging.....

And found that the natural fix for transformer-based models like this is gradient clipping. But the only way to do it right now is to pass `gradient_clip_val` in the `Trainer` config, which I did, and it worked but only because I remembered to re-add it manually every time I resumed training from a checkpoint. When I forgot once, training quietly resumed without clipping and diverged again before I caught it.

The core issue is that `gradient_clip_val` has no home in the model itself. 
It only lives in [Trainer config](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/tests/test_all_v2/utils.py#L42-L44) (where it is hardcoded into the shared test setup). This means:
 
- It is not saved in the checkpoint, so every resume requires you to reconstruct Trainer arguments from memory or docs.
- Models like `TimeXer`, `DLinear`, `SAMformer`, `TFT v2`, and `TIDE v2` cannot declare their own clipping needs. Every model is treated the same regardless of architecture, even though transformers are more sensitive to gradient explosions than linear models.

&ensp;

Add `gradient_clip_val: float | None = None` and `gradient_clip_algorithm: str = "norm"` to [BaseModel.__init__](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model_v2.py#L45)) and override [configure_gradient_clipping](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.core.LightningModule.html#lightning.pytorch.core.LightningModule.configure_gradient_clipping) (the Lightning hook called once per step before the optimizer):

```python
def configure_gradient_clipping(self, optimizer, gradient_clip_val=None, gradient_clip_algorithm=None):
    if self.gradient_clip_val is not None:
        total_norm = torch.nn.utils.clip_grad_norm_(self.parameters(), max_norm=float("inf"))
        self.log("grad_norm", total_norm, on_step=True, ...)
        self.clip_gradients(optimizer, gradient_clip_val=self.gradient_clip_val, ...)
    else:
        super().configure_gradient_clipping(optimizer, gradient_clip_val, gradient_clip_algorithm)
```

When gradient_clip_val=None (default), nothing changes, existing Trainer-based clipping still works. When set on the model, clipping is model-owned, saved in the checkpoint, and inherited by every subclass.

[TslibBaseModel](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_tslib_base_model_v2.py#L38-L56
) also needs the same two params added and forwarded to super().__init__(). The existing save_hyperparameters() call there will pick them up automatically with no other changes.

Add few tests to prevent silent regression in future.

&ensp;

Keeping it Trainer-only works for simple cases but fails at checkpoint resumption and makes it impossible for individual models to declare their own clipping requirements. A custom callback could apply clipping per step but would bypass Lightning's clip_gradients(), which handles AMP gradient unscaling and getting that right manually is fragile and still does not solve the checkpoint portability problem.

&ensp;

**Additional context**
The outcome of this will be:
- DLinear and TimeXer get gradient_clip_val in their constructor for free since they go through TslibBaseModel. , 
- SAMformer ,TFT v2, and TIDE v2 inherit the clipping behavior but each needs its own __init__ updated individually to expose the param
- For foundation models #1959 which extend BaseModel directly, having this there before they land means no retrofit work needed.

&ensp;

**Out of scope:** 
I discovered that Samformer, TFT v2, and TIDE v2 each manage their own save_hyperparameters() call independently, so each needs its own __init__ updated separately. Deferring to keep this PR focused.

Related: #2132, #1959 

Would really appreciate your insights on whether this is the right approach and if there's a better way to handle it before going further. @phoeenniixx @PranavBhatP

sujanyd:
Reference Issues/PRs
Fixes #2237
What does this implement/fix? Explain your changes.
This PR moves gradient_clip_val and gradient_clip_algorithm from the Trainer config into the model itself by modifying BaseModel in _base_model_v2.py.
Changes made:

_base_model_v2.py: Added gradient_clip_val and gradient_clip_algorithm parameters to _init_, stored as self.gradient_clip_val and self.gradient_clip_algorithm. Overrode configure_gradient_clipping to prioritize the model's own values over the Trainer's, and added a UserWarning when a model marked sensitive_to_gradient_explosions=True has gradient_clip_val=None.
_tslib_base_model_v2.py: Updated _init_ to accept and forward the new parameters to super()._init_.
_timexer_v2.py, _samformer_v2.py, _tft_v2.py: Set gradient_clip_val=0.1 as the default and self.sensitive_to_gradient_explosions = True since these transformer-based models are prone to gradient explosions.

What should a reviewer concentrate their feedback on?

The priority logic in configure_gradient_clipping - model's gradient_clip_val takes precedence over Trainer's value. Is this the right behavior?
The default value of 0.1 for transformer models - is this a sensible default or should it be left to the user?
Whether the UserWarning for sensitive_to_gradient_explosions models is the right mechanism, or if a stricter approach is preferred.

Did you add any tests for the change?
Yes, added tests/test_gradient_clipping_v2.py with the following tests:

test_base_model_gradient_clipping_params - verifies gradient_clip_val and gradient_clip_algorithm are correctly stored in the model.
test_configure_gradient_clipping_priority - verifies model's value takes priority over Trainer's value when configure_gradient_clipping is called.
test_sensitive_models_defaults - verifies TimeXer, Samformer, and TFT all have gradient_clip_val=0.1 and sensitive_to_gradient_explosions=True by default.
test_sensitive_model_warning - verifies a UserWarning is raised when a sensitive model has gradient_clip_val=None.

Any other comments?
This fix ensures that gradient clipping config travels with the model checkpoint automatically, eliminating the silent divergence bug that occurred when resuming training from a checkpoint without manually re-specifying gradient_clip_val in the Trainer.
PR checklist

 The PR title starts with [ENH]
 Added/modified tests

### DIFF
--- a/.github/workflows/pypi_release.yml
+++ b/.github/workflows/pypi_release.yml
@@ -48,7 +48,7 @@ jobs:
           python -m build --wheel --sdist --outdir wheelhouse
 
       - name: Store wheels
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: wheels
           path: wheelhouse/*
@@ -102,7 +102,7 @@ jobs:
       id-token: write
 
     steps:
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@v8
         with:
           name: wheels
           path: wheelhouse
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -195,7 +195,7 @@ jobs:
           coverage xml
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v5
+        uses: codecov/codecov-action@v6
         if: always()
         continue-on-error: true
         with:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,16 @@
 # Contributing guide
 
-You can find our contributing guide on our [website](https://www.sktime.net/en/latest/get_involved/contributing.html).
+You can find our general contributing guide on our [website](https://www.sktime.net/en/latest/get_involved/contributing.html).
+
+## Implementing New Models (v2)
+
+For contributors looking to implement new models under the v2 architecture, we provide a consolidated template:
+
+- **V2 Model Template**: [pytorch_forecasting/models/v2_template.py](file:///Users/sujanyd/Desktop/pytorch-forecasting/pytorch_forecasting/models/v2_template.py)
+
+This file contains both the `BaseModel` and `Base_pkg` subclasses needed to register and test your model. Follow the `CONTRIBUTOR TODO` comments in the file. Key methods include:
+
+Follow the `CONTRIBUTOR TODO` comments in these files to implement your model. Key methods to implement/override include:
+- `forward`: Primary model logic.
+- `training_step`: Optional, for custom training logic (e.g., teacher forcing).
+- `get_test_train_params`: In the package class, provides parameters for automated testing.
--- a/build_tools/changelog.py
+++ b/build_tools/changelog.py
@@ -193,6 +193,6 @@ if __name__ == "__main__":
     if diff["total_commits"] != len(pulls):
         raise ValueError(
             "Something went wrong and not all PR were fetched. "
-            f'There are {len(pulls)} PRs but {diff["total_commits"]} in the diff. '
+            f"There are {len(pulls)} PRs but {diff['total_commits']} in the diff. "
             "Please verify that all PRs are included in the changelog."
         )
--- a/docs/source/tutorials/stallion.ipynb
+++ b/docs/source/tutorials/stallion.ipynb
@@ -41,7 +41,6 @@
    "source": [
     "import warnings\n",
     "\n",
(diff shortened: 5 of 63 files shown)


StrikerEureka34:
Hi @sujanyd , thanks for contributing!
But I am already working on this, and we haven't received any inputs from the maintainers about the feasibility or requirements of this issue yet. So I feel raising a PR is a little premature at this stage.

In the meanwhile you can refer the roadmap and take other issues which are currently under discussion and not taken, thanks!

PR #2245, opened 2026-03-29
Fix TslibDataModule setup re-entrancy guards

### LINKED ISSUE #2218: TslibDataModule.setup() re-entrancy guard checks wrong attribute name, causing full dataset recreation on every call
**Describe the bug**

While reviewing `TslibDataModule`, I noticed that `setup()` contains a guard on line 723 that is meant to avoid re-creating the training and validation datasets if they have already been built. The intent is clear — this kind of guard is standard practice to prevent redundant, expensive work when `setup()` is called more than once (e.g., by a PyTorch Lightning `Trainer` internally).

However, the guard is checking the wrong attribute names and never actually triggers.

**Relevant code** (`_tslib_data_module.py`, line 722–725):

```python
if stage == "fit" or stage is None:
    if not hasattr(self, "_train_dataset") or not hasattr(self, "_val_dataset"):
        self._train_windows = self._create_windows(self._train_indices)
        ...
```

The attributes are stored on the instance as `self.train_dataset` and `self.val_dataset` (no leading underscore), but the guard checks for `self._train_dataset` and `self._val_dataset` (with a leading underscore). Because those private names are never assigned anywhere, `hasattr` always returns `False`, so the entire dataset setup — including `_create_windows()`, large Dataset construction, and random index shuffling — runs unconditionally on every `setup()` call.

**To Reproduce**

```python
dm.setup("fit")
dm.setup("fit")  # Should skip, but rebuilds everything from scratch

# Confirm the guard never fires:
print(hasattr(dm, "_train_dataset"))  # Always prints False
print(hasattr(dm, "train_dataset"))   # Prints True after setup
```

**Expected behavior**

Calling `setup("fit")` a second time should detect the already-existing datasets and return early, both for performance and for determinism. The current behavior also introduces a reproducibility issue: since `torch.randperm()` is called every time, repeated `setup()` invocations silently produce different train/val/test splits, which can cause confusing results during training.

**Proposed fix**

The fix is a single-line change — correct the attribute names in the `hasattr` guard:

```python
# Before (line 723, buggy):
if not hasattr(self, "_train_dataset") or not hasattr(self, "_val_dataset"):

# After (fixed):
if not hasattr(self, "train_dataset") or not hasattr(self, "val_dataset"):
```

**Location**

`pytorch_forecasting/data/_tslib_data_module.py`, line 723

Ajeem-git:
## Summary
- reuse the original randomized split indices when setup() is called repeatedly
- check the actual dataset attributes instead of nonexistent private names before rebuilding datasets
- add a regression test covering repeated setup(stage="fit") calls

## Testing
- python3.11 -m py_compile pytorch_forecasting/data/_tslib_data_module.py pytorch_forecasting/data/tests/test_tslib_data_module.py

Closes #2218

### DIFF
--- a/pytorch_forecasting/data/_tslib_data_module.py
+++ b/pytorch_forecasting/data/_tslib_data_module.py
@@ -697,30 +697,31 @@ class TslibDataModule(LightningDataModule):
 
         # this is a very rudimentary way to handle the splits when
         # the dataset is of size equal to 1 or 2.
-        self._indices = torch.randperm(total_series)
-        if total_series == 1:
-            self._train_indices = self._indices
-            self._val_indices = self._indices
-            self._test_indices = self._indices
-        elif total_series == 2:
-            self._train_indices = self._indices[0:1]
-            self._val_indices = self._indices[1:2]
-            self._test_indices = self._indices[1:2]
-        else:
-            self._train_size = int(self.train_val_test_split[0] * total_series)
-            self._val_size = int(self.train_val_test_split[1] * total_series)
+        if not hasattr(self, "_indices"):
+            self._indices = torch.randperm(total_series)
+            if total_series == 1:
+                self._train_indices = self._indices
+                self._val_indices = self._indices
+                self._test_indices = self._indices
+            elif total_series == 2:
+                self._train_indices = self._indices[0:1]
+                self._val_indices = self._indices[1:2]
+                self._test_indices = self._indices[1:2]
+            else:
+                self._train_size = int(self.train_val_test_split[0] * total_series)
+                self._val_size = int(self.train_val_test_split[1] * total_series)
 
-            self._train_indices = self._indices[: self._train_size]
-            self._val_indices = self._indices[
-                self._train_size : self._train_size + self._val_size
-            ]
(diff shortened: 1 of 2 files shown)


Meet-Ramjiyani-10:
Nice work ! 
Verified the attribute names are now correct (`train_dataset` vs `_train_dataset`).

**One question:** Should `stage="test"` and `stage="predict"` also have re-entrancy guards? Currently only `"fit"` is protected.

Otherwise LGTM.

Ajeem-git:
Good catch. test was already guarded with a check that only creates the dataset once, but predict was not consistent with that yet. I pushed a small follow-up to make predict re-entrant as well, and added a regression test for repeated setup(stage="predict") calls.

Meet-Ramjiyani-10:
Cool! The predict stage guard and regression test look correct. LGTM 

One small note for future consistency: the test stage still uses the old _test_dataset guard (if not hasattr(self, "_test_dataset"):). Since _test_dataset is never assigned, this guard doesn’t actually prevent re-creation. Might be worth a follow-up PR to align it with the fit/predict pattern. Not a blocker for this PR though.

PR #2254, opened 2026-04-07
[BUG] Ensure `CompositeMetric.__add__` and `__mul__` operators are immutable

Muhammad-Rebaal:
Fixes: #2253

## Description
The `__add__` and `__mul__` operators for `CompositeMetric` were previously modifying `self` in-place instead of returning newly instantiated objects. This violates typical python magic-method behavior and caused silent logical bugs where building multiple different composite sets out of a base metric modified the base itself permanently.

## Changes Made
- Updated `__add__` and `__mul__` in `_base_metrics.py` to copy the internal arrays and construct a fresh `CompositeMetric` instance off the copied data.

CC : @phoeenniixx , @PranavBhatP , @fkiraly


phoeenniixx (changes requested):
Thanks!
Can you please add a minimal test to make sure it is working as expected?

Muhammad-Rebaal:
Hi @phoeenniixx,
Kindly review this PR.

phoeenniixx (changes requested):
I think this is a design choice? that `a+b=a`?
This helps with memory as well. As we dont create a new instance rather change the existing one.
Is there any case where you would need `a` as well as `a+b`? if this is true, only then we should create a new instance, otherwise the existing implementation is more memory-friendly.

What are your thoughts on this?
Also pinging @agobbifbk and @fkiraly for their thoughts on this

fkiraly (approved):
I actually think that operation dunders should *never* modify the arguments.

I consider this such a grave mismatch with expectations that I would even say it is a bug.

Consider computing x = 3, then a = x + 4, and then suddenly x evaluates to 7 instead. Addition / multiplication dunders are expected to be non-mutating and "pure" functions.

fkiraly:
@phoeenniixx, what do you think?

phoeenniixx:
@fkiraly, I agree with this, I was just sceptical about the design choice here. It completely depends on the mental models of what `a+b` means. 
The addition being immutable is not an issue here ig, as `Metric` is not memory intensive that much...?

I will merge this then, i just wanted to know what you think about this.

phoeenniixx (approved):

PR #2256, opened 2026-04-07
[BUG] Safe fallback for retrieving index from `QuantileLoss.to_prediction` when `0.5` is unavailable

Muhammad-Rebaal:
References : Fixes #2255

## Description
If a user creates an instance of `QuantileLoss` defining bounding target quantiles (e.g., `[0.1, 0.9]`) but excluding the direct median probability `0.5`, performing a `.to_prediction()` pass crashes parsing the output.
## Changes Made
- Introduced a conditional flow handling. It first checks `if 0.5 in self.quantiles`, returning exact. If absent, it establishes the minimum offset matching `0.5`, safely reverting to the closest approximated center. Modified `pytorch_forecasting/metrics/quantile.py`.

CC : @phoeenniixx , @PranavBhatP , @fkiraly


phoeenniixx (changes requested):
Thanks!
Can you please add a minimal test to make sure it is working as expected?

Muhammad-Rebaal:
Hi @phoeenniixx ,
Can you please re-run the failed job `HTTP 504 Gateway Time-out` when trying to download the `stallion.parquet` dataset.

phoeenniixx (approved):
Nice!

PR #2257, opened 2026-04-14
[BUG] correct inverted condition for `n_plotting_samples` default in `DeepAR`

### LINKED ISSUE #2234: [BUG] DeepAR: Inverted condition causes n_plotting_samples to remain None when n_validation_samples is None
**Describe the bug**

In the DeepAR model initialization, there is a logic error where the condition for setting `n_plotting_samples` is inverted, causing it to remain `None` when `n_validation_samples` is also `None`.

The buggy code in `pytorch_forecasting/models/deepar/_deepar.py` lines 143-147:
```python
if n_plotting_samples is None:
    if n_validation_samples is None:
        n_plotting_samples = n_validation_samples  # BUG: assigns None to None!
    else:
        n_plotting_samples = 100
```

When both `n_plotting_samples` and `n_validation_samples` are `None`, the code assigns `n_validation_samples` (which is `None`) to `n_plotting_samples`, leaving it as `None` instead of assigning the default value of 100.

The condition `if n_validation_samples is None:` should be `if n_validation_samples is not None:`.

**To Reproduce**

```python
from pytorch_forecasting.models import DeepAR

# Create DeepAR without specifying n_validation_samples or n_plotting_samples
# (Both default to None in function signature)
model = DeepAR(cell_type="LSTM")

# Check the resulting n_plotting_samples value
print(f"n_plotting_samples: {model.hparams.n_plotting_samples}")
# Output: n_plotting_samples: None
# Expected: n_plotting_samples: 100

# This None value is later used in create_log():
# n_samples = [self.hparams.n_validation_samples, self.hparams.n_plotting_samples][self.training]
# When training=True, n_samples will be None, which can cause downstream errors
```

**Expected behavior**

When `n_validation_samples` is `None`, the code should default `n_plotting_samples` to 100. The corrected code should be:

```python
if n_plotting_samples is None:
    if n_validation_samples is not None:  # FIXED: inverted condition
        n_plotting_samples = n_validation_samples
    else:
        n_plotting_samples = 100
```

**Additional context**

The `n_plotting_samples` value is used in the `create_log` method (line 479) where it is passed as `n_samples` to prediction and quantiles functions. When `None`, this can cause issues in downstream calculations that expect an integer.

This bug affects all users who instantiate DeepAR without explicitly setting `n_validation_samples` or `n_plotting_samples`, which are both `None` by default in the function signature (line 65-66).

**Versions**
<details>

```
System:
    python: 3.12.0
executable: /Library/Frameworks/Python.framework/Versions/3.12/bin/python3
   machine: macOS-26.3.1-arm64-arm-64bit

Python dependencies:
          pip: 23.2.1
pytorch-forecasting: 1.6.1
        torch: 2.9.1
    lightning: 2.6.1
        numpy: 2.3.5
        scipy: 1.15.3
       pandas: 2.3.3
```

</details>

haoyu-haoyu:
#### Reference Issues/PRs

Fixes #2234. Fixes #2246.

The \`DeepAR\` constructor docstring says:

> \`n_plotting_samples : int, optional\` ... Defaults to \`n_validation_samples\` if not None or 100 otherwise.

but the implementation flipped the branches:

\`\`\`python
if n_plotting_samples is None:
    if n_validation_samples is None:
        n_plotting_samples = n_validation_samples   # assigns None back
    else:
        n_plotting_samples = 100                     # only fires when validation samples are set
\`\`\`

So when both arguments are left at their default \`None\`, \`n_plotting_samples\` stays \`None\`, and \`create_log()\` later passes \`None\` as the sample count to the prediction/quantiles utilities.

#### Changes

- Replace the inverted \`if/else\` with a single conditional expression that matches the docstring:
  \`n_plotting_samples = n_validation_samples if n_validation_samples is not None else 100\`
- Add a parametrized regression test (\`test_n_plotting_samples_default\`) that asserts the resolved value for the all-defaults case (100), and for two explicit \`n_validation_samples\` values (50 and 200).

- Whether the docstring's documented behavior is the intended contract — the fix aligns the code with the docstring rather than the other way round.

### DIFF
--- a/pytorch_forecasting/models/deepar/_deepar.py
+++ b/pytorch_forecasting/models/deepar/_deepar.py
@@ -141,10 +141,9 @@ class DeepAR(AutoRegressiveBaseModelWithCovariates):
         if logging_metrics is None:
             logging_metrics = nn.ModuleList([SMAPE(), MAE(), RMSE(), MAPE(), MASE()])
         if n_plotting_samples is None:
-            if n_validation_samples is None:
-                n_plotting_samples = n_validation_samples
-            else:
-                n_plotting_samples = 100
+            n_plotting_samples = (
+                n_validation_samples if n_validation_samples is not None else 100
+            )
         if static_categoricals is None:
             static_categoricals = []
         if static_reals is None:
--- a/tests/test_models/test_deepar.py
+++ b/tests/test_models/test_deepar.py
@@ -221,6 +221,30 @@ def test_predict_samples(model, dataloaders_with_covariates):
 
 
 @pytest.mark.parametrize(
+    "n_validation_samples,expected_n_plotting_samples",
+    [(None, 100), (50, 50), (200, 200)],
+)
+def test_n_plotting_samples_default(
+    dataloaders_with_covariates, n_validation_samples, expected_n_plotting_samples
+):
+    """Regression for #2234/#2246: when ``n_plotting_samples`` is left at its
+    default of ``None``, it should resolve to ``n_validation_samples`` if that
+    is given (matching the docstring) and fall back to ``100`` only when
+    ``n_validation_samples`` is also ``None``. The previous condition was
+    inverted and left ``n_plotting_samples`` as ``None`` in the all-defaults
+    case.
+    """
+    dataset = dataloaders_with_covariates["train"].dataset
+    model = DeepAR.from_dataset(
+        dataset,
+        hidden_size=5,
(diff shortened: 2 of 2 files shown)


phoeenniixx (approved):
Thanks @haoyu-haoyu !
But I would suggest to please carefully review whatever the agent is feeding you from future, the PR description is not rendered correctly because of that :)

PR #2262, opened 2026-04-16
[BUG] Fix tslib v2 static categorical/continuous feature mix-up

### LINKED ISSUE #2261: [BUG]: Static categorical/continuous features are not separated in v2 output
Describe the bug
In the tslib v2 data path, _TslibDataset.__getitem__ returned the same processed_data["static"] tensor for both static_categorical_features and static_continuous_features. This mixes feature types and breaks expected static-input semantics for models that rely on separated categorical vs continuous static tensors.

To Reproduce

Python
# Build a dataset with both:
# - one static categorical feature
# - one static continuous feature
# Then inspect a batch from tslib v2 dataloader.

batch = next(iter(dataloader))
x, y = batch

print(x["static_categorical_features"].shape)
print(x["static_continuous_features"].shape)
# before fix: both tensors came from the same source/static tensor
Expected behavior
static_categorical_features should contain only static categorical columns, and static_continuous_features should contain only static continuous columns, each with correct dimensions and no duplication/type leakage.

Additional context
Fix implemented in /home/runner/work/pytorch-forecasting/pytorch-forecasting/pytorch_forecasting/data/_tslib_data_module.py by splitting static features via metadata-defined indices.
Regression coverage added in /home/runner/work/pytorch-forecasting/pytorch-forecasting/tests/test_models/test_timexer_v2.py.

Versions
Repository: Chetansahney/pytorch-forecasting
Area: tslib v2 data module + Timexer v2 tests
Exact package/environment versions were not provided in this thread.

Chetansahney:
Reference Issues/PRs
Fixes #2261 

What does this implement/fix? Explain your changes.
-Fixes a bug in tslib v2 dataset output where both static_categorical_features and static_continuous_features were populated from the same static tensor.
-The change now splits static features using metadata indices so categorical and continuous static tensors are emitted separately and correctly.

What should a reviewer concentrate their feedback on?
-Correctness of static feature index mapping from metadata
-Tensor slicing behavior when one static type is missing/empty
-Backward compatibility of dataloader batch structure for tslib v2 consumers

Did you add any tests for the change?
Yes. Added/updated regression coverage in:
/home/runner/work/pytorch-forecasting/pytorch-forecasting/tests/test_models/test_timexer_v2.py
and validated with: python -m pytest tests/test_models/test_timexer_v2.py -q

Any other comments?
No.

PR checklist
- The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or    
   improving code, [DOC] - writing or improving documentation or docstrings.
- Added/modified tests
- Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with pre-commit install.  
To run hooks independent of commit, execute pre-commit run --all-files

### DIFF
--- a/pytorch_forecasting/data/_tslib_data_module.py
+++ b/pytorch_forecasting/data/_tslib_data_module.py
@@ -219,8 +219,37 @@ class _TslibDataset(Dataset):
             x["future_relative_time_idx"] = torch.arange(0, prediction_length)
 
         if processed_data["static"] is not None:
-            x["static_categorical_features"] = processed_data["static"].unsqueeze(0)
-            x["static_continuous_features"] = processed_data["static"].unsqueeze(0)
+            static_features = processed_data["static"]
+            if not isinstance(static_features, torch.Tensor):
+                static_features = torch.tensor(static_features, dtype=torch.float32)
+            static_features = static_features.flatten()
+
+            static_feature_names = metadata["feature_names"]["static"]
+            static_categorical_names = set(
+                metadata["feature_names"]["static_categorical"]
+            )
+
+            static_categorical_indices = [
+                idx
+                for idx, name in enumerate(static_feature_names)
+                if name in static_categorical_names
+            ]
+            static_continuous_indices = [
+                idx
+                for idx, name in enumerate(static_feature_names)
+                if name not in static_categorical_names
+            ]
+
+            x["static_categorical_features"] = (
+                static_features[static_categorical_indices].unsqueeze(0)
+                if len(static_categorical_indices) > 0
+                else torch.zeros((1, 0), dtype=static_features.dtype)
+            )
+            x["static_continuous_features"] = (
+                static_features[static_continuous_indices].unsqueeze(0)
+                if len(static_continuous_indices) > 0
+                else torch.zeros((1, 0), dtype=static_features.dtype)
+            )
 
         if "target_scale" in processed_data:
             x["target_scale"] = processed_data["target_scale"]
--- a/tests/test_models/test_timexer_v2.py
+++ b/tests/test_models/test_timexer_v2.py
@@ -440,3 +440,4 @@ def test_static_features_are_split_by_type_in_tslib_output():
     x, _ = next(iter(dm.train_dataloader()))
     assert x["static_categorical_features"].shape[-1] == 1
     assert x["static_continuous_features"].shape[-1] == 1
+    
--- a/tests/test_models/test_timexer_v2.py
+++ b/tests/test_models/test_timexer_v2.py
@@ -440,4 +440,3 @@ def test_static_features_are_split_by_type_in_tslib_output():
     x, _ = next(iter(dm.train_dataloader()))
     assert x["static_categorical_features"].shape[-1] == 1
     assert x["static_continuous_features"].shape[-1] == 1
-    
--- a/tests/test_models/test_timexer_v2.py
+++ b/tests/test_models/test_timexer_v2.py
@@ -394,3 +394,49 @@ def test_integration_with_datamodule(model, basic_tslib_data_module):
             assert test_output["prediction"].shape[1] == model.prediction_length
         except StopIteration:
             print("Test set is empty, skipping test testing")
+
+
+def test_static_features_are_split_by_type_in_tslib_output():
+    """Ensure static categorical and continuous tensors are separated in v2 output."""
+    np.random.seed(42)
+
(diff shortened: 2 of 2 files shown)


phoeenniixx (approved):
Thanks a lot! for catching this!

Also FYI @PranavBhatP @agobbifbk - please review if you feel any changes are required

pytorch_forecasting/data/_tslib_data_module.py
```diff
@@ -219,8 +219,37 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
             x["future_relative_time_idx"] = torch.arange(0, prediction_length)
 
         if processed_data["static"] is not None:
-            x["static_categorical_features"] = processed_data["static"].unsqueeze(0)
-            x["static_continuous_features"] = processed_data["static"].unsqueeze(0)
+            static_features = processed_data["static"]
+            if not isinstance(static_features, torch.Tensor):
+                static_features = torch.tensor(static_features, dtype=torch.float32)
+            static_features = static_features.flatten()
```
agobbifbk, line 225: Just a curiosity, which is the dimension of processed_data["static"]?

pytorch_forecasting/data/_tslib_data_module.py
```diff
@@ -219,8 +219,37 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
...
+            static_continuous_indices = [
+                idx
+                for idx, name in enumerate(static_feature_names)
+                if name not in static_categorical_names
+            ]
+
+            x["static_categorical_features"] = (
+                static_features[static_categorical_indices].unsqueeze(0)
+                if len(static_categorical_indices) > 0
+                else torch.zeros((1, 0), dtype=static_features.dtype)
```
agobbifbk, line 246: Are those zero values checked somewhere after in the code?

Chetansahney:
Hi @agobbifbk , thanks for the review, here are the answers to both of your questions.
1. So, processed_data["static"] comes from series["st"], which is created as:

torch.tensor(data[_static].iloc[0].to_numpy(copy=True) if _static else [])
So its shape is 1D: (n_static_features,) (or (0,) if no static features).

After that, .flatten() keeps it 1D, and later unsqueeze(0) turns selected parts into (1, n_selected_static_features).


2. Not as values
Those torch.zeros((1, 0), ...) tensors are mainly a shape placeholder (empty last dimension), and downstream code checks the dimension/size, e.g. in TFT v2:

static_cat.size(2) == 0 / static_cont.size(2) == 0 

So the logic branches on “do static features exist?” via shape, not on whether tensor values are zero. The zeros themselves are never semantically validated; they just keep batching/concatenation consistent.

Chetansahney:
I am looking forward for more reviews and feedbacks @agobbifbk @phoeenniixx

Chetansahney:
Hi @agobbifbk , hope the review is completed. Can you please provide me with the feedbacks? :)

agobbifbk:
Fine for me thx!

agobbifbk:
I have another thought: since this model can work with nans, and the current d2 layer discards samples with nan, maybe a flag is needed in the d2 allowing to create samples with nans. @phoeenniixx maybe you are aware if in the loss computation the target masking is supported....

phoeenniixx:
>  @phoeenniixx maybe you are aware if in the loss computation the target masking is supported....

Yes, I believe masking is present in metrics (see [here](https://github.com/sktime/pytorch-forecasting/blob/818a84d7cc293de01fd1ad046ce668a5d98d44a9/pytorch_forecasting/metrics/base_metrics/_base_metrics.py#L894))

agobbifbk:
So we can also open an issue for allowing the dataloader create samples with nans, what do you think? By default it is false, but we can turn it to true in case of ttm (and also chronos afaik).

phoeenniixx:
> So we can also open an issue for allowing the dataloader create samples with nans, what do you think? By default it is false, but we can turn it to true in case of ttm (and also chronos afaik).

I agree this is a good idea! Would you or @Chetansahney like to open the issue? Or I can if you want... We should keep this in high priority (for FMs) imo, what do you think @agobbifbk ?

Also, should we merge this PR? or do we need to make some changes in here?

FYI @PranavBhatP, @fkiraly

Chetansahney:
Hi @phoeenniixx , If you want I can go ahead opening the issue to add the dataloader to include sample NaN values, do LMK.

PR #2275, opened 2026-04-24
[BUG] LogNormalDistributionLoss validation error and enable integration tests

### LINKED ISSUE #1921: [BUG] `LogNormalDistributionLoss` behaves strange across different OS with `DecoderMLP`
**Describe the bug**

`LogNormalDistributionLoss` behaves strange across different OS with `DecoderMLP`. This specifically behaves strangely for `param-1` in `_decoder_mlp_pkg` used to test the model. For more info look at [this](https://github.com/sktime/pytorch-forecasting/pull/1908#issuecomment-3029226812) comment and the subsequent comments.

The loss-param fixture fails for different OS (like windows) but passes for OS like ubuntu. But this is also not always the case, when you make some change to `min_encoder_length` and `min_prediction_length`, it may pass for windows and fail for MacOS.

**To Reproduce**

Just remove the ` "tests:skip_by_name"` tag from the `DecoderMLP_pkg` and this will trigger the specific fixture that fails.

**Disclaimer**: If you test it locally on only one OS, it may pass as we've still not been able to diagnose the main issue and for which OS it fails for which param. This is quite random.

**Expected behavior**

The test should pass

**Additional context**

Look at [this](https://github.com/sktime/pytorch-forecasting/pull/1908#issuecomment-3029226812) comment and the subsequent comments.
and this discord thread: https://discord.com/channels/1075852648688930887/1080978709412069426/1391679541449068689

</details>

harshsomankar123-tech:
#### Reference Issues/PRs
Fix #1921 
This PR resolves a `ValueError` encountered when using `LogNormalDistributionLoss` with target tensors containing zero values and enables the previously skipped integration tests for `DecoderMLP`.

**Problem:**
The `LogNormal` distribution in PyTorch is defined only for strictly positive values ($x > 0$). When the target tensor contains zeros (which commonly occurs due to padding or specific data scenarios), calling `distribution.log_prob(y_actual)` triggers a validation error:
`ValueError: Expected value argument to be within the support (GreaterThan(lower_bound=0.0))...`

**Solution:**
* **Clamping in `LogNormalDistributionLoss`:** Overrode the `loss` method in `LogNormalDistributionLoss` to clamp `y_actual` to a minimum value of `1e-4` before calculating the log probability. This ensures the input is always within the distribution's support. The threshold of `1e-4` matches existing sanitization logic used elsewhere in the package (e.g., for `PackedSequence` test inputs).
* **Enabled Integration Tests:** With the root cause addressed, removed `test_integration[DecoderMLP-base_params-1-LogNormalDistributionLoss]` from the `tests:skip_by_name` list in `_decodermlp_pkg.py` to re-enable testing.

- [ ] Review the clamping logic and confirm the `1e-4` threshold is acceptable for general use cases.
- [ ] Verify that the un-skipped integration tests run properly in the CI pipeline.

- I enabled the previously skipped integration tests for `DecoderMLP` and `LogNormalDistributionLoss`.
- Verified the fix locally using a standalone reproduction script with zero-value targets. 
- Successfully ran the test suite: `pytest pytorch_forecasting/tests/test_all_estimators.py -k "DecoderMLP and LogNormalDistributionLoss"`.

#### Any other comments?
N/A

#### PR checklist

- [x] Added/modified tests


harshsomankar123-tech:
@phoeenniixx Could you please take a look when you have some time? 
Thanks!

pytorch_forecasting/models/mlp/_decodermlp_pkg.py
```diff
@@ -18,9 +18,7 @@ class DecoderMLP_pkg(_BasePtForecaster):
         "capability:flexible_history_length": True,
         "capability:cold_start": True,
         "python_dependencies": ["cpflows"],
-        "tests:skip_by_name": [
-            "test_integration[DecoderMLP-base_params-1-LogNormalDistributionLoss]"
-        ],
+        "tests:skip_by_name": [],
```
phoeenniixx: I think you can safely remove this tag, it wont lead to errors (and keep the tag directory clean)

harshsomankar123-tech: @phoeenniixx Thanks! PTAL

phoeenniixx (changes requested):
Thanks!
I have a few suggestions/questions:
- Can you please use numpydoc style for docstrings
- Do we need `1e-4` for `clamp`? We could use smaller values (like `1e-12` etc) if we just need to prevent division by zero? 
- Also, would it be better to expose this choice of clamp min value to the params of `loss`, so that the user can make their own choice depending on the data?

harshsomankar123-tech:
Thanks for the review @phoenniixx!

I’ve addressed all the suggested changes:

- Updated the docstrings to follow numpydoc style for consistency with the existing codebase.
- Yes, I agree it’s better to reduce the clamp value from 1e-4 to 1e-12 since we only need protection against log(0)/division edge cases.
- Exposed clamp_min as a configurable parameter in __init__() so users can tune it based on their dataset if needed.

PTAL when you have some time

phoeenniixx (approved):
Thanks a lot!

Issue #2277, opened 2026-05-04
[ENH] Improve Documentation for `pytorch-forecasting` (both v1 and v2)


phoeenniixx:
**Is your feature request related to a problem? Please describe.**
Currently the `pytorch-forecasting` documentation is cluttered and have no clear distinction between v1 and v2 APIs. Also, we still dont have references of new models that have been added, and the UX is also not good

**Describe the solution you'd like**
Improve the documentation-
### for v1

- [x] make the `models` table automated and add new models which are not present - #2236 

### for v2

- [x] Add some references for v2 to readme as well
- [x] Create a new header for v2 API
- [x] make the `models` table automated and add new models which are not present.
- [x] Add separate tutorials for v2 - should be easily distinguishable from v1
- [x] Add API references
- [x] Extension Templates
- [x] separate v2 models and datamodules from v1 
- [x] Clearly mention that `metrics` are still the same as v1

phoeenniixx:
FYI @agobbifbk, @fkiraly, @PranavBhatP

Soham-47:
hey @phoeenniixx , could I take this up?

phoeenniixx:
Hi @Soham-47 
Yes, ofcourse!
For now, I would suggest improving `metrics` table and generate it directly from the tags and `metric_pkg`, similar to what @IgnazioDS is doing in #2236

PranavBhatP:
Can we have something like `sktime-mcp` for ptf @phoeenniixx ?

phoeenniixx:
I think this is a good idea to have an mcp for `ptf`, but imo, we first need to get v2 and FM integration and make the API stable....
FYI @fkiraly @agobbifbk

phoeenniixx:
Done by #2279

PR #2279, opened 2026-05-10
[ENH] Add a separate documentation for `pytorch-forecasting` v1 and v2

phoeenniixx:
#### Reference Issues/PRs

Fixes parts of  #2277 (separation of v1 and v2)

This PR create separate header for v2 documentation and creates clear distinction between both the versions and adds clearer api references
### Note
I had to make some changes to the `ptf_V2_example` tutorial as it had all headings in H1 (using `#`) and this was leading to rendering issues where `readthedocs` was rendering all the H1 headings as the title of the notebook.

#### checklist for this PR

- [ ] Add some references for v2 to readme as well
- [x] Create a new header for v2 API
- [x] Add API references
- [x] separate v2 models and datamodules from v1 
- [x] Clearly mention that `metrics` are still the same as v1
- [x] make the `models` table automated and add new models which are not present. Waiting for #2236 to complete
- [x] Add separate tutorials for v2 - should be easily distinguishable from v1 
- [ ] Review v1 documentation and make any changes if needed


felipeangelimvieira:
One possibility: use gh-pages and dump documentation to specific directories for each version. Example:
https://github.com/felipeangelimvieira/prophetverse/tree/gh-pages

There is a html at index that redirects to the latest version. You can then add a custom domain.

phoeenniixx:
Thanks a lot for sharing this! I will look into it

felipeangelimvieira:
There's a CI that dumps the doc files when there's a release. You can also manually edit gh-pages branch.

fkiraly (changes requested):
I think this is really nice, content-wise!

My comments relate to thinking in "user journeys", namely, how a user would enter the page and then search for content.

Currently, the design is a bit cumbersome for reaching the "v2" pages, they are not clearly linked or advertised. The same issue is also present now in some v1 pages.

One way to deal with this is putting a "navigation element" at the top of the page. This does not need to be anything special, it can be a short explanatory paragraph or a box, e.g., on the models page

"we are currently reworking the API to a streamlined v2 interface. The v2 API is currently a beta, but already available in the package for testing. See more details here (link).

For v1 (current) model reference, see below.

For v2 (beta) model reference, follow this link."

Not necessarily exactly like this, this can probably be improved, but the idea I am trying to convey is that there should be an element which is:

* at the top, so easy to find
* signposting content and where to find what
* clickable to get the user there quickly.

The problem is that currently:

* links are often only at the very bottom of the content page, or even only in the side bar, so hard to find.
* context is not provided, so the user in their journey may encounter the navigation elements and not understand what the v1 vs v2 business is about

Generally, imagine (from multiple sensible starting points) how a user - especially, a new user - might navigate the documentation. Think about what they know and do not know yet, and consider if at any point they would be confused about what they see.

phoeenniixx:
Thanks for the review @fkiraly! 

I tried adding some of the info in models and data page, but I agree it is still not clear enough when I try it from an outsider's perspective.
I have some confusions about how to make it clear though:
- Should I advertise v1 and v2 at all places? (Models, Data, API etc)

     - what I mean by this is  should I :
          1) add complete information (everything about D1,D2,M,P layer) and warnings of v1 and v2 at all places (almost same content - which is entirety of API-v2 documentation in all the pages - models, data, API etc). 
          2) or should it be like this:

               -  Data page has just info about v1 and v2 data pipeline
               - Models page has just info about v1 and v2 models 
               - And so on

     - The problem with the second approach I feel is that sometimes it is not clear what to add and what not to add. If I just add info about lets say models to the Models page, it still need info about D2 layer and `pkg` layer to make the usage clear. But if I do this, then the models page becomes populated with info that might not be directly connected to models itself?
    - And first approach feels like duplication and a LOT OF USELESS info gets thrown at the user which they might already know because they have seen it in the previous page.
    - I think we need to create a layout of the webpage first - what to keep where. Without it, I believe it is getting overly complicated and cluttered.
- Should we have two separate headers like rn for API or should we follow the pattern of having a "parent" API page which redirects to v1 or v2 - this would make the header sec a bit less "cluttered" but you wont know that there are 2 versions of the API until you click on the API header.
- What to do about `pkg` layer of v1? Do we need to create a separate documenation for that as well? Because unlike v2, in v1, `pkg` layer is mainly just for test fixtures and nothing else. If yes, then it would go with `pkg` layer documentation of v2, in a similar way it is for models and data - there is a parent page which then links to `pkg` layer of v1 and/or v2.
- I am not sure how to do this (or if it is even feasible), but I think would it be better that we create a completely separate documentation for both versions. The landing page has just basic intro to the package and then links to diffrent pages for both the versions. The headers would also be reduced - Home, Getting started, Faq, installation (or maybe that is merged with Getting started), API versions.
     - Then here, we create a totally independent documentations for both the versions.
     - Unlike now, where everything is still overlapping
     - All the pages connected to the headers would have info about v1 and v2 with links to their MAIN API page. and from there you move to wherever you want, but in the same API ver you chose. It would be like a separate env for both of them.

My current Plan what I think should be added:
- Some clear indication on the landing page about v1 and v2
- We should merge "getting started" and installation headers into one - this will make the header sec less cluttered imo
- Add v1 and v2 indications at almost all places (or not?). I am a little confused, see above.
- Add separate tutorials page

Please let me know what you think here?

PranavBhatP:
@phoeenniixx the content looks great! Could you maybe render these docs and share the html files (as a a zip file mayb)? Imo it would be easier to review the structure of the docs in that way.

phoeenniixx:
image
you can look at the rendering by clicking on "view details"

phoeenniixx:
I think this is almost ready - just need to review v1 documentation

fkiraly:
@phoeenniixx, replying to your questions with my opinions:

> - Should I advertise v1 and v2 at all places? (Models, Data, API etc)

I would make this dependent on the likely user journey of people entering the page at the most likely entry points, e.g., webpage and readme.
A typical user journey will look like start page -> link or tab, so the common "second pages" are the pages that can be reached via the tab at the top. These pages are also likely google search results that make it the primary entry point.
Based on that, I would advertise the existence of v1 and v2 at the top of these pages - but not "everywhere". More precisely, in the form of:

* ensure context and signposting properly on every "second page", that is:
* the page is primarily v1
* at the top, it has a box that links to v2, and clearly says that this is in beta, to try out and available in the current package.
* I would already prepare the v2 pages as they would be in the v2 documentation, and just link there. At the top, add a similar box that tells people this is the 

Regarding some issues you correctly recognize and how to deal with them:

> The problem with the second approach I feel is that sometimes it is not clear what to add and what not to add. If I just add info about lets say models to the Models page, it still need info about D2 layer and `pkg` layer to make the usage clear. But if I do this, then the models page becomes populated with info that might not be directly connected to models itself?

Same approach as with modularizing code - anticipate questions that a user may have - but you do not necessarily need to answer them on the page itself. You can have one page related to the general design, and link there from every "v2 page" that explains the API.

> And first approach feels like duplication and a LOT OF USELESS info gets thrown at the user which they might already know because they have seen it in the previous page.

Solution to this is to keep the notice very short. 3 rows of text max, no grey wall.

> I think we need to create a layout of the webpage first - what to keep where. Without it, I believe it is getting overly complicated and cluttered.

v1 is already too cluttered, too many grey walls of text. A redesign is necessary, but might be out of scope for this PR.

> Should we have two separate headers like rn for API or should we follow the pattern of having a "parent" API page which redirects to v1 or v2 - this would make the header sec a bit less "cluttered" but you wont know that there are 2 versions of the API until you click on the API header.

I think too cluttered, and it adds a new node - one more click for the standard user journeys. Instead, I would show v1 by default but offer v2 as an optional (one more click) alternative with top of the page signposting to the content of current page and the v2 alternative.

> What to do about `pkg` layer of v1? Do we need to create a separate documenation for that as well? Because unlike v2, in v1, `pkg` layer is mainly just for test fixtures and nothing else. If yes, then it would go with `pkg` layer documentation of v2, in a similar way it is for models and data - there is a parent page which then links to `pkg` layer of v1 and/or v2.

It exists in v1 and v2, and it is a new addition. I think we may have to create a new page to explain it - but whether it needs to be top level or a new page, I am not sure. It could work well on the same page as the estimator overview - which could explain the tags/package system, together with the table of estimators obtained from the tags which are now a packaging condcern.

> My current Plan what I think should be added:
> - Some clear indication on the landing page about v1 and v2

Yes, good idea.

> - We should merge "getting started" and installation headers into one - this will make the header sec less cluttered imo

Hm, I would deduplicate but not merge. Getting started should be about the python code exclusively, and link to the installation page as a precondition. As discussed last week, I would suggest to have a minimal installation (e.g., a few lines including pip install) at the top, and link to the "installation" page for more details.

> - Add v1 and v2 indications at almost all places (or not?). I am a little confused, see above.

my suggestions above

- Add separate tutorials page

Yes, good idea. Currently, the tutorials are not properly displayed and the page looks a bit low quality without signposting or context.

fkiraly (changes requested):
Overall, I think this is very nice! Close to how I would have written it.

Some smaller comments:
* boxes at the top
    * I would suggest to think about how to shorten it even further. E.g., just provide the links with self-contained text, and employ telegraph style, instead of writing long English sentences.
    * At the moment, it is not correct to call the v1 API "legacy". Legacy means it is already outdated, while currently v2 is in beta, so v1 cannot be officially outdated.
    * the header text says "important" and "warning". "warning" is appropriate on v2, but "important" feels a bit odd on v1.
 * models v2 content: I would suggest to present code for two alternative layers of usage. One directly at the `lightning` model (which some users may prefer), and another with the `sklearn`-like fit/predict usage.
 * API v2: I would avoid "adoption of sklearn-like API", instead I would suggest to use "unified API with exchangeable models" or similar. Because "sklearn-like API" implies that the v1 `lightning` based API can no longer be used, but that is not true, you can interact on the lower level with the models too.
 * "package" - this is a problematic header, since it is not self-explanatory. The user will think it is about the package itself. I also think it belongs on the "models" page, which should explain that models can be accessed at two different layers. "raw" `lightning` model and a packaging layer with tags, API introspection, etc. This is perhaps hard to write and design. Optionally, a separate page for the package layer - accessible only from the "models" page - might be an option here. However, the "list of models" layer relies on tags.
     * one option to address this might be to cut the two pages differently: model API and model list. Model list has the list of available models with (some) tags, and at the top links to the model API page on usage. And the model API page link back to list of models. The separation would be "how to use the models" and "what models are there", where the former explains the separation between the `lightning` and the package layer.

docs/source/models_v2.rst
```diff
@@ -0,0 +1,133 @@
...
+    # 2. D2 Layer Configuration
+    datamodule_cfg = dict(
+        max_encoder_length=30,
+        max_prediction_length=1,
+        batch_size=32,
+        categorical_encoders={
+            "category": NaNLabelEncoder(add_nan=True),
+            "static_feature_cat": NaNLabelEncoder(add_nan=True),
+        },
+        scalers={
```
fkiraly: afaik this is not yet working code, since the scalers PR is not yet merged?

May be worth keeping in a separate PR as a desired end state, but rn I would only merge what already actually runs. Perhaps even add a test that just runs the code in the documentation exactly as-is, to ensure all code in the documentation is valid.

docs/source/models_v2.rst
```diff
@@ -0,0 +1,133 @@
...
+    model_pkg.fit(dataset)
+
+    # Generate predictions (You can also pass a DataModule or Dataloader here)
+    preds = model_pkg.predict(dataset, return_info=["index", "x", "y"])
+
+
+Implementing new architectures
+-------------------------------
+
+Please see the Extension Templates to understand the process and design of the implementations.
```
fkiraly: if you mention the extension templates, link them, or it will not be useful.
Also compare https://www.sktime.net/en/stable/developer_guide/add_estimators.html

docs/source/models_v2.rst
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    # Fit the model (You can pass the D1 dataset or a D2 DataModule here)
+    model_pkg.fit(dataset)
+
+    # Generate predictions (You can also pass a DataModule or Dataloader here)
+    preds = model_pkg.predict(dataset, return_info=["index", "x", "y"])
+
+
+Implementing new architectures
```
fkiraly: Suggestion: separate user guide more clearly from developer guide. Adding developer instructions can be very confusing to users.

A good solution is to have a separate area of the docs that is a "developer guide" and collects instructions for developers only. Whereas most of the remaining docs is for users (with occasional links to the developer guide).

There are some sensible exceptions to this separation, e.g., the install docs, where all types of install (stable from pip, latest from GitHub, and developer/editable install with dev dependencies) are listed.

docs/source/models_v2.rst
phoeenniixx: Passing `scalers` would still work, but it wont do anything, as the module takes in the scalers but it doesnot implement them. So, the code would not throw any error (or fail), it would just take in the scalers, but in its current implementation would not actually scale the data.

I added this as once the scalers PR is merged (which would be done shortly hopefully), we wouldnt have to update this part of code. In the current state, it would not scale the data, in future it would scale the data.

Its like the outer sockets in a building are already fitted in the wall, but we still need to connect those sockets to the main power supply. Plugging into the sockets would not do anything, but once the main power supply is connected, the appliance would start working automatically, we wont have to fit in the sockets after we have the main power supply.

what do you think?

docs/source/models_v2.rst
phoeenniixx: Yes, I would attach the link once #2297 and #2285 are merged and we have a complete folder structure of the extension-templates. I think these PRs would be merged before this PR for documentation would be merged, so update would be easy

phoeenniixx:
I think the design is doable now. We still need to add more content (like more explaination, examples, developer's guide etc) but I think that can be done later. Now that we have a clear separate docs, we should move forward with the feature parity?

But I still need to add links to extension-templates, but that would happen once both https://github.com/sktime/pytorch-forecasting/pull/2297 and https://github.com/sktime/pytorch-forecasting/pull/2285 gets merged

fkiraly (changes requested):
Looks great!

Small requests:
* "try the new API v2" does not advertise well enough that this is not released yet. "Try the pre-release v2 API". Then "An improved v2 API is in development. Try it out before release:"
* similarly, on other pages, it needs to become clear that this is not released yet. Try to signpost this clearly, without adding too much text.
* P layer does more than testing framework integration. I would add metadata, tags, and dependency management.

fkiraly (approved):
Perfect

PR #2284, opened 2026-05-18
 [ENH] M mode test for TimeXer v2

andersendsa:
#### Reference Issues/PRs

Resolves a codebase TODO.

 Added an M mode (multiple series) test to `test_timexer_v2.py` to resolve the actionable TODO comment requesting tests for multiple series forecasting now that the feature is successfully implemented.Improves test coverage and validates that the TimeXer v2 model correctly processes multi-target datasets.

* Verify that the data format in the test aligns with the expected M mode usage.

Yes, added `test_m_mode` to `tests/test_models/test_timexer_v2.py`.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/docs/source/tutorials/stallion.ipynb
+++ b/docs/source/tutorials/stallion.ipynb
@@ -41,7 +41,6 @@
    "source": [
     "import warnings\n",
     "\n",
-    "\n",
     "warnings.filterwarnings(\"ignore\")  # avoid printing out absolute paths"
    ]
   },
--- a/docs/source/tutorials/stallion.ipynb
+++ b/docs/source/tutorials/stallion.ipynb
@@ -27,7 +27,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "collapsed": false,
     "jupyter": {
@@ -40,6 +40,7 @@
    "outputs": [],
    "source": [
     "import warnings\n",
+    "\n",
     "\n",
     "warnings.filterwarnings(\"ignore\")  # avoid printing out absolute paths"
    ]
--- a/docs/source/tutorials/stallion.ipynb
+++ b/docs/source/tutorials/stallion.ipynb
@@ -27,7 +27,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false,
     "jupyter": {
@@ -40,6 +40,7 @@
    "outputs": [],
    "source": [
     "import warnings\n",
+    "\n",
     "\n",
     "warnings.filterwarnings(\"ignore\")  # avoid printing out absolute paths"
    ]
--- a/tests/test_models/test_timexer_v2.py
+++ b/tests/test_models/test_timexer_v2.py
@@ -394,29 +394,24 @@ def test_integration_with_datamodule(model, basic_tslib_data_module):
         except StopIteration:
             print("Test set is empty, skipping test testing")
 
-    def test_m_mode(sample_multivariate_multi_series_data):
-        """Test M mode (multiple series) with TimeXer model."""
-        df = sample_multivariate_multi_series_data
-        dataset = TimeSeries(
-            data=df,
-            time="time_idx",
-            target=["target1", "target2"],  # M mode uses multiple targets
-            group=["group_id"],
-            num=["temperature", "humidity", "pressure", "wind_speed"],
-            known=["temperature", "humidity", "pressure", "wind_speed", "time_idx"],
-        )
 
-        data_module = TslibDataModule(
-            time_series_dataset=dataset,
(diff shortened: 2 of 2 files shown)


andersendsa:
Hi @phoeenniixx  this pr passes all the tests and is ready for review

docs/source/tutorials/stallion.ipynb
phoeenniixx, line 1: Please revert this change

tests/test_models/test_timexer_v2.py
```diff
@@ -394,3 +393,46 @@ def test_integration_with_datamodule(model, basic_tslib_data_module):
...
+        data_module = TslibDataModule(
+            time_series_dataset=dataset,
+            batch_size=2,
+            context_length=12,
+            prediction_length=8,
+            train_val_test_split=(0.7, 0.15, 0.15),
+        )
+        data_module.setup()
+        metadata = data_module.metadata
+
```
phoeenniixx: why are we not using `pkg` class here? I think we could use that and then we dont need to create these classes or anything and everything would happen under the hood, we just need to pass configs?

phoeenniixx (changes requested):
Thanks!
I have added some suggestions, please have a look at them

andersendsa:
Hi @phoeenniixx  i have done the required changes and the pr is ready for review

tests/test_models/test_timexer_v2.py
```diff
@@ -394,3 +393,56 @@ def test_integration_with_datamodule(model, basic_tslib_data_module):
...
+    )
+
+    # Build datamodule and model under the hood to test forward pass
+    pkg.datamodule = pkg._build_datamodule(dataset)
+    pkg.datamodule.setup(stage="fit")
+
+    pkg._build_model(metadata=pkg.datamodule.metadata, **pkg.model_cfg)
+
+    train_dataloader = pkg.datamodule.train_dataloader()
+    batch = next(iter(train_dataloader))[0]
```
phoeenniixx, line 441: I think if you call `pkg.fit()`, `pkg.predict()`, you wouldnt have to create the data module etc separately. Or is this intentional?

andersendsa: > I think if you call `pkg.fit()`, `pkg.predict()`, you wouldnt have to create the data module etc separately. Or is this intentional?

Yes, avoiding pkg.fit() is intentional. For multiple targets (M mode), the dataloader returns targets as a list ([target1, target2]), but the underlying metric layer incorrectly unpacks lists as (target, weight), causing pkg.fit() to crash. By manually triggering the forward pass instead, we bypass this metric bug and successfully verify the model architecture handles M mode correctly.

phoeenniixx: This should be raised as an issue then! we should have a look at it. Can you please raise an issue and add your findings there? Please make sure to add some code to reproduce the error, so that we can diagnose it

phoeenniixx:
Can you please explain how we are handling M tests without the pipeline able to handle the `multiloss` and multi-target completelty

andersendsa:
> Can you please explain how we are handling M tests without the pipeline able to handle the `multiloss` and multi-target completelty

Right now, the PyTorch Forecasting v2 training pipeline (i.e., pkg.fit()) crashes on M mode datasets because the metrics layer (specifically LightningMetric.update which MultiLoss uses) isn't designed to handle target being a list of multiple target tensors. It incorrectly tries to unpack [target1, target2] as (target, weight), which causes a shape/dimension mismatch.

Because we cannot train the model end-to-end with multiple targets using the built-in pkg.fit() pipeline, we have to bypass the metrics layer to test if the M mode architecture works. Here is how we are handling the M mode tests without full pipeline support:

- Instantiate components manually (under the hood): We use TimeXer_pkg_v2 to hold the configurations, and instead of calling pkg.fit(), we explicitly call pkg._build_datamodule(dataset) and pkg._build_model(...). This gives us a fully initialized dataloader and a fully instantiated PyTorch model architecture.

- Fetch a batch directly: We pull a single batch directly from the dataloader (batch = next(iter(train_dataloader))[0]).

- Execute a forward pass without loss calculation: We put the model in evaluation mode (pkg.model.eval()) and execute the forward pass manually using output = pkg.model(batch) inside a torch.no_grad() context.

- Verify the output shape: Because we bypassed the metric/loss computation completely, the model successfully produces the raw prediction tensors. We then assert that output["prediction"].shape == (batch_size, prediction_length, num_targets).

Summary: We are handling M mode testing by checking the model's core architecture and forward pass in isolation, bypassing the high-level Lightning metric/loss calculations that are currently bugged/unsupported for multi-target datasets. This proves the internal layers of TimeXer scale correctly to multiple targets, even though the surrounding training wrapper does not yet support it..

andersendsa:
Hi @phoeenniixx  are there any changes required in this pr or is this pr good to merge ?

Issue #2004, opened 2025-12-04
[ENH] Add Autoformer model


Satarupa22-SD:
Referencing #1992 
The goal of this task is to assess the viability of integrating Autoformer into the current V2 interface and, if feasible, proceed with implementing the integration within the system.

CC: @fkiraly

SrijanShovit:
Hey @fkiraly  @phoeenniixx Can I take this up?

Satarupa22-SD:
@SrijanShovit I am currently working on this :)

SrijanShovit:
> [@SrijanShovit](https://github.com/SrijanShovit) I am currently working on this :)

ok you continue. Let me know in case you need any help.

gkbishnoi07:
is this issue still open?? can i work on this issue @fkiraly

Satarupa22-SD:
Hi @gkbishnoi07 , I am currently working on this issue as a part of the sktime mentorship program. If you are interested in models, perhaps you can refer to #1992. Thanks!
CC: @fkiraly

harshsomankar123-tech:
@phoeenniixx @fkiraly , I'd like to work on this issue if no one is currently working on it.
Plz assign me!

PR #2285, opened 2026-05-21
[ENH] Add extension templates for v1 neural networks

### LINKED ISSUE #1953: [ENH] extension templates for neural network classes

harshsomankar123-tech:
#### Reference Issues/PRs
Fixes #1953
I took reference from #2049.

This PR adds a standardized, folder-based extension template for v1 neural network models under `extension_templates/v1_network_template/`. It addresses the ambiguity regarding folder structure, metadata, and testing setup for new model contributors by providing a copy-pasteable blueprint.

Key changes:
* **`model.py`**: A boiler-plate Lightning model template covering the constructor, `_pkg()`, `from_dataset()`, and `forward()`. It includes comments on `self.hparams` read-only behavior and optional override placeholders (`to_prediction`, `to_quantiles`).
* **`_examplenetwork_pkg.py`**: A private package container exposing model metadata (`_tags`) with detailed commentary on their purpose/values, and required CI test fixtures (`get_base_test_params`, `_get_test_dataloaders_from`) to allow automatic discovery and execution of tests for any new model.
* **`README.md`**: A step-by-step user guide detailing how to copy, rename, and implement the template files when adding a new estimator, including details on mandatory test parameters.
* **`__init__.py`**: Basic initialization to package the directory cleanly.

* Verify that `extension_templates/v1_network_template/` is correctly positioned outside of the auto-registration directories so it does not interfere with standard model listing.
* Confirm the docstrings, instructions, and TODO markers inside the template files are clear and easy for external contributors to follow.

Not applicable (this adds template skeleton files and a guide, not a working model).

#### PR checklist
- [ ] Added/modified tests


extension_templates/v1_network_template/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,169 @@
...
+        - A good parameter set should primarily satisfy two criteria:
+          1. Low testing time: Chosen set of parameters should have a low testing time
+             (ideally in the magnitude of a few seconds for the entire test suite).
+             Avoid default values that result in "big" models which increase test time
+             or risk causing test worker runner crashes/timeouts.
+          2. Wide range of coverage: There should be a minimum of two parameter
+             sets with different sets of values to ensure a wide range of code path
+             coverage.
+          3. Avoid external dependencies in defaults: Do not require external
+             packages in the default parameter dictionary unless absolutely necessary.
```
phoeenniixx: Please add a point that one fixture should simply test the default, ie, the first test-param (or fixture) should be an empty `dict` (`{}`).

extension_templates/v1_network_template/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,169 @@
...
+            {"hidden_size": 8},
+            {"hidden_size": 16},
+        ]
+
+    @classmethod
+    def _get_test_dataloaders_from(cls, params):
+        """Return train and validation dataloaders for testing.
+
+        This is REQUIRED for v1 models in CI. The method should create
+        dataloaders from the test data scenarios provided by the framework.
```
phoeenniixx: Can you please add more info like what are the things it changes and why is it needed and what should a person exactly do in this method

Tell them about `make_datalaoders` and different `data_scenarios` and what to expect from each `data_scenario`.

extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,278 @@
...
+        ----------
+        hidden_size : int, default=16
+            An example hyperparameter. Replace with your model's actual
+            hyperparameters.
+        **kwargs
+            Additional keyword arguments passed to the parent
+            ``BaseModel.__init__`` class constructor.
+        """
+        # todo: add any custom model hyperparameters to the constructor signature above
+
```
phoeenniixx: also please add a pointer that we should collect all the `__init__` params here, and once collected, that params cannot be changed, you would need to create private args for that.
See how `sktime` explains it: https://github.com/sktime/sktime/blob/bd8fe08f118763a7296c85592acf900d397f6155/extension_templates/forecasting_simple.py#L134-L139

extension_templates/v1_network_template/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+A minimal neural network template that should:
+
+- Inherit from an appropriate v1 base class (e.g., `BaseModel` or a relevant subclass).
+- Define at least the following required methods:
+  - `__init__`
+  - `_pkg`
+  - `from_dataset`
+  - `forward`
+- Keep optional methods commented out (e.g., `to_prediction` or `to_quantiles` for probabilistic/quantile networks) and only implement them when custom post-processing, rescaling, or CDF calculations are needed.
```
phoeenniixx: I think we should remove them if we are not using them? To keep the final code clean?

extension_templates/v1_network_template/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+At minimum, `_tags` should include:
+
+- `info:name`
+- `info:pred_type`
+- `info:y_type`
+- `capability:exogenous`
+- `info:compute`
+- `authors`
+- `python_dependencies` (optional)
```
phoeenniixx: There are few more tags I assume? But unfortunately they are not consolidated in a list anywhere, you would need to look at different models to see what are the tags here. Like these tags:
https://github.com/sktime/pytorch-forecasting/blob/fd619aacda8acb4ea3bd65a787beef712d09a481/pytorch_forecasting/models/temporal_fusion_transformer/_tft_pkg.py#L16-L19

extension_templates/v1_network_template/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+- Create a valid model instance.
+- Exercise different configurations of the model.
+- Run quickly in CI.
+
+Example (illustrative only):
+
+```python
+return [
+    {"hidden_size": 8},
```
phoeenniixx: Please add an empty `dict` as well here, we should have one fixed fixture for all models, where we test the defaults of the models

phoeenniixx (changes requested):
Thanks! This look really good. Please have a look at some suggestions below.

Also, can you please add that the reusable `layers` of the models should go into `layers` folder? Maybe in the `README` and in the `model.py`? Please highlight this point as well.

extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,290 @@
...
+             ``self.hparams.hidden_size`` inside forward() or other methods.
+
+        3. Read-Only self.hparams Constraint:
+           - **CRITICAL**: The ``self.hparams`` object is read-only and should NEVER
+             be modified, overwritten, or mutated after initialization.
+           - If you need to handle default arguments, perform scaling/rescaling,
+             or derive values (e.g., setting up dimensions based on input
+             configurations), write them to private instance attributes (attributes
+             starting with a leading underscore), for example:
+             ``self._hidden_size = self.hparams.hidden_size``.
```
phoeenniixx: actually I meant something like this:
if `__init__` has 2 args: `param_a`, `param_b`, then we do something like this:

```python
def __init__(self, param_a, param_b)
     self.param_a = param_a
     self.param_b = param_b
     # if we need to make any changes to param_a, then
     # we do something like this
     self._param_a = somefn(param_a)
```
Once all the args of the `__init__` has been collected, they can never be changed, you need to use private params if you need to make any changes

harshsomankar123-tech: @phoeenniixx Thanks!

harshsomankar123-tech:
@phoeenniixx @PranavBhatP PTAL
Thanks!

extension_templates/v1_network_template/model.py
```diff
@@ -0,0 +1,299 @@
...
+             instance attributes prefixed with a leading underscore.
+           - For example, if ``__init__`` has arguments ``param_a`` and ``param_b``,
+             we capture them via ``self.save_hyperparameters()`` and then if we need
+             to modify or derive a value from ``param_a``:
+
+             .. code-block:: python
+
+                 # Once collected, self.hparams.param_a cannot be changed.
+                 # If you need to make changes, use private params:
+                 self._param_a = some_function(self.hparams.param_a)
```
phoeenniixx: why are we accessing this through `hparams`? Why not directly? is it necessary to save eveyrthing to `hparams`?

harshsomankar123-tech: @phoeenniixx Thanks! for Review
I think using `self.hparams` here is mainly to follow the standard PyTorch Lightning pattern with `self.save_hyperparameters()`. It helps with things like automatic checkpoint restoration and experiment logging.
Also, I don’t think everything necessarily needs to go into `hparams` we can always exclude values with `ignore=[]` if they aren’t true hyperparameters.
The reason I referenced `self.hparams.param_a` in the example was to show that once values are collected through `save_hyperparameters()`, they’re generally treated as immutable. So if we need to derive or modify something from them later, we should store that separately in a private attribute like `self._param_a`.

WDYT??

extension_templates/v1_network_template/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,199 @@
...
+
+        dwc = data_with_covariates()
+        dwc.assign(target=lambda x: x.volume)
+
+        dl_default_kwargs = dict(
+            target="target",
+            time_varying_known_reals=["price_actual"],
+            time_varying_unknown_reals=["target"],
+            static_categoricals=["agency"],
+            add_relative_time_idx=True,
```
phoeenniixx: I think adding a complete code could be misleading that this part of the code HAS to be here in all cases - which is not true. I think rather than a complete implementation, we should add pointers how to choose different data_scenarios and like for some specific `losses` we need to make some changes to teh data itself to make them complatible, I think that is what it should be here

extension_templates/v1_network_template/model.py
phoeenniixx: What i mean was there are some params which we dont save to `hparams` no? These `hparams` are just the params which are used everywhere. There are some params which are specific to a model. adding everything to `save_hyperparameters` is not right I think, like in some implementations we ignore `loss` etc.
Also, we dont need to access these params using `self.hparam`, you can directly access them by `self.parama` etc

harshsomankar123-tech:
@phoeenniixx @fkiraly PTAL 
Thanks!

phoeenniixx (changes requested):
I think it still lacks some clear indications (like which methods are required, etc) and some of the things are quite too much verbose I think. I dont think we need so much info about hyperparams.

You can also take a look at #2297 to see what I mean by this, that PR adds extension templates for v2

harshsomankar123-tech:
@phoeenniixx Thanks for the review!

I've aligned the v1 templates with the v2 style from #2297, added "implement this is mandatory" markers, reduced verbosity, and clearly listed mandatory and optional methods upfront. I also removed a large section of code to make the templates cleaner and easier to follow.

PTAL when you have some time.

extension_templates/v1/network/_model_pkg.py
```diff
@@ -0,0 +1,143 @@
...
+        "info:compute": 2,  # 1 = light, 3 = medium, 5 = heavy
+        "info:pred_type": ["point"],  # "point", "quantile", or "distr"
+        "info:y_type": ["numeric"],  # "numeric" or "category"
+        "authors": ["your-github-handle"],
+        "capability:exogenous": True,
+        "capability:multivariate": True,
+        "capability:pred_int": False,
+        "capability:flexible_history_length": True,
+        "capability:cold_start": False,
+        "python_dependencies": [],
```
phoeenniixx, line 84: I think that explaination about each tag was really good, and informative :)

extension_templates/v1_network_template/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,143 @@
...
+            PyTorch DataLoaders.
+        """
+        # todo: choose the appropriate data scenario for your model.
+        #
+        # Available scenarios from pytorch_forecasting.tests._data_scenarios:
+        #   data_with_covariates() + make_dataloaders() - general purpose
+        #   dataloaders_fixed_window_without_covariates() - no covariates
+        #   dataloaders_with_different_encoder_decoder_length() - var lengths
+        #   dataloaders_multi_target() - multivariate
+        #
```
phoeenniixx: The explaination of these methods were also really good

phoeenniixx (changes requested):
Brilliant!
I think it is almost complete, please look at some suggestions below, otherwise it is nice!
I would just suggest 2 more things:
- I dont think we need `__init__.py` in the folder as this is not a module and just a folder that has boiler plate code, that MUST BE COPIED and not imported
- Also, can you please rename the folder as just v1 and inside that folder, add model (or network) named folder that has the rest of the files

extension_templates/v1/network/_model_pkg.py
harshsomankar123-tech: @phoeenniixx Thanks!

phoeenniixx: I would suggest to add them back :)

extension_templates/v1/network/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,199 @@
...
+        dl_default_kwargs = dict(
+            target="target",
+            time_varying_known_reals=["price_actual"],
+            time_varying_unknown_reals=["target"],
+            static_categoricals=["agency"],
+            add_relative_time_idx=True,
+        )
+        dl_default_kwargs.update(data_loader_kwargs)
+        dataloaders = make_dataloaders(dwc, **dl_default_kwargs)
+        return dataloaders
```
phoeenniixx: we dont need this now i think?

harshsomankar123-tech: IMO, _get_test_dataloaders_from() is still strictly required for v1 models. The v1 test runner in test_all_estimators.py  line 384 calls it directly to obtain dataloaders for integration tests, unlike the v2 runner which handles datamodule instantiation automatically. Because of this, I think we should keep this method (and its default example implementation) in the v1 template to avoid breaking the existing v1 testing workflow. What do you think?

phoeenniixx: I meant the code inside this method, the code is not necessarily fixed and may change depending on the model.
Eg: see [`tft_pkg`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/temporal_fusion_transformer/_tft_pkg.py), it has different implementation as it needs to handle different losses.

phoeenniixx: we absouletly need this method, but the implementation is not fixed

extension_templates/v1/network/model.py
```diff
@@ -0,0 +1,152 @@
...
+Testing - required for pytorch-forecasting test framework:
+    Use the ``_pkg`` class for this. See _examplenetwork_pkg.py for more info.
+"""
+
+# todo: write an informative docstring for the file or module, remove the above
+
+import torch
+
+from pytorch_forecasting.data.timeseries import TimeSeriesDataSet
+from pytorch_forecasting.models.base import BaseModel
```
phoeenniixx: Also, i remember we have multiple `BaseModel`s, I think it would be great if we could add when to use which BaseModel here.

extension_templates/v1/network/model.py
```diff
@@ -0,0 +1,158 @@
...
+
+# todo: write an informative docstring for the file or module, remove the above
+
+import torch
+
+from pytorch_forecasting.data.timeseries import TimeSeriesDataSet
+
+# Choose the appropriate base class:
+# - Use ``BaseModel`` if the model does NOT use/support covariates (e.g., N-BEATS).
+# - Use ``BaseModelWithCovariates`` if the model supports static and/or time-varying
```
phoeenniixx, line 42: I think we have more base classes - like the autoregressive ones. Please have a look at the classes here: https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/base/_base_model.py

extension_templates/v1/network/model.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, hidden_size: int = 16, **kwargs):
+        # save the hparams
+        # you can ignore some params that are not true hyperparameters:
+        #   self.save_hyperparameters(ignore=["loss"])
+        self.save_hyperparameters()
+        super().__init__(**kwargs)
+
+        # IMPORTANT: the self.hparams should never be overwritten or mutated
+        # for handling defaults etc, write to other attributes, e.g.,
+        #   self._hidden_size = some_function(self.hparams.hidden_size)
```
phoeenniixx, line 83: Please make this 
```python
self._hidden_size = some_function(self.hidden_size)
```

extension_templates/v1/network/_examplenetwork_pkg.py
```diff
@@ -0,0 +1,201 @@
...
+Purpose of this implementation template:
+    quick implementation of new estimators following the template
+    NOT a concrete class to import! This is NOT a base class or concrete class!
+    This is to be used as a "fill-in" coding template.
+
+How to use this implementation template to implement a new estimator:
+- make a copy of the template in a suitable location, give it a descriptive name.
+    - the name of the file should be prefixed with an underscore and end with ``_pkg``.
+    For eg, for ExampleNetwork, the name of the file would be
+    _examplenetwork_pkg.py.
```
phoeenniixx: actually it should `_example_network_pkg.py`. Please separate two words with an underscore (`_`)

phoeenniixx (changes requested):
THanks!
Just a few nit-picks

extension_templates/v1/network/model.py
harshsomankar123-tech: @phoeenniixx Done PTAL
Thanks!

harshsomankar123-tech:
@phoeenniixx, could you please assign #2004 to me? I've already started working on it.
Thanks!

phoeenniixx:
Please match the file names as well. If its `model.py`. then it should be `model_pkg.py` and if it is `example_network.py` , then it should be `example_network_pkg.py` or similar

phoeenniixx (approved):
Nice! I think this is ready

PR #2286, opened 2026-05-21
[ENH] Move datamodules to a single folder

### LINKED ISSUE #2283: [ENH] Restructure folder structure to create clear distinctions between v1 and v2
Currently v1 and v2 code resides side by side and it is hard for the user to understand which part of the code is v1 and which is from the version 2.
Like:
- in [`data.timeseries`](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/data/timeseries) `TimeSeriesDataset` of v1 and D1 layer of v2 lies together.
- the Datamodules lie without any folder inside [`data`](https://github.com/sktime/pytorch-forecasting/tree/main/pytorch_forecasting/data) itself.

We should create clear distinctions and folders for different modules of v1 and v2. It would also make, imo, creating separate documentations (see #2277) for both the versions

**Suggested Designs**

1. Separation in between modules
    
    ```
    pytorch-forecasting
    |- data
         |- datamodules
                    |- tslib datamodule and encoder-decoder datamodule
         |- timeseries
                    |- TimeSeriesDataset (v1)
                    |- TimeSeries (V2 D1)
         |- Rest of the files (encoders, samplers etc)
         |- Tests
     |- models
            |- v1
                  |- v1 models (and pkg classes)
             |- v2
                  |- v2 models (and pkg classes)
      |- rest of the modules 
                 (they would not need any changes as they are almost the same for both the versions I think)
    ```

FYI @fkiraly @agobbifbk @PranavBhatP

phoeenniixx:
Fixes #2283 
This PR moves the datamodules to a single folder in `data/data_modules`. This would be helpful in separating datamodules from v1 parts as well as giving them their own desginated space in the package from where the users can easily import any datamodule without knowing the name of the file of the specific datamodule


fkiraly (changes requested):
I think this is a necessary step, but I wonder whether we can do something more downwards and upwards friendly, e.g.,

* call the new module `data_module` and not `data_modules` - similar to the previous file name
* export the modules on the level of `pytorch_forecasting.data`

phoeenniixx:
> * export the modules on the level of `pytorch_forecasting.data`

Do we want that? would that not conflict with v1? currently only v1 artifacts can be accessed using `pytorch_forecasting.data`, if we add v2 artifacts there as well, that may lead to confusion what part is v1 and what is not?

fkiraly (approved):
Happy with this - I think you are right about the confusion when exporting in too many places.

PR #2287, opened 2026-05-22
[ENH] Add missing tags in v2 models

phoeenniixx:
The v2 models dont have complete tags from the registry. This PR adds the missing tags


fkiraly (approved):
As discussed, lgtm!

PR #2292, opened 2026-05-23
[BUG] Fix NHiTS NameError when dataset has static reals but no time-varying encoder covariates

### LINKED ISSUE #2281: [BUG] NHiTS fails when TimeSeriesDataSet includes static real variables
## Summary
 
When using `NHiTS.from_dataset()` with a `TimeSeriesDataSet` that includes **static real variables**, the model fails during the forward pass with:

```
NameError: free variable 'encoder_features' referenced before assignment
```

This occurs whenever `add_target_scales=True` (or any other configuration that adds static reals).

```python
from pytorch_forecasting import TimeSeriesDataSet, NHiTS

data = generate_ar_data(seasonality=10.0, timesteps=400, n_series=100, seed=42)

dataset = TimeSeriesDataSet(
    data,
    time_idx="time_idx",
    target="value",
    group_ids=["series"],
    max_encoder_length=60,
    max_prediction_length=20,
    add_target_scales=True,   # <-- introduces static real variables
)

dataloader = dataset.to_dataloader(train=True, batch_size=8)
model = NHiTS.from_dataset(dataset)

batch, _ = next(iter(dataloader))
model(batch)  # crashes
```

 ## Observed Error
 
The failure occurs inside `NHiTS.forward()` when static variables are present:

```
x_s = torch.concat([encoder_features[name][:, 0] for name in self.static_variables], dim=1)
NameError: free variable 'encoder_features' referenced before assignment
```

This matches the traceback in the notebook.

 ## Root Cause Analysis
 
`self.static_size > 0` triggers the static‑feature branch in `forward()`.  
But `encoder_features` is never defined because the call to:

```python
encoder_features = self.extract_features(x, self.embeddings, period="encoder")
```

is only set `if self.encoder_covariate_size > 0`.

This happens either:

- (most likely) because `encoder_covariate_size()` does not include `len(self.hparams.static_reals)`, so the model incorrectly assumes there are no encoder covariates and skips feature extraction, or

- because `encoder_features` is not defined in the block `if self.static_size > 0`.

 ## Expected Behavior
 
Unless I am mistaken, N‑HiTS should support static real variables (as other models do), or at minimum should not crash.

 ## Proposed Fix
 
Either:

 1. Preferred:  
 Update `encoder_covariate_size()` to include static reals:
```python
return len(self.hparams.time_varying_encoder_reals) + len(self.hparams.static_reals)
        return len(
            set(self.hparams.time_varying_reals_encoder) - set(self.target_names)
        ) + sum(
            self.embeddings.output_size[name]
            for name in self.hparams.time_varying_categoricals_encoder
        ) + len(
            self.hparams.static_reals
        )
```

 2. Alternative (would this be correct?):  
 In `forward()`, define `encoder_features` when `self.static_size > 0`:
```python
        if self.static_size > 0:
            encoder_features = self.extract_features(
                x, self.embeddings, period="encoder"
            )
```

 ## Environment
 
pytorch-forecasting version: 1.7.0

PyTorch version: 2.10

Lightning version: 2.6.1

OS: Windows 11

Siraryansingh:
Closes #2281.

#### Summary

Fixes a `NameError` crash in `NHiTS.forward()` that occurs whenever a `TimeSeriesDataSet` introduces static real variables (e.g. `add_target_scales=True`) but has no time-varying encoder covariates.

#### Root Cause

`encoder_features` was only defined inside `if self.encoder_covariate_size > 0`, but the static-variable branch below it referenced `encoder_features` unconditionally. `encoder_covariate_size` correctly excludes static reals (they are not time-dependent), so the guard was `0` while `static_size > 0`, causing the crash.

#### Fix

Hoist the `extract_features("encoder")` call to fire whenever **either** encoder covariates **or** static variables are present. The time-varying `encoder_x_t` tensor is still built only when `encoder_covariate_size > 0`, so model weights and dimensions are unchanged.

#### Changes

<ul>
<li><code>pytorch_forecasting/models/nhits/_nhits.py</code> - one-line guard change + restructured encoder block</li>
<li><code>tests/test_models/test_nhits.py</code> - 3 regression tests covering forward-only, combined covariates, and full training loop</li>
<li><code>CHANGELOG.md</code> - entry added under v1.7.0 Fixes</li>
</ul>

####Minimal Reproducer

```python
from pytorch_forecasting import TimeSeriesDataSet, NHiTS

data = generate_ar_data(seasonality=10.0, timesteps=400, n_series=100, seed=42)
dataset = TimeSeriesDataSet(
    data, time_idx="time_idx", target="value", group_ids=["series"],
    max_encoder_length=60, max_prediction_length=20,
    add_target_scales=True,
)
model = NHiTS.from_dataset(dataset)
batch, _ = next(iter(dataset.to_dataloader(train=True, batch_size=8)))
model(batch)  # NameError before this fix
```


phoeenniixx (changes requested):
code  quality is failing, please use `pre-commit`.

Siraryansingh:
Thanks for the review @phoeenniixx! I've run pre-commit locally  ,all hooks pass now (ruff, ruff-format, trailing whitespace, etc.). The force-push in the second commit reverted the unintended formatting changes and keeps the diff minimal  only the actual bugfix in forward(). Ready for re-review.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Version release focusing on:
 
 ### Fixes
 
+* [BUG] Fix `NHiTS.forward()` `NameError` when dataset has static reals but no time-varying encoder covariates (#2281) @Siraryansingh
```
phoeenniixx: Please dont change the changelog, during releases, maintainers do it on their own

Siraryansingh:
Hi @phoeenniixx,

Thanks for the feedback, missed out on that one my bad.

I have reverted the changes to `CHANGELOG.md` completely, so the PR now only contains the actual bugfix in `_nhits.py` and the three regression tests in `test_nhits.py`. 

I've also verified that all local unit tests pass successfully and all `pre-commit` hooks (including `ruff` and `ruff-format`) are fully passing.

Siraryansingh:
Please take a look @phoeenniixx :)

PR #2294, opened 2026-05-25
[BUG] Ensure deterministic groups across different processes

### LINKED ISSUE #2293: [BUG] Using `hash`  in `TimeSeries` is inconsistent with `ddp/ddp_spawn`
**Describe the bug**

In the `TimeSeries` class https://github.com/sktime/pytorch-forecasting/blob/9476f456384b618532bd02ea470be62b16d52387/pytorch_forecasting/data/timeseries/_timeseries_v2.py#L253 

`hash()` on a str depends on `PYTHONHASHSEED` which is fixed per interpreter at startup. Under a `ddp` / `ddp_spawn` strategy, each rank is a separate, freshly spawned interpreter, so the same `group_id` hashes to a different integer on each rank. Models then train on inconsistent group identities across ranks, with no error raised.

**To Reproduce**

```python
from importlib.metadata import PackageNotFoundError, version
import json
import os
import subprocess
import sys

CHILD = r"""
import json
import pandas as pd
from pytorch_forecasting.data.timeseries import TimeSeries

rows = [
    {"gid": gid, "time": t, "target": float(t)}
    for gid in ["aa", "bb", "cc", "dd"]
    for t in range(3)
]
df = pd.DataFrame(rows)

ts = TimeSeries(data=df, time="time", target="target", group=["gid"])
mapping = [[str(gid), int(ts[i]["group"][0])] for i, gid in enumerate(ts._group_ids)]
print(json.dumps(mapping))
"""


def group_mapping(seed: str) -> dict[str, int]:
    """Build a TimeSeries in a fresh interpreter pinned to PYTHONHASHSEED=seed."""
    env = {**os.environ, "PYTHONHASHSEED": seed}
    proc = subprocess.run(
        [sys.executable, "-c", CHILD],
        check=True,
        capture_output=True,
        text=True,
        env=env,
    )
    last_line = proc.stdout.strip().splitlines()[-1]
    return dict(json.loads(last_line))


def _ver(dist: str) -> str:
    try:
        return version(dist)
    except PackageNotFoundError:
        return "unknown"


def run() -> int:
    import platform

    print("python              :", platform.python_version())
    print("platform            :", platform.platform())
    print("pytorch-forecasting :", _ver("pytorch-forecasting"))
    print("torch               :", _ver("torch"))
    print("pandas              :", _ver("pandas"))
    print()

    m1 = group_mapping("1")
    m2 = group_mapping("2")

    print("PYTHONHASHSEED=1 :", m1)
    print("PYTHONHASHSEED=2 :", m2)
    print()

    mismatches = {g: (m1[g], m2[g]) for g in m1 if m1[g] != m2[g]}
    if mismatches:
        print("BUG REPRODUCED: same group_id maps to a different `group` per seed")
        for g, (a, b) in mismatches.items():
            print(f"  group_id={g!r}: seed1={a}  seed2={b}")


if __name__ == "__main__":
    run()

```

**Expected behavior**

This will show that different hashes across different seeds for the same groups

**Additional context**

**Versions**
<details>

</details>

fnhirwa:
#### Reference Issues/PRs

fixes #2293 

Fix non-deterministic group ids across ddp ranks by using a stable contiguous index instead of `hash()`

Check the reproducibility code in the issue and run it against the current state of the PR.

Added the tests to ensure the determinism and contigous group ids

#### PR checklist

- [x] Added/modified tests


fnhirwa:
@phoeenniixx should I go ahead and merge?

pytorch_forecasting/data/tests/test_tslib_data_module.py
```diff
@@ -530,3 +530,23 @@ def test_multivariate_target():
     assert (
         y.shape[-1] == 2
     ), "Target should have two dimensions for n_features for multivariate target."
+
+
+def test_group_index():
```
phoeenniixx: I think this test should lie here: https://github.com/sktime/pytorch-forecasting/blob/main/tests/test_data/test_d1.py ?

phoeenniixx (changes requested):
Thanks! 
Sorry for late review, just a small suggestion. Thanks a lot for spotting this issue!

phoeenniixx (approved):
Thanks!

PR #2297, opened 2026-05-28
[ENH] Add extension templates for `pytorch-forecasting` v2

phoeenniixx:
This PR adds extension templates for `pytorch-forecasting` v2

Checklist for this PR:

- [x] Extension template for models and `model_pkg` class
- [x] Extension template for D2 layer
- [ ] Minimal Extension template for D1 layer
- [x] Minimal Extension template for `metrics`


phoeenniixx:
Do we need an extension template for D1 layer as well? I mean that would be simply just `__init__()` and `__getitem__()` no? most of the things happens in Datamodule anyway.
FYI @agobbifbk, @fkiraly, @PranavBhatP

fkiraly (changes requested):
Nice!

I think this may need an additional set of instruction for the case of implementing a model. I.e., that an implementer either needs to *find* a D2 layer class that fits, or create that, and a model class. This can be short.

Perhaps as a README like in the v1 case?

phoeenniixx:
> I think this may need an additional set of instruction for the case of implementing a model. I.e., that an implementer either needs to _find_ a D2 layer class that fits, or create that, and a model class. This can be short.
> 
> Perhaps as a README like in the v1 case?

I actually did add this in the `model_pkg.py` See here: https://github.com/phoeenniixx/pytorch-forecasting/blob/1790a8d43a028f766e3b453e28d1356e0a864f61/extension_templates/v2/model_simple/model_pkg.py#L107-L112

Do you think it is clear enough, or should we add a readme?

fkiraly:
I think that is too far in the body of the template. There should be an explanation at the start, or in a readme.md (in one of the folders).

phoeenniixx:
Done

fkiraly (approved):
Brilliant! The readme is exactly what I was thinking of.

Later we should link this from the developer guide.

PR #2300, opened 2026-06-01
[BUG] Ensure writable NumPy array before tensor indexing in `decoded_index`

### LINKED ISSUE #2299: [BUG] Non‑writable NumPy array used for tensor slicing triggers PyTorch warning
## Summary
 
When calling `.filter()` on a `TimeSeriesDataSet`, PyTorch emits a `UserWarning` indicating that a non‑writable NumPy array is being used for tensor indexing. This happens because the property `.decoded_index` internally slices a `torch.Tensor` using a read‑only NumPy array derived with `pandas.DataFrame.to_numpy()`.

The warning appears only once per Python process (due to PyTorch’s internal suppression), but it indicates a real underlying issue.

---

```python
import pandas as pd
pd.options.mode.copy_on_write = True

# Create a small synthetic dataset with noise

import numpy as np

data = generate_ar_data(seasonality=4, timesteps=8, n_series=2, seed=42)
data

# Define a TimeSeriesDataSet:
from sklearn.preprocessing import StandardScaler
from pytorch_forecasting import TimeSeriesDataSet, EncoderNormalizer 

# Define the dataset
self = TimeSeriesDataSet(
    data,
    time_idx="time_idx",
    target="value",
    group_ids=["series"],
    max_encoder_length=4,
    max_prediction_length=2,
)

self.decoded_index
```

---

 ## Warning message
 
```
UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program.
```

Running the first cell triggers the warning.  
Subsequent runs do not show it because PyTorch suppresses this warning after the first occurrence.

---

 ## Root cause
 
Inside `.filter()`, the `.decoded_index` property is accessed.  
At that point:

- a `pandas.DataFrame.to_numpy()` call produces a **read‑only** NumPy array

- this array is used to slice a `torch.Tensor`

- PyTorch detects the non‑writable array and emits the warning

---

 ## Expected behavior
 
No warning should be triggered.  
The NumPy array should be made writable before being used for tensor indexing.

---

 ## Proposed fix
 
Insert an explicit `copy=True` in `pandas.DataFrame.to_numpy()` to ensure the NumPy array is writable before slicing the tensor.

Example:

```python
index = df.index.to_numpy(copy=True)   # ensure writable
```

---

 ## Environment

- pytorch-forecasting: 1.7.0

- pytorch: 2.11.0+cu126

- pandas: 2.3.0

- python: 3.11

---

 ## Additional notes
 
This warning originates from PyTorch’s C++ backend and is suppressed after the first emission, so it may be easy to miss. The underlying issue is still present.

---

cngmid:
## Summary
This PR fixes #2299.

A warning raised by PyTorch when `.decoded_index` uses a non‑writable NumPy array (derived from `pandas.DataFrame.to_numpy()`) to index a `torch.Tensor`.

PyTorch emits the following warning on first occurrence:

> UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program.

Although the warning is suppressed after the first emission, it indicates a real underlying issue.

---

 ## Root cause
`pandas.DataFrame.to_numpy()` returns a **read‑only** NumPy array in this context.  
This array is then used for tensor slicing inside `.decoded_index`, which triggers PyTorch’s safety check.

---

 ## Fix
The fix is minimal and safe:

- explicitly copy the NumPy array before using it for tensor indexing  
- ensures the array is writable  
- avoids the PyTorch warning entirely

Example change:

```python
index = df.index.to_numpy(copy=True)   # ensure writable
```

---

 ## Why no test is included
 
PyTorch suppresses this warning after the first emission in a process, making it impractical to reliably detect in automated tests.  
The fix is straightforward and low‑risk.

---

 ## Impact
 
- eliminates the PyTorch warning

- avoids undefined behavior

- improves robustness when filtering datasets

- no API changes

- no performance impact

---

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries.py
@@ -1934,8 +1934,8 @@ class TimeSeriesDataSet(Dataset):
             pd.DataFrame: index that can be understood in terms of original data
         """
         # get dataframe to filter
-        index_start = self.index["index_start"].to_numpy()
-        index_last = self.index["index_end"].to_numpy()
+        index_start = self.index["index_start"].to_numpy(copy=True)
+        index_last = self.index["index_end"].to_numpy(copy=True)
         index = (
             # get group ids in order of index
             pd.DataFrame(


phoeenniixx (approved):
Thanks!
FYI @fkiraly @agobbifbk

PR #2302, opened 2026-06-03
&PranavBhatP [ENH] Add feature scaling and normalization to `EncoderDecoderDataModule`

### LINKED ISSUE #1979: [ENH] Add Support for Feature Scaling in D2 DataModule Layer.
The `EncoderDecoderTimeSeriesDataModule` (D2 layer) currently does not apply feature scaling (for continuous variables) as part of its preprocessing pipeline. 

The scalers can be implemented within the `_preprocess_data` method and called during `setup`. We fit the scalers on the training data and then transform each individual batch inside the `_ProcessedEncoderDecoderDataset`.

**Additional context**
#1974 - Work items for PTFv2.

phoeenniixx:
#### Reference Issues/PRs
fixes #1979, stacks on #1983

- Implements feature scaling for continuous targets and features in the D2 layer using input scalers and `target_normalizer` params inside the data module.

- Adds a new `ScalerAdapter` that handles type changes (`tensor` <-> `numpy`) and handles different scalers and normalizers.

- Adds tests to `test_data_module`. Most of the tests are same as #1983, just minor updates to a few tests
   - Removes `sklearn` scalers from the tests for `target_normalizer` as `target_normalizer` is mainly for `ptf` normalizers only and the `sklearn` scalers donot return info like scaling params for multiple gorups etc. I think we dont need to support those scalers for target_normalizer. But as we use a single Adapter for all scalers/normalizers, if the user DO PASS the `sklearn` scalers to `target_normalizer`, it would easily work and not cause any issues.
   - Removes the check for `_target_normalizer_fitted` for `EncoderNormalizer` as for this normalizer, it normalizes everything on the fly within `__getitem__`, so i dont think we should make this flag `True` at all for this normalizer to work correctly.
   
- Removes `_apply_encoder_normalizer_on_target` and  `_apply_encoder_normalizer_on_feat` and takes the type checking for multiple scalers to the adpater.
- Uses `MutliNormalizers` for multi-target now in place of `lists` 

#### Todo

- [x] Add "auto" `target_normalizer` selection
- [x] Add support for `GroupNormalizer`.
- [x] Add support for `NanLabelEncoder`.
- [x] Remove the changes to the `BasePkg` and make this PR mainly just about adding the scaling and removing the parts that save, load the things.


phoeenniixx:
FYI @agobbifbk, @fkiraly

agobbifbk:
THX @phoeenniixx! I have a question here:
```python

        self._train_indices = self._split_indices[: self._train_size]
        self._val_indices = self._split_indices[
            self._train_size : self._train_size + self._val_size
        ]
        self._test_indices = self._split_indices[self._train_size + self._val_size :]
```        
we suppose that `len(self.time_series_dataset)>1` at least here no? Most of the time you have a single dataframe (e.g. hourly energy consumption), did you tested all the API also in this edge case? 

If I have a single time series, there will be no validation and test set, meaning that: the scalers will be fitted on the whole timeseries,  and I can not evaluate if the model is learning the data or the patterns (there is no test set). I know that we decided to introduce temporal splits in a second moments, but since this will affect the scaling process I was wondering if we can write this part just once. Feel free to discard my comment, but somehow we need to remember this :-)

phoeenniixx:
> THX @phoeenniixx! I have a question here:
> 
> ```python
>         total_series = len(self.time_series_dataset)
>         self._split_indices = torch.randperm(total_series)
> 
>         self._train_size = int(self.train_val_test_split[0] * total_series)
>         self._val_size = int(self.train_val_test_split[1] * total_series)
> 
>         self._train_indices = self._split_indices[: self._train_size]
>         self._val_indices = self._split_indices[
>             self._train_size : self._train_size + self._val_size
>         ]
>         self._test_indices = self._split_indices[self._train_size + self._val_size :]
> ```
> 
> we suppose that `len(self.time_series_dataset)>1` at least here no? Most of the time you have a single dataframe (e.g. hourly energy consumption), did you tested all the API also in this edge case?
> 
> If I have a single time series, there will be no validation and test set, meaning that: the scalers will be fitted on the whole timeseries, and I can not evaluate if the model is learning the data or the patterns (there is no test set). I know that we decided to introduce temporal splits in a second moments, but since this will affect the scaling process I was wondering if we can write this part just once. Feel free to discard my comment, but somehow we need to remember this :-)

I agree, the splitting of data is an issue with the current pipeline and your concern is valid, on that edge case it would certainly cause issues, but to be honest, I have no other solution then adding the splitting. I think that should be the next thing we should work upon? @Muhammad-Rebaal is already working on this (see #2087) and he will be working on this full time under ESoC starting June 19, I think we could focus on this at that time? What do you think?

I think once we have the splitting mechanism, this would be easier imo

I think this PR would be followed by #2087 shortly as June 19 is near :)

tests/test_data/test_data_module.py
```diff
@@ -436,7 +439,18 @@ def test_different_train_val_test_split(sample_timeseries_data):
...
+@pytest.mark.parametrize(
+    "normalizer_list",
+    [
+        [TorchNormalizer(), EncoderNormalizer()],
+        [EncoderNormalizer(), GroupNormalizer()],
+        [GroupNormalizer(), NaNLabelEncoder(add_nan=True)],
+        [TorchNormalizer(), GroupNormalizer()],
+        [NaNLabelEncoder(add_nan=True), EncoderNormalizer()],
+        [EncoderNormalizer(), TorchNormalizer()],
+    ],
```
phoeenniixx, line 458: Clarification:
Here we are not using `sklearn` scalers, as `MultiNormalizer` doesnot support `sklearn` scalers. See here the type hint:
https://github.com/sktime/pytorch-forecasting/blob/7b00758766a1480e8e5269f3deb42066f6f30d47/pytorch_forecasting/data/encoders.py#L1360-L1366

Question: should the adapter handle it `sklearn` scalers separately or should we make changes to `MultiNormalizer` instead to support the `sklearn` scalers? I think second option is better (and potentially harder)?

phoeenniixx: https://github.com/sktime/pytorch-forecasting/blob/7b00758766a1480e8e5269f3deb42066f6f30d47/pytorch_forecasting/data/timeseries/_timeseries.py#L973-L974

v1 always move to `MultiNormalizer` when we are using multi-target, but apparently we can't pass `sklearn` scalers there, but for single target it works??

EDIT: I now remembered, in v1 you cant pass `sklearn` scalers as `target_normalizers`, so it doesnt even work for single target I think

phoeenniixx:
Hi @fkiraly @agobbifbk @PranavBhatP 
I think this PR is ready for the first round of reviews. 
I have tried to replicate v1 feats as much as I could understand

phoeenniixx:
On the side note:
I think we should take the private dataset class outside to keep the structure clean - will happen in a separate PR

phoeenniixx:
I think we should remove the saving and loading methods for data module (`get_scaler_state` adn `set_scaler_set`), as I think we still need to decide on the exact design for this

pytorch_forecasting/adapters/scaler_adapters.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    Unified array-in / tensor-out interface for single and multi-target scalers.
+
+    Accepts torch.Tensor, np.ndarray, or pd.Series as input. Output is always
+    a torch.Tensor. Conversions are performed only when the underlying scaler
+    requires a specific type:
+
+    Parameters
+    ----------
+    scaler : sklearn scaler, TorchNormalizer, EncoderNormalizer, NaNLabelEncoder,
```
fkiraly: better: document the exact locations these are coming from, and also assumed API contracts.

pytorch_forecasting/adapters/scaler_adapters.py
```diff
@@ -0,0 +1,219 @@
...
+             or MultiNormalizer
+    """
+
+    def __init__(self, scaler):
+        self._scaler = scaler
+        self.is_multi = isinstance(scaler, MultiNormalizer)
+        self._is_sklearn = isinstance(scaler, _SKLEARN_SCALERS)
```
fkiraly: it is better to use API contracts than checks for specific classes, the latter is quite brittle.
For "is_sklearn", I would instead use logic as in `sktime` `is_sklearn_transformer`, and simply assume that any transformer passed is a scaler

pytorch_forecasting/adapters/scaler_adapters.py
```diff
@@ -0,0 +1,219 @@
...
+        t = _to_tensor(data)
+        return t.squeeze(-1) if (t.ndim == 2 and t.shape[1] == 1) else t
+
+    def fit(self, data: ArrayLike, X: pd.DataFrame = None) -> "ScalerAdapter":
+        """Fit the scaler.
+
+        Parameters
+        ----------
+        data : tensor, ndarray, or Series
+            Shape ``(n_samples,)`` for single-target or
```
fkiraly, line 117: should we also think about what happens if we pass an (1 x m) tensor?

Or an (n x m) tensor in the single-target capable case?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -58,17 +65,24 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
...
+        :py:class`~pytorch_forecasting.data.encoders.MultiNormalizer`.
+        By default an appropriate normalizer is chosen automatically.
 
     categorical_encoders : Optional[Dict[str, NaNLabelEncoder]], default=None
         Dictionary of categorical encoders.
 
     scalers :
-    Optional[Dict[str, Union[StandardScaler, RobustScaler,
-                        TorchNormalizer, EncoderNormalizer]]], default=None
+    Optional[Dict[str, Union[RobustScaler, StandardScaler, MinMaxScaler, MaxAbsScaler,
```
fkiraly: extremely hard to understand for a user. Instead, combine plain English with a list of possible API assumptions, and within these examples of compatible examples.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -58,17 +65,24 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
         Whether to add target scaling information.
     add_encoder_length : Union[bool, str], default="auto"
         Whether to include encoder length information.
-    target_normalizer :
-        Union[NORMALIZER, str, List[NORMALIZER], Tuple[NORMALIZER], None],
-         default="auto"
-        Normalizer for the target variable. If "auto", uses `RobustScaler`.
+    target_normalizer : torch transformer, str, list, tuple, optional, default="auto"
```
fkiraly: it should be described what the default does.

I would also make `None` the default - this is as current, and avoids a dependency on eager logic that does not scale well (i.e., loading all data and calling a sklearn normalizer)

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -648,39 +972,52 @@ def setup(self, stage: str | None = None):
         self._test_indices = self._split_indices[self._train_size + self._val_size :]
 
         if stage is None or stage == "fit":
+            self._resolve_target_normalizer(self._train_indices)
+            if not self._target_normalizer_fitted:
+                print(self._target_normalizer_fitted, self._target_normalizer._scaler)
```
fkiraly: stray print statement?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -648,39 +972,52 @@ def setup(self, stage: str | None = None):
...
+                self._train_preprocessed = self._preprocess_all_data(
+                    self._train_indices
+                )
+                self._val_preprocessed = self._preprocess_all_data(self._val_indices)
+
                 self.train_windows = self._create_windows(self._train_indices)
                 self.val_windows = self._create_windows(self._val_indices)
 
                 self.train_dataset = self._ProcessedEncoderDecoderDataset(
-                    self.time_series_dataset,
```
fkiraly, line 656: why is there a change of arguments?

fkiraly (changes requested):
Very nice, I think the separation is now very clean!

* I think usage should be centered around *no preprocessing being done*. This allows maximum user flexibility and avoids coupling to `pandas` or eager logic.
    * Most importantly, we *must* avoid mandatory coupling to `pandas` and eager logic in the default.
* as a consequence of the above, I see the change in args of `_ProcessedEncoderDecoderDataset` very critically. I think we want to keep the `TimeSeries` input, and pre-processing should be `TimeSeries` to `TimeSeries`.
* for clearer structure, I would reconsider how `_preprocess_all_data` vs `_preprocess_data` is structured. Not exactly sure how, but I think this needs to be reworked more cleanly along semantic lines. `_preprocess_all_data` seems too "small", and `_preprocess_data` feels "too many indentations".

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: I think in v1, we have `auto` as default. But I agree, it is coupling with `pandas`.
Then I think do we really need this? I mean we could keep `None` as default and the user passes the transformer they want, that would
- Make the interface simpler
- Reduce coupling
- Also, as `auto` is not default, and the user wants to transform the data, they would instead pass the transformer they want, I think `auto` would've been useful as a default, but otherwise it is not that useful? 
   - Keeping it was a trade-off (coupling vs user convenience), but if we remove it from the default, it is better to not to have it at all

What do you think?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: yeah i forgot to remove this, I was checking if the `auto`  was working as expected and using the scalers as per requirement

phoeenniixx: Wil remove this

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: we are passing the preprocessed data in place of  raw `TimeSeries`.
Please see the comment below regarding this in detail.

phoeenniixx:
> * as a consequence of the above, I see the change in args of `_ProcessedEncoderDecoderDataset` very critically. I think we want to keep the `TimeSeries` input, and pre-processing should be `TimeSeries` to `TimeSeries`.

@fkiraly, Can you please explain why we should we do this? I mean that is kind of coupling no? The data module would call the data from `TimeSeries` , preprocess it and change it back to `TimeSeries` - INSIDE the data module class itself. This new "processed" `TimeSeries` would be used by the INTERNAL class of data module to create the data laoders. If everything is happening inside the class, why do we need to call the `TimeSeries` class and create a new instance of this class inside the data module and use that instead? 
Also, The `TimeSeries` class is meant to have just the tensors that are not processed. This class has no relation to the processing, no? 

Please correct me if I am misunderstanding the suggestion. But I am not very comfortable of creating a new instance of `TimeSeries` inside data module, that would be 2 separate layers colliding into one another, no?

Or are you suggesting that pre-processing happen somehow just outside the datamodule? 


> * for clearer structure, I would reconsider how `_preprocess_all_data` vs `_preprocess_data` is structured. Not exactly sure how, but I think this needs to be reworked more cleanly along semantic lines. `_preprocess_all_data` seems too "small", and `_preprocess_data` feels "too many indentations".

Hmm, I agree it is getting messy there, I would look at how to clean those methods. Thanks for pointing it out!

fkiraly:
> @fkiraly, Can you please explain why we should we do this? I mean that is kind of coupling no? The data module would call the data from `TimeSeries` , preprocess it and change it back to `TimeSeries` - INSIDE the data module class itself. This new "processed" `TimeSeries` would be used by the INTERNAL class of data module to create the data laoders. If everything is happening inside the class, why do we need to call the `TimeSeries` class and create a new instance of this class inside the data module and use that instead? 

Case consistency: if there are no scalers etc, then the data reaching the model is `TimeSeries`. But if we have some scalers, we would have a different type? You couple the internals through having to detect and handle different output cases, where it would not have to be.

Type consistency for representation of concept. `TimeSeries` is the primary type to represent time series with index etc in `tensor`-s. The output of preprocess_etc matches the conceptual model, so we should use `TimeSeries` to represent it, instead of inventing a new type.

> Also, The TimeSeries class is meant to have just the tensors that are not processed. This class has no relation to the processing, no?

As stated above, I see `TimeSeries` not as representing "unprocessed tensors", but the abstract data type "time series with index (and certain variable roles)".

phoeenniixx:
Sorry, I am getting a bit confused here.
From my understanding, `TimeSeries` is a part of the data pipeline - D1 layer. And it should not collide with D2 layer of the pipeline. 

But from the above explanation, it seems it is more of a "data-type" than a layer.
If that is true, I think we need to create a new definition of `TimeSeries`. 

I understand what you are trying to say - without scaling it would be a `TimeSeries` object that is sent to the datamodule for data loading and it would be coupled if we send some other data type when we are using scalers. But that is why every time we call the data inside `__getitem__`, we call preprocess_data first - that always provide a `dict` for the data loading (with or without scalers)

If we were to pass `TimeSeries` object and we do return a `TimeSeries ` object after preprocessing as well, then I think we should not advertise it as D1 layer but as a data type that is expected by the data pipeline of `pytorch-forecasting`.

What do you think? I think we need a discussion on this so that I can better understand your mental model on this.

fkiraly:
@phoeenniixx, I think you are right!

I was confusing `TimeSeries` (which takes as constructor input `pd.DataFrame`) with the `__getitem__` `dict`-of-tensors output. Indeed `TimeSeries` is a specific D1 data type, and we do not want to convert back to `pd.DataFrame` inside. Instead the `dict`-of-tensors output that you are now using is the more consistent choice here.

pytorch_forecasting/adapters/scaler_adapters.py
phoeenniixx: > what happens if we pass an (1 x m) tensor?

That would mean "m" targets with 1 sample each no? Meaning `(n_samples, n_targets)` format that it already handles?

> an (n x m) tensor in the single-target capable case?

I am not sure if this is possible? This tensor would multi-target no? How this is a single target?

phoeenniixx:
I tried adding different "strategies" for different scalers, I think this should make this code cleaner.

pytorch_forecasting/adapters/scaler_adapters.py
```diff
@@ -0,0 +1,205 @@
...
+    EncoderNormalizer,
+    GroupNormalizer,
+    MultiNormalizer,
+    NaNLabelEncoder,
+)
+
+
+def get_scaler_strategy(scaler) -> _ScalerStrategy:
+    """Single dispatch point: the only place that inspects scaler type."""
+    if _is_sklearn_scaler(scaler):
```
fkiraly: suboptimal for the dispatch/plugin pattern because you have two loci of extension - the class itself, and the registry here.

Better:
- has an `is_applicable` or similar method in the `_ScalerStrategy` subclass
- in `get_scaler_strategy`, import the list of all strategies, and loop through their `is_applicable`, then return the one that applies

If you export the list of strategies in the same module where you add them (or even collect them using `all_objects`), then there is a *single* extension locus.

pytorch_forecasting/adapters/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    """Convert any array-like to a float32 tensor."""
+    if isinstance(data, torch.Tensor):
+        return data.to(dtype)
+    elif isinstance(data, pd.Series):
+        return torch.tensor(data.to_numpy(), dtype=dtype)
+    return torch.tensor(np.asarray(data), dtype=dtype)
+
+
+def _is_sklearn_scaler(scaler):
+    return isinstance(scaler, _SKLEARN_SCALERS)
```
fkiraly: why not just check whether it is an sklearn transformer? That would be more extensible.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -58,18 +64,36 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
...
-    Optional[Dict[str, Union[StandardScaler, RobustScaler,
-                        TorchNormalizer, EncoderNormalizer]]], default=None
-        Dictionary of feature scalers.
+    scalers : optional, default=None
+        Mapping of continuous feature names to their designated scaling instances.
+
+        Defaults to ``None`` - an Identity pass-through, leaving the raw
+                    feature values untouched.
+        Supported scaler options for individual feature keys include:
+        * **PyTorch Forecasting Normalizers**:
```
fkiraly, line 89: I think you need an empty newline for this to render properly in the docs

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -733,10 +1085,19 @@ def collate_fn(batch):
             "target_past": torch.stack([x["target_past"] for x, _ in batch]),
             "encoder_time_idx": torch.stack([x["encoder_time_idx"] for x, _ in batch]),
             "decoder_time_idx": torch.stack([x["decoder_time_idx"] for x, _ in batch]),
-            "target_scale": torch.stack([x["target_scale"] for x, _ in batch]),
             "encoder_mask": torch.stack([x["encoder_mask"] for x, _ in batch]),
             "decoder_mask": torch.stack([x["decoder_mask"] for x, _ in batch]),
         }
+        if isinstance(batch[0][0]["target_scale"], list | tuple):
```
fkiraly, line 1090: this does not look right - why do we add this special case treatment here?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -648,39 +979,60 @@ def setup(self, stage: str | None = None):
...
                 self.test_dataset = self._ProcessedEncoderDecoderDataset(
-                    self.time_series_dataset,
                     self,
                     self.test_windows,
+                    self._test_preprocessed,
                     self.add_relative_time_idx,
                 )
         elif stage == "predict":
             predict_indices = torch.arange(len(self.time_series_dataset))
+            self._predict_preprocessed = {
```
fkiraly: repeated, it looks like this should be a method

fkiraly (changes requested):
Looks mostly ok.

Cimments above. Where very indented pieces are code are a added, consider adding a new private method instead and calling.

pytorch_forecasting/adapters/scaler_adapters.py
phoeenniixx: Yeah using `all_objects` now. But for that I had to make the classes public as `all_object` can't discover the private classes. But that is not much of an issue I think

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: Because for multi-target, this would be a `list` and tensor in other case

pytorch_forecasting/adapters/utils.py
phoeenniixx: I can do that, but I am a bit skeptical, do ALL the sklearn transformers follow the same strategy and are compatible to `SklearnStrategy`?

phoeenniixx: Also, `ptf` metrics use `TransformerMixin` from `sklearn` as their parent as well - same as `sklearn` transformers

So, I am not sure, if we can differentiate `ptf` metrics and `sklearn` transformers without that?
How does it happen in `sktime`? I tried to read it, but it calls a lot of things internally and I am getting lost.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: THanks. I have created separate methods for each of these

pytorch_forecasting/adapters/utils.py
fkiraly: > do ALL the sklearn transformers follow the same strategy and are compatible to SklearnStrategy

I think you might additionally need that the columns stay the same, but still it would be more extensible than it is right now.

(perhaps check and raise an error if an incompatible transformer is passed? Thoug I think it is not necessary in this PR)

fkiraly: > How does it happen in `sktime`? I tried to read it, but it calls a lot of things internally and I am getting lost.

Since `sklearn` does not have a clean object typing system (or secondary tag system - it has one but it is really bad), I think we check inheritance and some basic API assumptions.

pytorch_forecasting/adapters/scaler_adapters.py
fkiraly: You can also write your own all_object variant. Maybe we should give the `scikit-base` one an option so the user can say what they want to include or exclude?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
fkiraly: it still does not look right and indicates there are some design failures with the multi-target case. But lets look at this separately.

tests/test_data/test_data_module.py
```diff
@@ -458,15 +478,239 @@ def test_multivariate_target():
         group=["group"],
         num=["feature1", "feature2"],
     )
-
     dm = EncoderDecoderTimeSeriesDataModule(
         time_series_dataset=ts,
         max_encoder_length=10,
         max_prediction_length=5,
         batch_size=4,
+        target_normalizer=normalizer_list,
```
fkiraly, line 487: we should also ensure the old default usage still works (no, or default normalizers)

fkiraly (changes requested):
I think this is ok now, while there are still some issues:

Blocking:
* we should ensure that "no normalizers" is a case that still works - including tests that work with it. If it is already so, please point me to the test - but I see that you are changing a test, for example.

Non-blocking:
* the multi-target metric case does not look working well, the API appears hacky (see comment above). This should be revisited soon.
* the topic about allowing generic `scikit-learn` transformers. Hard-coding individual transformers is not good practice, instead an API contract should be assumed. Again, this is something we can look at separately.

phoeenniixx:
> * we should ensure that "no normalizers" is a case that still works - including tests that work with it. If it is already so, please point me to the test 

I have already the test for `None` for single target [here](https://github.com/phoeenniixx/pytorch-forecasting/blob/4a38e5ccc6ea38958b17f10df757aad7260725b6/tests/test_data/test_data_module.py#L575), but I  forgot to  add that for multi-target. I have pushed the changes now. Thanks for pointing it out!

> but I see that you are changing a test, for example.

I just added `target_normalizer` there, as previously we didnt support it, but now we support it, so it would be better to have those passed to the datamodule as well. Do you think I should move it out? Although the `None` case is replicating the test that was before...

phoeenniixx:
> * the multi-target metric case does not look working well, the API appears hacky (see comment above). This should be revisited soon.

Hmm, I cant see how to sort this out, you mean there should not be case distinctions in `collate_fn`? We might need to move this to a new method or adapter then. Or do you mean this should be handled a layer above in `__getitem__` somehow?

> * the topic about allowing generic `scikit-learn` transformers. Hard-coding individual transformers is not good practice, instead an API contract should be assumed. Again, this is something we can look at separately.

Yeah, I was a bit confused how to collect all the transformer here, as I could see all the `sklearn` transformers use `TransformerMixin`, but so does the `ptf` metrics, so this would not work in here, then second common class I could see was `OneToOneFeatureMixin`, but that is not common to all the transformers like  `KernelCenterer` doesnt use it. Then I saw [these lines](https://github.com/sktime/sktime/blob/a1d78b5a3574b1c0950b208cd5f28a64efe006d7/sktime/utils/sklearn/_scitype.py#L66-L69) in `sktime`, maybe we coykd try something like this in `ptf` as well. Then we should be able to collect all teh transformers?
```python
    is_in_sklearn = issubclass(obj, SklearnBaseEstimator)
    is_in_sktime = issubclass(obj, BaseObject)

    is_sklearn_est = is_in_sklearn and not is_in_sktime
```


EDIT: It actually works! I found a class [TransformMixIn](https://github.com/phoeenniixx/pytorch-forecasting/blob/4a38e5ccc6ea38958b17f10df757aad7260725b6/pytorch_forecasting/data/encoders.py#L134), which is a parent for all the `ptf` transformers, so I used it. THe issue is it's name is VERY close to `sklearn`'s `TransformerMixin`, but ig that is not an issue? As this is a private method and is deep in the stack, i dont think most people would read it and get confused?

tests/test_data/test_data_module.py
phoeenniixx: Yeah I didnt add the tests for multi-target case, Thanks, I have added them as well now.

phoeenniixx:
> > * the multi-target metric case does not look working well, the API appears hacky (see comment above). This should be revisited soon.
> 
> Hmm, I cant see how to sort this out, you mean there should not be case distinctions in `collate_fn`? We might need to move this to a new method or adapter then. Or do you mean this should be handled a layer above in `__getitem__` somehow?

Even v1 had similar way handling the `target_scale` : 
https://github.com/sktime/pytorch-forecasting/blob/c4f5aac341082c56951bed2f352dc3bf6087ce35/pytorch_forecasting/data/timeseries/_timeseries.py#L2469-L2486

The issue is that for single target we return a `tensor` and for multi-target it is a `list`, we need to check this distinction somewhere anyhow, no? Or we should make a unified return, `list` , even for single target. But that would mean we might need to update the things in some places

Even same would happen for the `target` as it can also be a `list` and a tensor (when we try to add weights for `MultiLoss`)

https://github.com/sktime/pytorch-forecasting/blob/c4f5aac341082c56951bed2f352dc3bf6087ce35/pytorch_forecasting/data/timeseries/_timeseries.py#L2496-L2523

fkiraly (approved):
I think this is fine now given the cleaner structure. The multitarget issue we can solve later, separately.

fkiraly:
May I request to merge https://github.com/sktime/pytorch-forecasting/pull/1983 first then this, in order to give @PranavBhatP credit? Should be unproblematic because this stacks on #1983.

phoeenniixx:
Actually #1983 and this branch has diverged now as we created a new PR for `save` and `load` (#2323), so merging that PR will create issues as this PR will not overwrite the changes in `BasePkg`. I will instead add @PranavBhatP a co-author to this PR to give him credit

Issue #1957, opened 2025-08-27
[ENH] refactor `_loss_mapping` to tags of metrics, entirely


fkiraly:
The `_loss_mapping` looks like it should be refactored to tags of metrics, entirely, in the respective `_pkg` classes.

Let´s do this! Though we should think about prioritization, or whether this is a "good first issue".

ParamThakkar123:
@fkiraly Should I take this up ?

fkiraly:
@ParamThakkar123, if you would like to, sure! This is not a good first issue though, as it requires some familiarity of the framework under the hood.

May I ask you, hence, first, to outline here in the issue the changes that you would make?

ParamThakkar123:
Hello @fkiraly sorry for the late reply, I forked the repository and made same changes to it I haven't started the pull request yet but this is kind of an outline how I would be making the changes: 

https://github.com/ParamThakkar123/pytorch-forecasting/commit/579cf5affe70475e3e64654f9dc753908c094d78

Would love any comments on these!

fkiraly:
Thanks! I thought primarily the `LOSSES_BY_PRED_AND_Y_TYPE` and the if/else to determine tensor sizes should be moved to tags.

The parameters are for tests, and they could be returned by a method like `_get_test_dataloaders_from`

ParamThakkar123:
Sure would make the changes to it.

ParamThakkar123:
Does this changes look good ? Should I start a PR then maybe you can review it @fkiraly ?

fkiraly:
@ParamThakkar123, yes, generally speaking, the changes look great! Maybe I'd name the tags differently, but that is more something for when the design is consolidated imo. There is some detail discussion (see what @phoeenniixx said) to be had on teh exact design though.

I would also advise you split the formatting that does not impact losses into a separate PR.

ParamThakkar123:
Sure @fkiraly , I will make the changes required, I fixed somethings locally and testing them out, once all the tests pass, I will move to fixing the formatting and make a separate pull request for it. Then maybe we can discuss the steps further.

Faakhir30:
I'd like to wrok on this one, I'm thinking of following design:

#### Regarding `LOSSES_BY_PRED_AND_Y_TYPE` and `get_compatible_losses`: 

Have `tests:compatible_losses` tag at pkg level, it should have string of supported losses directly, e.g.,
```python
"tests:compatible_losses": [
            "MAE",
[...]

PR #2313, opened 2026-06-16
[ENH] Create a unified Test framework and a Base class for Data Modules in v2

### LINKED ISSUE #2298: [ENH] Create a unified Test framework and a Base class for Data Modules in v2
Currently, the data_modules are tested using standalone tests, I think we should have a unified test framework (like we have one for `metrics` and `models`) for data modules as well. 
A lot of tests, imo, are duplicate and can be added to the unified test framework. This would also give us a chance to add `tags` to this module of the package

I think we should also have a Base class for this module, the class that all data modules can inherit from. tTo prevent any code duplication

**Additional context**
I think we could extend this to D1 layer as well. But I am not sure if this is a good idea, as I think D1 layer is going to stay the same (only one `TimeSeries` class) for atleast in the near future. So here the test framworks would still work kinda standalone.

Faakhir30:
#### Reference Issues/PRs

Fixes #2298 

- Create `BaseTimeSeriesDataModule` in `_base_data_module.py` with shared init, split, setup template, metadata cache, dataloader factories (implement private abstract methods in child classes for DM specific logic)
- Refactor `EncoderDecoderTimeSeriesDataModule` and `TslibDataModule` to inherit base
- Add `_BasePtDataModule`, `EncoderDecoderDataModule_pkg`, `TslibDataModule_pkg` (tags and test param factories)
- Create `test_all_data_modules.py` + `_datamodule_config.py` using `BaseFixtureGenerator`
- Register `TestAllDataModules` in `test_class_register.py`
- Migrate duplicate tests and remove/slim `test_data_module.py` and `test_tslib_data_module.py`
- Update/Add `extension_templates/v2/datamodule/` stub
- restructure files in `data_modueles`

Yes, `test_all_data_modules.py`

#### PR checklist

- [x] Added/modified tests


extension_templates/v2/data_module/_datamodule_pkg.py
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @classmethod
+    def get_cls(cls):
+        from extension_templates.v2.data_module.data_module import MyDataModule
+
+        return MyDataModule
+
+    @classmethod
+    def get_test_timeseries(cls, **kwargs):
```
phoeenniixx: Why are we importing the data scenarios here? I think that should happen inside the test framework and not at the `pkg` level?

pytorch_forecasting/data/data_module/base/_base_data_module.py
```diff
@@ -0,0 +1,240 @@
...
+        self.val_windows: list | None = None
+        self.test_windows: list | None = None
+        self.predict_windows: list | None = None
+        self.train_dataset: Dataset | None = None
+        self.val_dataset: Dataset | None = None
+        self.test_dataset: Dataset | None = None
+        self.predict_dataset: Dataset | None = None
+
+    @staticmethod
+    def _coerce_target_normalizer(
```
phoeenniixx, line 104: I think this part still need clear design (how to support the scaler and `target_normalizer`), #2302 will be useful here, but for now I would suggest not  to add these methods

pytorch_forecasting/data/data_module/base/_base_data_module.py
```diff
@@ -0,0 +1,240 @@
...
+    @abstractmethod
+    def _context_length(self) -> int:
+        """Return encoder/context window length."""
+
+    @abstractmethod
+    def _prediction_length(self) -> int:
+        """Return decoder/prediction window length."""
+
+    @abstractmethod
+    def _create_windows(self, indices: torch.Tensor) -> list[tuple[int, int, int, int]]:
```
phoeenniixx, line 160: If we have this method, why do we have `self.train_windows` etc in `__init__`, I think this method would return the windows no? We dont need to create them at the initialization, but only when we actually need it - during data loading etc.

pytorch_forecasting/data/data_module/_base_data_module.py
```diff
@@ -0,0 +1,240 @@
...
+        """Compute metadata for model initialization.
+
+        This property returns a dictionary containing the shapes and key information
+        related to the time series model.
+        """
+        if self._metadata is None:
+            self._metadata = self._prepare_metadata()
+        return self._metadata
+
+    def _compute_split_indices(self):
```
phoeenniixx: I think splitting is not mature enough in v2, we should add this once we have a clear design?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -623,100 +577,6 @@ def _create_windows(self, indices: torch.Tensor) -> list[tuple[int, int, int, in
 
         return windows
 
-    def setup(self, stage: str | None = None):
```
phoeenniixx: Why are we removing this `setup`? The `setup` of `BaseDataModule` is not enough and is very generic - which is a good thing. But the children could (and should) have their own specific setup when needed..

phoeenniixx (changes requested):
Thanks! and welcome to `pytorch-forecasting`.
This looks nice!
- I think we need a dataset class in `BaseDataModule`? Or not?
- The `setup` etc methods can be data module specific, the base class is not able to handle those, which is expected. So, I think we should not remove them, or is there a way to generalise them?
- I think the data set of `EncoderDecoderTimeSeriesDataModule` could also go out of the class (like `TsLibDataModule`)?
- Please create separate folders for base, encoder-decoder and `tslib` and then add those data modules and pkg class to that.
- Also, for time being, we should keep the test files for both data modules, we need to make sure we dont miss any test coverage!

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
Faakhir30: > The setup of BaseDataModule is not enough

For tslib and enc-dec data module, I think it _is_ sufficient apart from spliting indices logic that we have removed for now.
My knowledge is limited to these two data modules as I have only seen these two, only thing different in `setup` of these two is dataset creation, and I do have that smaller abstract method `_build_dataset` overriden in both subclases. 
So, atleast in these two cases, idts subclasses needs thier own setup.

pytorch_forecasting/data/data_module/_base_data_module.py
Faakhir30: sure, done.

pytorch_forecasting/data/data_module/base/_base_data_module.py
Faakhir30: > why do we have self.train_windows etc in __init__

At **init** attributes are just being assigned as `None` to avoid `hasattr` checks that data modules originilay had. e.g., `if not hasattr(self, "test_dataset"):`

> but only when we actually need it - during data loading etc.

If this comment is suggesting to avoid saving datasets and windows as attributes at all, then reason behind saving them is caching behavior of windows/datasets in `setup` call, we wont be able to cache creation of dataset object without saving them.

pytorch_forecasting/data/data_module/base/_base_data_module.py
Faakhir30: logic of this method is borowed from subclasses being refactored, as discussed in friday meeting, keeping this for now.

extension_templates/v2/data_module/_datamodule_pkg.py
Faakhir30: sure, we can uncouple this thing, but then we will have to maintain a map of datamodule `batch_format` tag to a D1 TimeSeries returning function used by tests.

currently done at `_datamodules_config.py`

**Edit**: should'nt `pkg` layer be responsible for describing "how to test" the class it encapsulates? was above comment made because of coupling test framework code `_data_senarios` to pkg layer?

Faakhir30:
>I think we need a dataset class in BaseDataModule? Or not?

Not required in current approach, i'm leaving `_build_dataset` as abstract method to offload dataset creation to subclasses, idts its possible to generalize `__getitem__` of dataset classes and have a base class attached  as default dataset class for `BaseDataModule`.
A previous comment regarding refactoring of datasets https://github.com/sktime/pytorch-forecasting/issues/2298#issuecomment-4724653903
 

>The setup etc methods can be data module specific, the base class is not able to handle those, which is expected. So, I think we should not remove them, or is there a way to generalise them?

Addressed at https://github.com/sktime/pytorch-forecasting/pull/2313#discussion_r3434027927

>I think the data set of EncoderDecoderTimeSeriesDataModule could also go out of the class (like TsLibDataModule)?

>Please create separate folders for base, encoder-decoder and tslib and then add those data modules and pkg class to that.

>Also, for time being, we should keep the test files for both data modules, we need to make sure we dont miss any test coverage!

Done, tests will be failing for now as old files are having indices/windowing logic etc that we removed in newer design, thanks!

phoeenniixx:
> Done, tests will be failing for now as old files are having indices/windowing logic etc that we removed in newer design

If we are removing it from the base class, why not add them back to the child classes?
When I meant we plan to "remove" those, I meant in the future, not in this PR. We need to replace that part of the code with some cleaner and more advanced version, but that is still a WiP. So, unless we have a clear picture of what to do, I suggest, leaving those parts as it is as they are on main.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: Hmm, I see, for these classes `setup` generalisation worked, but that would not be the case when we add FMs to this package - FMs tend to handle data in a bit different way. But that is not an issue, we would just override this default behaviour. For now, if we are able to generalise, then it is a good thing! Nice!

phoeenniixx: >  I think it is sufficient apart from spliting indices logic that we have removed for now.

I think we could create some abstract methods for splitting and windowing and then the child classes create those? Would that make sense? 
I think windowing can how they are handled can be different for different types of model families (or datamodules). What do you think @agobbifbk @Faakhir30 @fkiraly

Faakhir30: windowing already has abstractmethod `_create_windows` that is implemented by subclasses.
will do same for splitting in next commit, thanks

extension_templates/v2/data_module/_datamodule_pkg.py
phoeenniixx, line 1: I think it would be good if you could add comments and docstrings explaining what the methods do and what to do in these methods. See existing extension templates of v2 to see what i mean

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
phoeenniixx, line 1: I think the name of the file still should be `_encoder_decoder_data_module`, having `_data_module` as the name of the file for totally two different types of the data module can be confusing imo. Although the folder makes it a bit clear, but I think still it owuld be better to keep the name as it is

(Same for pkg classes of data module)
For extension templates, the naming is ok, but for the package itself, I am a bit skeptical

pytorch_forecasting/tests/_data_scenarios.py
```diff
@@ -372,3 +372,109 @@ def dataloaders_fixed_window_without_covariates():
...
+        time="time",
+        target="target",
+        group=["group"],
+        num=["cont_feat1", "cont_feat2", "known_future"],
+        cat=["cat_feat"],
+        known=["known_future"],
+    )
+
+
+def make_tslib_timeseries(**kwargs):
```
phoeenniixx: why are we making two different scenarios for the data modules? I think they both should be able to handle the similar type of `TimeSeries` class?
Also, Is the already avaiable method to create `TimeSeries` not enough? Do we really need to add new methods for this?

pytorch_forecasting/tests/_datamodule_config.py
```diff
@@ -0,0 +1,35 @@
...
+    make_encoder_decoder_timeseries,
+    make_tslib_timeseries,
+)
+
+EXCLUDE_DATA_MODULES = [
+    "ClassName",
+]
+
+EXCLUDED_TESTS = {}
+
```
phoeenniixx: would it be better to have `tests:skip_by_name` tag here? that is what we use for the test framework for the models

pytorch_forecasting/tests/test_all_data_modules.py
```diff
@@ -0,0 +1,571 @@
...
+            for col in time_series_metadata["cols"]["x"]
+            if time_series_metadata["col_type"].get(col) == "F"
+            and time_series_metadata["col_known"].get(col) == "K"
+        ]
+    )
+    return known_cat_count, known_cont_count
+
+
+def _tslib_known_feature_counts(metadata):
+    """Count known categorical and continuous features for tslib format."""
```
phoeenniixx: We could count these feats from the `TimeSeries` metadata (like we are doing for `EncoderDecoderTimeSeriesDataModule`. Same method can handle both cases here or?

phoeenniixx (changes requested):
Thanks!
It is moving in a good direction!
I think we can reduce the redundancy here though. I think there are a lot of things that both these datamodules do in a very similar way. So, in the tests, they should be tested in a simiar way and under same tasks as well. In future, if some datamodule deviates from this pattern, we could add the new tests then

pytorch_forecasting/tests/_data_scenarios.py
Faakhir30: > Also, Is the already avaiable method to create TimeSeries not enough? Do we really need to add new methods for this?

in data senarios, i dont see any function returning a single generic `TimeSeries`, there is one to return three timeseries (train, test, val)

> I think they both should be able to handle the similar type of TimeSeries class?

originaly had two functions as `test_tslib_data_module` and `test_data_module.py` had seperate timeseries for testing, but yes, we can unify this to single timeseries, done

pytorch_forecasting/tests/_datamodule_config.py
Faakhir30: I do like tag approach, it provides better dev experience, also original reason why we went in that direction in sktime repo. sktime/sktime#8496

- As I read through codebase, I am seeing we are not documenting/maintaing tags in `registry/_tags` unlike sktime /skpro, is this intentional? or a future item? (i'd like to work on it if possible)

- I also see that ptf project has almost created convention of `EXCLUDED_TESTS` in other pkgs testing framework too, we should add `tests:skip_by_name` support to all of them? (will open seperate PR for this, if maintainers approves, otherwise will fix for only data modules tests here)

Faakhir30:
moved `_preprocess_data` to base too, now i think we only have `_build_dataset`, `collate_fn`, `_prepare_metadata`, and `_create_windows` at subclasses level. I'd prefer not to refactor these further as they are tightly coupled to data modules metadata format and/or subclass specific params at `init`.


**Unifying Dataset getitem format**

I think `test_all_datamodules` should be Dataset agnostic, I mean there are `if`-`else if` blocks in testing rn due to different formats being returned by Dataset `__getitem__` of both data modules. For example:
 
https://github.com/Faakhir30/pytorch-forecasting/blob/bc07b60f0434cb6e56dc86b0d77fbbfcf83d1802/pytorch_forecasting/tests/test_all_data_modules.py#L101-L115

I'm unsure if its possible to generalize dataset item format in longer run??, confirmation from maintiners appriciated @phoeenniixx @fkiraly. I think we can unify atleast current two data modules' datasets formats though.

- one way is to unify at dataset level, have dataset base class too if possible (if answer to above question is YES)
- another approach can be to have a method at pkg level that maps role of keys in dataset item to unified semantics, and tests can be generalized on this unified semantics. (if answer to above question is NO, lowers risk of future problems with implementing dataset to this format)

e.g.
```python
class EncoderDecoderDataModule_pkg:
    ...
    @classmethod
    def get_tensor_roles(cls):
        """Map generic history/future roles to encoder-decoder batch keys."""
        return {
            "history_cat": "encoder_cat",
            "history_cont": "encoder_cont",
            "future_cat": "decoder_cat",
            "future_cont": "decoder_cont",
            "history_length": "encoder_lengths",
            "future_length": "decoder_lengths",
            "history_mask": "encoder_mask",
            "future_mask": "decoder_mask",
            "history_time_idx": "encoder_time_idx",
            "future_time_idx": "decoder_time_idx",
            "history_target": "target_past",
            "groups": "groups",
        }
```
and in testing we can do something like:
```
roles = object_pkg.get_tensor_roles()
assert x[roles["history_cat"]].shape[0] == object_instance.context_length()
```

phoeenniixx:
> * another approach can be to have a method at pkg level that maps role of keys in dataset item to unified semantics, and tests can be generalized on this unified semantics. (if answer to above question is NO, lowers risk of future problems with implementing dataset to this format)

I am more inclined towards this. One of the motivations to create `tslib` datamodule was the TERMINOLOGY. I dont think we should change the var names here. Also, as we add more feats (like scalers etc), these datamodules will grow apart into different entities, if we unify them now, it may get problematic in future.


You could create a dict like strucutre and use that in the tests itself as well... (another option that saves the contributor to add a new method in the `pkg`). Where you collect the names which have same (or similar) meanings and then use the common words as maps
Like this:
```python
var_map={
       "history_cat": ["history_cat", "encoder_cat", ...],
         ....
}
```

In this case, if we add a new data module, and that has similar terminology, we wont have to implement a method there, insteaed we would use this map. If a new name appear like `history_names` (or whatever), we just add taht to the list and move on.

This dict would lie in the test framework and would be used by the tests only

Not sure if this makes things easier or more complicated?
What are your thoughts @Faakhir30 @fkiraly @agobbifbk?

Faakhir30:
> You could create a dict like strucutre and use that in the tests itself as well... 

In most of cases users will have to update this `var_map`, i'd perfer to have it coupled to pkg layer.
I think i am seeing this problem of "saving mappings at test config level vs pkg layer" at multiple fronts.
 I think it is better to standardize pkg layer to be the default place of describing "how to test  the class that I encapsulate" via tags and/or methods.
>(another option that saves the contributor to add a new method in the pkg).

As for reducing boilerplate, I'd prefer to have `utils/defaults/_data_modules.py` etc to have default maps/methods/functions, etc, user can override the keys they need at pkg layer while reducing boilerplate.
```python
from utils.defaults import TENSOR_ROLES

class DataModule_pkg:
        def get_tensor_roles(self):
               return {**TENSOR_ROLES,
                                  "overriden_key": "new_value",
                              }
```

I came to this approach while trying to:
- moving `EXCLUDED_TESTS` logic to pkg layer 
- defining this unified dataset item mapping for generalizing tests
- also, while designing solution for #1957

phoeenniixx:
> In most of cases users will have to update this `var_map`, i'd perfer to have it coupled to pkg layer. I think i am seeing this problem of "saving mappings at test config level vs pkg layer" at multiple fronts. I think it is better to standardize pkg layer to be the default place of describing "how to test the class that I encapsulate" via tags and/or methods.

Umm, I think there are not many variations of `encoder_cat` (and other variables), so in most cases, they dont have to add anything in the dict. Also, to add a new data module, the contributor would have to first read and see ALL the naming conventions which might not even be related to the work they are doing. Also, I feel it makes `pkg` a bit too much coupled with the test framework. The `pkg` layer should just provide basic info to the framework, not handle everything for it.


> As for reducing boilerplate, I'd prefer to have `utils/defaults/_data_modules.py` etc to have default maps/methods/functions, etc, user can override the keys they need at pkg layer while reducing boilerplate.
> 
> ```python
> from utils.defaults import TENSOR_ROLES
> 
> class DataModule_pkg:
>         def get_tensor_roles(self):
>                return {**TENSOR_ROLES,
>                                   "overriden_key": "new_value",
>                               }
> ```
> 

Can you explain it? Sorry, i didnt understand what you are trying to say here.

Faakhir30:
> Umm, I think there are not many variations of encoder_cat (and other variables), so in most cases, they dont have to add anything in the dict. Also, to add a new data module, the contributor would have to first read and see ALL the naming conventions which might not even be related to the work they are doing. Also, I feel it makes pkg a bit too much coupled with the test framework. The pkg layer should just provide basic info to the framework, not handle everything for it.

alright, moving in suggested direction

> Can you explain it? Sorry, i didnt understand what you are trying to say here.

ahhh, yes, that default mapping concept is broken for this senario, I just copied that from design i was doing for another similar problem at #1957 , appologies

extension_templates/v2/data_module/_datamodule_pkg.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    # todo: fill out datamodule tags here
+    # tags are inherited from parent class if they are not set
+    _tags = {
+        # Required: identifies this as a v2 datamodule package (do not change).
+        "object_type": "datamodule_v2",
+        # Labels the batch layout for format-specific test assertions.
+        # Use "encoder_decoder" or "tslib" when your keys match an existing format.
+        "batch_format": "encoder_decoder",
```
phoeenniixx, line 50: In future we would have more types of datamodules, would it make sense to have something lik `"info:datamodule_type"`? Though I cant see how it would be different from `name` tag. 
I think almost all datamodules would have their own batch formats no? Why else someone would habe to create a new type of datamodule in place of using (or updating and then using) the already implemented data mdoules?

extension_templates/v2/data_module/_datamodule_pkg.py
```diff
@@ -0,0 +1,144 @@
...
+        # Human-readable name; should match the Lightning datamodule class name.
+        "info:name": "MyDataModule",
+        # Whether tests should run static-feature scenarios
+        # (adds static cols to D1 data).
+        "capability:static_features": False,
+        # Whether tests should run multivariate-target scenarios.
+        "capability:multivariate_target": False,
```
phoeenniixx: I think all datamodules should have `static_feats` and `multivariate` target? No?

pytorch_forecasting/data/data_module/base/_base_data_module.py
```diff
@@ -0,0 +1,353 @@
...
+        batch : list of tuple[dict, target]
+            Samples as returned by the processed dataset.
+
+        Returns
+        -------
+        tuple[dict, target]
+            Collated ``x`` dict and ``y`` (tensor or list of tensors for multivariate).
+        """
+
+    def _train_shuffle(self) -> bool:
```
phoeenniixx: should this be a property?

phoeenniixx (changes requested):
Nice!
I have left some comments for the extension templates.
Also, for the `test_all_data_modules`, should we create a new file which hosts all the test logic and then we call it in there? SImilar to [`_integration`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/tests/test_all_v2/_test_integration.py) and   and [`test_integration`](https://github.com/sktime/pytorch-forecasting/blob/4d8d97cd3e85a15a9b90a38dfb0afc819d8e8aa4/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py#L32) in `test_all_v2`? That would keep this file clean and we could write the docstring that explains what each test does in `test_all_data_modules`

extension_templates/v2/data_module/_datamodule_pkg.py
Faakhir30: > I think almost all datamodules would have their own batch formats no?

I am also assuming this, `batch_format` being the primary identifier of a data module. If this is the truly the case, then idts we need that `info:datamodule_type` tag. 
We can introduce "info:datamodule_type" later if ever needed.

extension_templates/v2/data_module/_datamodule_pkg.py
Faakhir30: I'm unsure, following your suggestion, removing these for now :)

Faakhir30:
> should we create a new file which hosts all the test logic and then we call it in there? SImilar to [_integration](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/tests/test_all_v2/_test_integration.py) and and [test_integration](https://github.com/sktime/pytorch-forecasting/blob/4d8d97cd3e85a15a9b90a38dfb0afc819d8e8aa4/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py#L32) in test_all_v2? That would keep this file clean and we could write the docstring that explains what each test does in test_all_data_modules

I'm fine with either approach, though i'd prefer current one, its easy to verify docstring against the logic rn, moving logic to another file for each test method seems redundant, I'll update it if you feel strong about this one, lmk.

phoeenniixx:
> > should we create a new file which hosts all the test logic and then we call it in there? SImilar to [_integration](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/tests/test_all_v2/_test_integration.py) and and [test_integration](https://github.com/sktime/pytorch-forecasting/blob/4d8d97cd3e85a15a9b90a38dfb0afc819d8e8aa4/pytorch_forecasting/tests/test_all_v2/test_all_estimators_v2.py#L32) in test_all_v2? That would keep this file clean and we could write the docstring that explains what each test does in test_all_data_modules
> 
> I'm fine with either approach, though i'd prefer current one, its easy to verify docstring against the logic rn, moving logic to another file for each test method seems redundant, I'll update it if you feel strong about this one, lmk.

It makes the methods too crowded though no?

phoeenniixx (changes requested):
I think we should also use the tag - `tests:skip_by_name` while collecting the fixtures. 
Also, if you want to keep the tests as it is (not moving them to a new location) - then that is how we move, but please add complete docstrings to it, explaining what each tests check?

Faakhir30:
> Also, if you want to keep the tests as it is (not moving them to a new location) - then that is how we move, but please add complete docstrings to it, explaining what each tests check?

I've improved tests docstrings somewhat. lmk, if further refinement is needed, thanks

> It makes the methods too crowded though no?

Yeah, valid point, but i'd prefer to break it into more unit tests (methods), if u feel some test needs to be further breaked, lmk, i mostly copied tests from existing files, will look into this more again.

> I think we should also use the tag - tests:skip_by_name while collecting the fixtures.

Yes, done now, previously i skipped it to as a future item to avoid large PR, but seems like its already quite large :)

Faakhir30:
Update:
- fixed merge conflicts
- most of work introduced by #2302 is kept at `encoder_decoder_data_module` for now in  `#region preprocessing` comment block
- after we have designed TSLibs preprocessing somewhat #2330, I'll revisit refactoring of preprocessing across data modules.

This PR is finally ready for final passes of review I guess ;)

pytorch_forecasting/data/data_module/base/_base_data_module.py
```diff
@@ -0,0 +1,355 @@
...
+        self.categorical_indices, self.continuous_indices = (
+            self._extract_feature_type_indices(self.time_series_metadata)
+        )
+
+        self._metadata = None
+        self.train_windows: list | None = None
+        self.val_windows: list | None = None
+        self.test_windows: list | None = None
+        self.predict_windows: list | None = None
+        self.train_dataset: Dataset | None = None
```
phoeenniixx, line 99: do we need these type hints?

pytorch_forecasting/data/data_module/base/_base_data_module.py
```diff
@@ -0,0 +1,355 @@
...
+        categorical_indices = []
+        continuous_indices = []
+        for idx, col in enumerate(time_series_metadata["cols"]["x"]):
+            if time_series_metadata["col_type"].get(col) == "C":
+                categorical_indices.append(idx)
+            else:
+                continuous_indices.append(idx)
+        return categorical_indices, continuous_indices
+
+    def _preprocess_data(self, series_idx) -> dict[str, Any]:
```
phoeenniixx, line 129: should this be an abstract method?

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -125,7 +126,7 @@ def __init__(
         | str
         | list[NORMALIZER]
         | tuple[NORMALIZER]
-        | None = None,
+        | None = "auto",
```
phoeenniixx: please dont change teh defaults, we should keep the default None - no normalization

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -136,84 +137,59 @@ def __init__(
...
-            )
-            if name in self._scalers
-        ]
+    def _context_length(self) -> int:
+        return self.max_encoder_length
+
+    def _prediction_length(self) -> int:
+        return self.max_prediction_length
+
+    def _build_dataset(self, windows: list[tuple[int, int, int, int]]) -> Dataset:
```
phoeenniixx: pleae look at `_make_dataset`, maybe just rename that method?

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -136,84 +137,59 @@ def __init__(
...
         train_val_test_split: tuple = (0.7, 0.15, 0.15),
     ):
-        self.time_series_dataset = time_series_dataset
+        super().__init__(
+            time_series_dataset=time_series_dataset,
+            target_normalizer=target_normalizer,
+            batch_size=batch_size,
+            num_workers=num_workers,
+            train_val_test_split=train_val_test_split,
+            add_relative_time_idx=add_relative_time_idx,
```
phoeenniixx, line 146: please collect the params in `self` first.

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -136,84 +137,59 @@ def __init__(
...
-                self.categorical_indices.append(idx)
-            else:
-                self.continuous_indices.append(idx)
-
         self._scalers = {
             k: ScalerAdapter(v) for k, v in _coerce_to_dict(scalers).items()
         }
         self._build_cont_scalers()
 
-    def _build_cont_scalers(self):
```
phoeenniixx, line 207: why are we removing this method? this is needed for scalers no?

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -313,36 +289,117 @@ def _prepare_metadata(self):
 
         return metadata
 
-    @property
-    def metadata(self):
-        """Compute metadata for model initialization.
+    def _create_windows(self, indices: torch.Tensor) -> list[tuple[int, int, int, int]]:
```
phoeenniixx: why reimplement this? I think we already have this implemented in the data module, no? Or are you moving the locations for some reason?

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
```diff
@@ -1016,7 +752,7 @@ def setup(self, stage: str | None = None):
                 self._fit_target_normalizer(self._train_indices)
             if not self._feature_scalers_fitted:
                 self._fit_scalers(self._train_indices)
-            if not hasattr(self, "train_dataset") or not hasattr(self, "val_dataset"):
+            if self.train_dataset is None or self.val_dataset is None:
```
phoeenniixx, line 743: why are we changing this, although it doesnt make any difference in the logic?

phoeenniixx (commented):
Just reviewed `EncoderDecoderDataModule`, but I have a doubt, why are we changing the sequence of the methods and lines? I see a lot of lines which are same but they have now been moved to some different location. Although this doesnot change the logic, but it leads to the confusion while reviewing - what exactly is the change and what part has just changed the location.

Can you please help me here and clean this a bit - we should always prefer to keep the changes to the minimum in the PRs, and make the changes only which are needed

phoeenniixx:
And readthedocs is failing - meaning there is some issue in the docs as well, from the traceback from the doc build (you can see the build by clicking on the readthedocs button on the CI), the problem is `EncoderDecoderDataModule`.

As you chnaged the path of the datamodule, you should also update it in the docs

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
Faakhir30: it does, I'm treating these at attributes set at base layer (considering these are general attributes all data modules will have)
Since we set the `self.train_dataset = None` at base init, this `hasattr` check will pass while it should not as `train_dataset` is `None`, so `is None` is the right check.
Its cleaner to set attributes at init when possible instead of setting at runtime and having getattr/hasattr.

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
Faakhir30: not removed, moved below, enclosed preprocessing related methods in a region, so that its easy to refactor logic of preprocessing once we have decided preprocessing steps of tslib too. 

btw how do you tag large code blocks for later refactor? I used [region comments for preprocessing related logic](https://code.visualstudio.com/docs/editing/userinterface#:~:text=such%20as%20//%23region%20comments%2C%20then%20the%20minimap%20shows%20the%20folding%20marker%20names.%20Folding%20markers%20are%20language%2Dspecific%2C%20so%20check%20which%20markers%20apply%20to%20your%20language.).

pytorch_forecasting/data/data_module/encoder_decoder/_encoder_decoder_data_module.py
Faakhir30: not needed, because parent `init` is saving these to `self`, these are common params in both data modules for now.

pytorch_forecasting/data/data_module/base/_base_data_module.py
Faakhir30: Not necessarily, but they help IntelliSense.

Faakhir30:
> And readthedocs is failing - meaning there is some issue in the docs as well, from the traceback from the doc build (you can see the build by clicking on the readthedocs button on the CI), the problem is EncoderDecoderDataModule.

Fixed, thanks! I thought it was another CI issue, similar to the sktime docs CI.

> why are we changing the sequence of the methods and lines? I see a lot of lines which are same but they have now been moved to some different location. Although this does not change the logic, but it leads to confusion while reviewing - what exactly is the change and what part has just changed location?

Yeah, you're right. From a review perspective, it ended up being quite messy, apologies. 
Resolving merge conflicts and trying to consolidate the preprocessing logic in one place caused this.

> Can you please help me here and clean this a bit - we should always prefer to keep the changes to the minimum in the PRs, and make the changes only which are needed

Done. reduced the diff by  ~850 lines by moving things back closer to their original locations. 
Hopefully it's much easier to review now.


---

```python
# in enc-dec setup:
        if stage is None or stage == "fit":
                self.val_dataset = self._build_dataset(self._val_indices)
                self.train_windows = self.train_dataset.windows
                self._train_preprocessed = self.train_dataset.preprocessed_data
                self.val_windows = self.val_dataset.windows
                self._val_preprocessed = self.val_dataset.preprocessed_data
```
- I dont like the idea of attaching windows/preprocessed version to data module. I think having them attached to dataset is enough, tests should obtain windows via `dm.train_dataset.windows` . wdyt?

Issue #2314, opened 2026-06-17
[BUG] Notebook tests failing on main


phoeenniixx:
**Describe the bug**

The tests for notebooks are failing on main, this happened due to recent releases in `matplotlib` and `sphinx`. I think we should fix them asap.

See the runs here: https://github.com/sktime/pytorch-forecasting/actions

The [last run that worked](https://github.com/sktime/pytorch-forecasting/actions/runs/27498970099/job/81278225731#step:5:143) (on June 14) had same `matplotlib` version as the [first run that failed ](https://github.com/sktime/pytorch-forecasting/actions/runs/27575387871/job/81522422718#step:5:143) (on June 15)
The problem is most probably the `pydata-sphinx-theme` version -  (`0.18.0` and `0.19.0`)

**Solution**

Solution is I think straightforward - remove `fig.show()`. Although I am not sure as I have not diagnosed it in detail.

Need to make sure if that is actually the case

**Additional context**

See the discord discussion: https://discord.com/channels/1075852648688930887/1516429302420476075

fkiraly:
seems to be `matplotlib`, according to #2135? Let´s merge the bound for now, and keep this open until we understand how to establish compatibility.

phoeenniixx:
> seems to be `matplotlib`, according to [#2135](https://github.com/sktime/pytorch-forecasting/pull/2135)? Let´s merge the bound for now, and keep this open until we understand how to establish compatibility.

Yes, I am confused why is this true? then why was the CI passing till june 14 ( `matplotlib` was released on June 11)?

I thought `sphinx` was the reason, but it didnt make any difference

phoeenniixx:
Ok found the reason, actually #2305 actually changed the tutorials to pass the tests. I directly looked at the actions (and not the PR) and I assumed the tutorials were not changed. This issue has been here since the release but I didnt notice it

PR #2315, opened 2026-06-17
[MNT] add an upper bound on `matplotlib` to prevent failing notebooks

phoeenniixx:
This PR adds an upper bound to  `matplotlib` to prevent failing notebooks on main.
See #2314


fkiraly (changes requested):
This is an upper bound on `matplotlib` though, not sphinx

PR #2316, opened 2026-06-18
[ENH] Implement DeepAR v2 model and package wrapper

### LINKED ISSUE #2007: [ENH] Add v2 interface support for DeepAR

harshsomankar123-tech:
#### Reference Issues/PRs
Fixes #2007.

This PR adds a v2-compliant implementation of the DeepAR model and its package wrapper, adhering to the new `EncoderDecoderTimeSeriesDataModule` and `Base_pkg` pattern (matching the structure used for TFT, DLinear, SAMformer, and TIDE).

Unlike the v1 implementation, this v2 model inherits directly from `BaseModel` (v2 LightningModule base) and implements autoregressive decoding, distribution mapping, and input target scaling in a clean, self-contained manner.

**Key Changes:**

1. **DeepAR v2 Model** (`pytorch_forecasting/models/deepar/_deepar_v2.py`)
   * Extends the new `BaseModel` interface.
   * Implements dynamic output projections compatible with `DistributionLoss` (e.g., `NormalDistributionLoss`).
   * Features an internal scaling mechanism that divides target inputs by `target_scale` before feeding them into the RNN to maintain numeric stability.
   * Implements a `rescale_prediction` method to project the model's parameters back into the 4-channel format (`[loc_scaler, scale_scaler, loc, scale]`) expected by PyTorch Forecasting distribution metrics in real space.
   * Supports teacher-forced decoding during training/validation/testing steps and autoregressive decoding during inference.

2. **DeepAR v2 Package Wrapper** (`pytorch_forecasting/models/deepar/_deepar_pkg_v2.py`)
   * Inherits from `Base_pkg` to enable high-level `.fit()` and `.predict()` APIs.
   * Configures package tags indicating capability support (e.g., exogenous variables, prediction intervals, etc.).
   * Registers test configurations with `NormalDistributionLoss` in `get_test_train_params` for testing different configurations (e.g., LSTM vs. GRU).

3. **Module Registration** (`pytorch_forecasting/models/deepar/__init__.py`)
   * Exports `DeepAR_pkg_v2` for discovery by the skbase estimator registry.

* The structural alignment with the new `BaseModel` and `Base_pkg` interface.
* The correctness of the internal scaling mechanism for RNN inputs.
* The `rescale_prediction` method and its compatibility with the expected 4-channel format for distribution metrics.
* The logic handling autoregressive decoding during inference versus teacher-forced decoding during training/validation.

Yes, the implementation was tested against both the new V2 suite and the existing V1 regression suite. All integration and unit tests for the new v2 package were run and passed successfully.

**1. V2 Estimator Tests (`test_all_estimators_v2.py`)**
* `test_doctest_examples[DeepAR]` ➔ PASSED
* `test_integration[DeepAR-0 / 1 / 2]` ➔ PASSED
* `test_checkpointing[DeepAR-0 / 1 / 2]` ➔ PASSED
* `test_predict_modes[DeepAR-0 / 1 / 2]` ➔ PASSED
* `test_pkg_linkage[DeepAR-DeepAR]` ➔ PASSED

**2. V1 Regression Tests (`test_deepar.py`)**
* All 17 existing tests ➔ PASSED (ensuring zero regression to the existing v1 codebase).

#### Any other comments?
N/A

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
+++ b/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
@@ -48,11 +48,15 @@ class DeepAR_pkg_v2(Base_pkg):
             Parameters to create testing instances of the class.
             Each dict contains model config and optionally datamodule_cfg.
         """
-        from pytorch_forecasting.metrics import NormalDistributionLoss
+        from pytorch_forecasting.metrics import (
+            LogNormalDistributionLoss,
+            NormalDistributionLoss,
+        )
 
         params = [
             dict(loss=NormalDistributionLoss()),
             dict(loss=NormalDistributionLoss(), cell_type="GRU"),
+            dict(loss=LogNormalDistributionLoss()),
             dict(
                 loss=NormalDistributionLoss(),
                 hidden_size=16,
--- a/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
+++ b/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
@@ -1,0 +1,73 @@
+"""DeepAR v2 package container."""
+
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class DeepAR_pkg_v2(Base_pkg):
+    """DeepAR v2 package container.
+
+    Wraps the v2 DeepAR model with the ``EncoderDecoderTimeSeriesDataModule``
+    data pipeline. Provides high-level ``fit`` and ``predict`` API through
+    ``Base_pkg``.
+    """
+
+    _tags = {
+        "info:name": "DeepAR",
+        "info:compute": 3,
+        "authors": ["jdb78", "harshsomankar123-tech"],
+        "capability:exogenous": True,
+        "capability:multivariate": False,
+        "capability:pred_int": True,
+        "capability:flexible_history_length": True,
+        "capability:cold_start": False,
+    }
+
+    @classmethod
+    def get_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models.deepar._deepar_v2 import DeepAR
+
+        return DeepAR
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data.data_module import (
+            EncoderDecoderTimeSeriesDataModule,
+        )
+
+        return EncoderDecoderTimeSeriesDataModule
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Return testing parameter settings for the trainer.
+
+        Returns
+        -------
+        params : list of dict
+            Parameters to create testing instances of the class.
+            Each dict contains model config and optionally datamodule_cfg.
+        """
+        from pytorch_forecasting.metrics import NormalDistributionLoss
+
+        params = [
+            dict(loss=NormalDistributionLoss()),
+            dict(loss=NormalDistributionLoss(), cell_type="GRU"),
+            dict(
+                loss=NormalDistributionLoss(),
+                hidden_size=16,
+                rnn_layers=1,
+            ),
+        ]
+
+        default_dm_cfg = {
+            "max_encoder_length": 4,
+            "max_prediction_length": 3,
+        }
+
+        for param in params:
+            current_dm_cfg = param.get("datamodule_cfg", {})
+            merged = {**default_dm_cfg, **current_dm_cfg}
+            param["datamodule_cfg"] = merged
+
+        return params
--- a/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
+++ b/pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
@@ -55,12 +55,14 @@ class DeepAR_pkg_v2(Base_pkg):
 
         params = [
             dict(loss=NormalDistributionLoss()),
-            dict(loss=NormalDistributionLoss(), cell_type="GRU"),
-            dict(loss=LogNormalDistributionLoss()),
+            dict(loss=NormalDistributionLoss(), cell_type="GRU", dropout=0.2),
+            dict(loss=LogNormalDistributionLoss(), optimizer="adamw"),
             dict(
                 loss=NormalDistributionLoss(),
                 hidden_size=16,
                 rnn_layers=1,
+                optimizer="sgd",
+                optimizer_params=dict(momentum=0.9),
             ),
         ]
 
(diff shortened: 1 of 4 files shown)


harshsomankar123-tech:
@phoeenniixx @fkiraly @PranavBhatP , PTAL. This PR is ready for review. 
Thanks!

pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Wraps the v2 DeepAR model with the ``EncoderDecoderTimeSeriesDataModule``
+    data pipeline. Provides high-level ``fit`` and ``predict`` API through
+    ``Base_pkg``.
+    """
+
+    _tags = {
+        "info:name": "DeepAR",
+        "info:compute": 3,
+        "authors": ["jdb78", "harshsomankar123-tech"],
```
phoeenniixx, line 17: I have a doubt: is this implementation taken from v1? It would be great if that is true!

I mean i think right now it is hard to move the models implementations (especially models like `DeepAR`) from v1 to v2 with minimal code changes.

pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
```diff
@@ -0,0 +1,73 @@
...
+        -------
+        params : list of dict
+            Parameters to create testing instances of the class.
+            Each dict contains model config and optionally datamodule_cfg.
+        """
+        from pytorch_forecasting.metrics import NormalDistributionLoss
+
+        params = [
+            dict(loss=NormalDistributionLoss()),
+            dict(loss=NormalDistributionLoss(), cell_type="GRU"),
```
phoeenniixx: could we try other losses as well? like other distribution losses

phoeenniixx (changes requested):
Thanks!
Few comments above
Also please add the API reference of the model here: https://github.com/sktime/pytorch-forecasting/blob/main/docs/source/m_layer_v2.rst#api-reference

pytorch_forecasting/models/deepar/_deepar_v2.py
```diff
@@ -0,0 +1,513 @@
...
+
+        # Last encoder target as seed for decoder
+        target_past = x["target_past"]
+        if target_past.dim() == 2:
+            target_past = target_past.unsqueeze(-1)
+        last_target = target_past[:, -1:, :]
+
+        # === DECODER ===
+        if y is not None:
+            # Teacher forcing: use ground-truth targets
```
agobbifbk, line 405: I have a comment here: this Teacher forcing is on-off or is it scheduled with some probability? I mean, y will be always not None during the training, while in inference it will be always None. It can be that the decoder maps y to the target and you have accurate prediction in training-validation but the REAL prediction sucks because `_decode_autoregressive` has not be trained!

pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
harshsomankar123-tech: @phoeenniixx Yes, this is essentially a direct port of the v1 DeepAR model adapted to work with the new v2 interface.
The core functionality remains the same as in v1, including the autoregressive recurrent network, target scaling, and probabilistic distribution-based predictions. The implementation has been refactored and simplified into a cleaner, self-contained class so it integrates naturally with the new v2 BaseModel and DataModule architecture while preserving the original DeepAR behavior.

pytorch_forecasting/models/deepar/_deepar_v2.py
harshsomankar123-tech: Yes @agobbifbk , it's purely on-off (100% teacher forcing in training, autoregressive in inference). This is how the original DeepAR paper (see:https://arxiv.org/abs/1704.04110) and v1 work, so I kept it the same for parity.
To stop it from cheating, the target input is shifted by 1 step (so at step t it only sees target at t-1). We could look into scheduled sampling later, but wanted to keep it consistent with v1 for now.

pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
```diff
@@ -0,0 +1,77 @@
...
+            Parameters to create testing instances of the class.
+            Each dict contains model config and optionally datamodule_cfg.
+        """
+        from pytorch_forecasting.metrics import (
+            LogNormalDistributionLoss,
+            NormalDistributionLoss,
+        )
+
+        params = [
+            dict(loss=NormalDistributionLoss()),
```
phoeenniixx, line 57: we shoul pass other params as well no?

pytorch_forecasting/models/deepar/_deepar_v2.py
```diff
@@ -0,0 +1,522 @@
...
+        else:
+            # Autoregressive: use predicted means
+            prediction = self._decode_autoregressive(x, hidden_state, last_target)
+
+        # Rescale prediction
+        prediction = self.rescale_prediction(prediction, x["target_scale"])
+
+        return {"prediction": prediction}
+
+    def training_step(
```
phoeenniixx, line 419: do we need these step methods? I thought `BaseModel` should've handled these?

harshsomankar123-tech: Yes, these overrides are required for DeepAR. Since DeepAR is an autoregressive model, it relies on teacher forcing during training, validation, and testing, which means the ground-truth target y needs to be passed to the decoder.

The default step methods in BaseModel only call self(x), leaving y=None. As a result, the forward pass falls back to autoregressive decoding, which is slower and can lead to less stable behavior. By overriding the step methods and calling self(x, y=y), we ensure the model follows the intended teacher-forcing path.

I also updated the overridden step methods to call self.log_metrics(y_hat, y, prefix=...), so all user-configured evaluation metrics are logged correctly during training, validation, and testing, keeping the behavior consistent with BaseModel.PTAL ad let me know ur thoughts??

pytorch_forecasting/models/deepar/_deepar_pkg_v2.py
harshsomankar123-tech: @phoeenniixx Thanks!,DONE

pytorch_forecasting/models/deepar/_deepar_v2.py
harshsomankar123-tech: @phoeenniixx @agobbifbk @fkiraly , could you please take a look when you have some time.
 Thanks!

PR #2321, opened 2026-06-20
[BUG] Fix NameError in NHiTS with static real covariates

### LINKED ISSUE #2281: [BUG] NHiTS fails when TimeSeriesDataSet includes static real variables

andersendsa:
#### Reference Issues/PRs
Closes: #2281

When configuring NHiTS with a TimeSeriesDataSet that includes static variables (e.g. `add_target_scales=True`), but no time-varying encoder covariates, the model incorrectly skips initializing `encoder_features` while still attempting to access it later during the forward pass. This causes a NameError.

This commit updates the `forward` method of NHiTS to extract features when *either* `encoder_covariate_size` or `static_size` is strictly greater than 0. `encoder_x_t` construction is safely kept gated behind only `encoder_covariate_size > 0` to preserve expected shapes.

A test case has been added to test_nhits.py to prevent regressions.

- The condition change in pytorch_forecasting/models/nhits/_nhits.py.
- Ensuring that decoupling the extraction of encoder_features from the concatenation into encoder_x_t cleanly maintains expected dimensions for both static and time-varying branches.

test_static_covariates_no_encoder_covariates to tests/test_models/test_nhits.py. The test uses a minimal reproducible configuration with add_target_scales=True and no time-varying covariates to verify the forward pass completes without throwing a NameError.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/nhits/_nhits.py
+++ b/pytorch_forecasting/models/nhits/_nhits.py
@@ -358,10 +358,12 @@ class NHiTS(BaseModelWithCovariates):
             output of model
         """
         # covariates
-        if self.encoder_covariate_size > 0:
+        if self.encoder_covariate_size > 0 or self.static_size > 0:
             encoder_features = self.extract_features(
                 x, self.embeddings, period="encoder"
             )
+
+        if self.encoder_covariate_size > 0:
             encoder_x_t = torch.concat(
                 [
                     encoder_features[name]
--- a/tests/test_models/test_nhits.py
+++ b/tests/test_models/test_nhits.py
@@ -216,3 +216,27 @@ def test_prediction_length(max_prediction_length: int):
         return_index=True,
         return_decoder_lengths=True,
     )
+
+
+def test_static_covariates_no_encoder_covariates():
+    from pytorch_forecasting.data.examples import generate_ar_data
+
+    data = generate_ar_data(seasonality=10.0, timesteps=400, n_series=10, seed=42)
+    dataset = TimeSeriesDataSet(
+        data,
+        time_idx="time_idx",
+        target="value",
+        group_ids=["series"],
+        time_varying_unknown_reals=["value"],
+        max_encoder_length=60,
+        max_prediction_length=20,
+        add_target_scales=True,  # <-- introduces static real variables
+    )
+
+    dataloader = dataset.to_dataloader(train=True, batch_size=8)
+    model = NHiTS.from_dataset(dataset)
+
+    batch, _ = next(iter(dataloader))
+
+    # Should not crash with NameError
+    model(batch)


andersendsa:
Hi @phoeenniixx  this pr passes all the tests and js ready for review

PR #2323, opened 2026-06-23
[ENH]  Add implementation of `save` and `load` to v2

phoeenniixx:
This PR adds a bare bone implementation of `save` and `load` in `Basepkg`

We still need to decide if this is a right way forward

### Changes

- Adds a `ArtifactRegistryCallback` for updating the `artifacts.yaml` with the new model chekcpoints in the training.
- Adds `_save` and `load` to `BasePkg`. 
    - We keep save a private method that would be called by `fit`.
    - `load` is a public method for now.

### Note for the reviewers
To see how the methods work, please look at the comments, TODOs and docstrings of the methods


fkiraly (commented):
I think this is the right direction (as discussed) - of course still quite bare-bones

pytorch_forecasting/base/_base_pkg.py
```diff
@@ -178,19 +175,253 @@ def _load_dataloader(
...
+        set by ``_init_empty_state``.
+        """
+        for name in self._CFG_KEYS:
+            path = artifacts.get(name)
+            if path is None:
+                continue
+            with open(path, "rb") as f:
+                setattr(self, name, pickle.load(f))  # noqa: S301
+
+    def _load_metadata_from_artifacts(self, artifacts: dict[str, Any]) -> None:
```
phoeenniixx, line 390: Clarfication: We can't create the datamodule now because it requires data at `__init__`. Instead, store the artifact paths and inject them when the datamodule is created (in `fit()` or `predict()`).

pytorch_forecasting/base/_base_pkg.py
```diff
@@ -178,19 +175,248 @@ def _load_dataloader(
...
+    def _populate_from_artifacts(self, artifacts: dict[str, Any]) -> None:
+        """Populate instance state from a parsed artifacts dictionary."""
+        self._load_cfgs_from_artifacts(artifacts)
+        self._load_metadata_from_artifacts(artifacts)
+        self._init_trainer_from_cfg()
+        self._store_pending_dm_artifacts(artifacts)
+        self._load_model_from_artifacts(artifacts)
+
+    def _load_cfgs_from_artifacts(self, artifacts: dict[str, Any]) -> None:
+        """Load configuration dictionaries from artifacts."""
```
phoeenniixx, line 382: Only loads keys that are present in ``artifacts``. missing keys (e.g., excluded at save time) are left as the empty defaults set by ``_init_empty_state``.

phoeenniixx:
Not tested - still need to update the datamodule and finally tests!

phoeenniixx:
I think this is ready for an initial review - is this what we want?
FYI @fkiraly @agobbifbk

pytorch_forecasting/base/_base_pkg.py
```diff
@@ -215,6 +447,10 @@ def fit(
             Directory to save artifacts.
         ckpt_kwargs : dict, optional
```
phoeenniixx, line 449: Should we add the link to documentation of `ModelCheckpoint` here, or, should we copy and add the whole docstring of the callback here itself? 

If second option - we need to make sure it doesn't make the docs "awkwardly long" and somehow highlight that this docstring is for kwargs of this callback and not something that should be passed as params, but as a `dict`...

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -568,6 +571,154 @@ def __getitem__(self, idx):
 
             return x, y
 
+    def save_artifacts(
+        self, artifact_dir: Path, include: list[str] = [], exclude: list[str] = []
```
phoeenniixx, line 819: shouuld we use `include`? Or `exclude` is enough?

pytorch_forecasting/base/_base_pkg.py
```diff
@@ -48,29 +56,22 @@ def __init__(
...
+
+        self.metadata = {}
         self.model = None
         self.trainer = None
         self.datamodule = None
-        if self.ckpt_path:
-            print(self.metadata)
-            self._build_model(metadata=self.metadata, **self.model_cfg)
-        else:
-            self.model = None
```
Faakhir30, line 73: I think this PR is no longer loading model from checkpoint at init or `fit` autmatically, if `ckpt_path` is provided at init time. 

Is this intentional? and users are required to use `load` explicitly to load from the checkpointed version.
Closer to how most HF models have `load_from_checkpoint`

phoeenniixx: actually yes!
After long discussions, we arrived at the decision that it would be best to keep `load` as a separate utility :)

fkiraly (changes requested):
While it goes in the right direction, I think this is not reliably possible to review without a list of key vignettes here, with intended state change and file changes clearly listed out.

I would also suggest to add these, and include in the user testing session later this week.

phoeenniixx:
Hi @fkiraly, I think the vignettes (and design) follows the one proposed here: https://github.com/sktime/enhancement-proposals/pull/51

Should I add those here as well in the PR?

phoeenniixx:
I will add a new nb for the user testing session covering the key vignettes for `save` and `load`

PR #2324, opened 2026-06-24
[ENH] Add multiloss support to `ptf` v2

### LINKED ISSUE #2308: [ENH] Add `MultiLoss` and multi-target support to `ptf-v2`
Currently, the `ptf-v2` pipeline doesnot support multi-target forecasts and it is hard to use `MultiLoss` here. We should add the support to the package.

phoeenniixx:
This PR adds support for `MultiLoss` to `ptf` v2

Fixes #2308
### Changes
- [x] Add `weights` to the `collate_fn` and `__getitem__` (of private dataset class of datamodule) return for `MultiLoss`. The return would be `(x, (y, weight))` in place of `(x,y)`.
    Done to:
    - [x] `EncoderDecoderTimeSeriesDataModule`
    - [x] `TslibDataModule`
- [x]  Add support to `BaseModel`. 
- [ ] Add tests and update the test framework - maybe should happen in a new PR instead - along with fixes if needed?

### DIFF
--- a/docs/source/tutorials/ptf_V2_example.ipynb
+++ b/docs/source/tutorials/ptf_V2_example.ipynb
@@ -1747,8 +1747,8 @@
    "source": [
     "print(\"First Predicted Value:\")\n",
     "print(\"Index:\", preds[\"index\"][0].item())\n",
-    "print(\"Prediction:\", preds[\"prediction\"][0].item())\n",
-    "print(\"Actual:\", preds[\"y\"][0].item())"
+    "print(\"Prediction:\", preds[\"prediction\"][0][0].item())\n",
+    "print(\"Actual:\", preds[\"y\"][0][0].item())"
    ]
   }
  ],
--- a/docs/source/tutorials/ptf_V2_example.ipynb
+++ b/docs/source/tutorials/ptf_V2_example.ipynb
@@ -1039,8 +1039,8 @@
    "source": [
     "print(\"First Predicted Value:\")\n",
     "print(\"Index:\", preds[\"index\"][0].item())\n",
-    "print(\"Prediction:\", preds[\"prediction\"][0].item())\n",
-    "print(\"Actual:\", preds[\"y\"][0].item())"
+    "print(\"Prediction:\", preds[\"prediction\"][0][0].item())\n",
+    "print(\"Actual:\", preds[\"y\"][0][0].item())"
    ]
   },
   {
--- a/pytorch_forecasting/models/base/_base_model_v2.py
+++ b/pytorch_forecasting/models/base/_base_model_v2.py
@@ -160,18 +160,63 @@ class BaseModel(LightningModule):
 
         return predict_callback.result
 
-    def to_prediction(self, out: dict[str, Any], **kwargs) -> torch.Tensor:
-        """Converts raw model output to point forecasts."""
-        # todo: add MultiLoss support
+    def to_prediction(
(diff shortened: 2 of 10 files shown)


phoeenniixx:
Tests will fail fs
Still need to add support to `tslib` modules

docs/source/tutorials/ptf_V2_example.ipynb
```diff
@@ -1039,8 +1039,8 @@
    "source": [
     "print(\"First Predicted Value:\")\n",
     "print(\"Index:\", preds[\"index\"][0].item())\n",
-    "print(\"Prediction:\", preds[\"prediction\"][0].item())\n",
-    "print(\"Actual:\", preds[\"y\"][0].item())"
+    "print(\"Prediction:\", preds[\"prediction\"][0][0].item())\n",
```
phoeenniixx, line 1042: Previously, we printed the very first item from the batch of predictions that is why it was `preds["predictions"][0]`, but now as we return `list` of prediction is all cases, to access the first item, it would be `preds["predictions"][0][0]`

pytorch_forecasting/tests/test_all_v2/_test_integration.py
```diff
@@ -23,7 +23,7 @@ def _integration(
     assert isinstance(predictions, dict)
     assert "prediction" in predictions
 
-    pred_tensor = predictions["prediction"]
+    pred_tensor = predictions["prediction"][0]
```
phoeenniixx, line 26: Now for single target as well we return a list, so this has to be updated - did the same for rest of the tests!

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -160,107 +160,118 @@ def predict(
...
+                return pred_out if isinstance(pred_out, (list, tuple)) else [pred_out]
+        if kwargs:
+            pred_out = self.loss.to_quantiles(pred_input, **kwargs)
+        else:  # in case passed kwargs do not exist
+            pred_out = self.loss.to_quantiles(pred_input)
+        return pred_out if isinstance(pred_out, (list, tuple)) else [pred_out]
+
+    def _coerce_targets_for_loss(self, y):
         """
+        Coerce target outputs to match loss function expectations.
```
phoeenniixx: The `DataModule` always returns a list of tensors (one per target), but metrics/losses expect a single tensor for single-target tasks.

tests/test_data/test_data_module.py
```diff
@@ -491,8 +493,8 @@ def test_multivariate_target(normalizer_list):
 
     x, y = dm.train_dataset[0]
     assert len(y) == 2
-    assert y[0].shape == (dm.max_prediction_length,)
-    assert y[1].shape == (dm.max_prediction_length,)
+    assert y[0][0].shape == (dm.max_prediction_length,)
```
phoeenniixx, line 496: `y` is a tuple of `(Y,weights)` and where `Y` is the target and `weights` are the weights for the target, here, `weights` will be `None` if not specified, so the tuple will be `(Y,None)`. 
And `Y` is a `list` of all targets, in case of single target, it will be a list of a single item
So, to access the first target, it will be `y[0][0]`

phoeenniixx:
To maintain the homogenity, now input and output (`y` and predictions) both are returned as lists for all cases. Keeping `y` as a list and having a case distinction in prediction (tensor for single target and list for multi-target) didnt make sense to me :)

PR #2327, opened 2026-06-29
[ENH] refactor _loss_mapping to tags of metrics, entirely

### LINKED ISSUE #1957: [ENH] refactor `_loss_mapping` to tags of metrics, entirely

Faakhir30:
#### Reference Issues/PRs

Closes #1957 

- refactor `_loss_mapping.py` to tags
- get compatible losses by filtering `all_objects` of type `metric` and specific `metric_type` based on `pred_type` and `y_type` 
- move test kwargs for losses to metric pkg level
- introduced method `get_test_train_params` for this

#### PR checklist

- [x] Added/modified tests


Faakhir30:
~only updated `Deeper` model and `BetaDistributionLoss` pkg for now.~
**TODO:** Will update remaining metric pkgs and model pkgs, after approval of design by maintainers 

cc @phoeenniixx @fkiraly 

**Edit:** Done, ready for revview

pytorch_forecasting/tests/test_all_estimators.py
```diff
@@ -195,14 +221,10 @@ def _generate_final_param_list(self, compatible_losses, base_params_list):
...
-                loss_name = loss_item.__name__
-                loss = loss_item
-            else:
-                loss_name = loss_item.__class__.__name__
-                loss = loss_item
-            loss_params = deepcopy(LOSS_SPECIFIC_PARAMS.get(loss_name, {}))
+        for loss_pkg in compatible_loss_pkgs:
+            loss = loss_pkg.get_cls()()
+            loss_name = loss.__class__.__name__
+            loss_params = deepcopy(loss_pkg.get_default_params())
```
phoeenniixx, line 213: should we add a way to pass some other "params" to losses here?
I think that would make it overcomplicated no? 
What do you think @Faakhir30 @fkiraly ?

pytorch_forecasting/metrics/_distributions_pkg/_multivariate_normal/_multivariate_normal_distribution_loss_pkg.py
```diff
@@ -26,3 +26,15 @@ def get_cls(cls):
         )
 
         return MultivariateNormalDistributionLoss
+
+    @classmethod
+    def get_default_params(cls):
```
phoeenniixx, line 31: It would be good to add a docstring here. Otherwise it feels like these params are used as fixtures for tests of metrics, while it is mainly used by the test-framework for the fixtures of tests for models

phoeenniixx (changes requested):
Nice!
I think it is ready! Just a small doubt and a suggestion (see above)
Please add docstrings to all public methods atleast - to prevent any confusions on what the methods do!

pytorch_forecasting/tests/test_all_estimators.py
Faakhir30: I think a solution will be to have some method that has dict of `losses: loss_kwargs ` at model pkg level and testing framework override for that particular model+loss combination.


Currently I'm unable to see any such case, we can add it later if any such case arrive.

fkiraly (changes requested):
Great work!

One small thing that I do not like is that `get_compatible_losses` is now moved into the `_get_compatible_loss_pkgs_for_model`, this feels like the wrong location for this.

I would move this to a private method in the base-model-package, `_get_compatible_loss_types`, which returns a list of str. That method is then called in the test, this way we separate concerns about "describe properties of the model" (a method of the model pkg), and "generate all losses" (stays `_get_compatible_loss_pkgs_for_model`)

phoeenniixx (approved):
Nice! This is good!
But before merging, can @fkiraly please confirm if he is also happy?

PR #2331, opened 2026-07-03
&Nidhicodes [ENH] Add support for nn losses to ptf-v2 

### LINKED ISSUE #1970: [ENH] Add support for `nn` losses to `ptf-v2`

Faakhir30:
#### Reference Issues/PRs

Fixes #1970

Followups and stacks on existing PR #2073 by @Nidhicodes

By @Nidhicodes:
- Introduces NNLossAdapter, a lightweight wrapper that adapts nn.Module losses to the ptf-v2 loss/metric API.
- Enables users to pass native PyTorch losses directly to models (e.g. DLinear(loss=nn.MSELoss())) without manual wrapping.
- Ensures compatibility with BaseModel.predict() by implementing to_prediction and to_quantiles.

The adapter:
- Handles (target, weight) inputs by applying weights internally while respecting the loss’s original reduction.
- Supports multi-target predictions by splitting [B, T, N] tensors and summing losses across targets.
- Enforces point-prediction-only usage (H=1) for losses that are not horizon-aware, with clear error messages otherwise.

Further updates:
- store in `self._loss` instead of `self.loss` 
- merge main
- integration tests with models
- fix for `tide` and other models where needed
- add support for non-point losses

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
+++ b/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
@@ -267,13 +267,11 @@ def coerce_to_pytorch_forecasting_metric(
     Metric or torch.nn.Module
         Loss/metric usable in ptf training (``forward(y_pred, y_actual)``).
     """
-    if isinstance(metric, Metric | MultiLoss | CompositeMetric):
-        return metric
-
     from pytorch_forecasting.metrics.nn_loss_adapter import NNLossAdapter
 
-    if isinstance(metric, NNLossAdapter):
+    if isinstance(metric, (Metric, MultiLoss, CompositeMetric, NNLossAdapter)):
         return metric
+
     # bare torch.nn loss (nn.Module, but not a torchmetrics Metric)
     if isinstance(metric, torch.nn.Module) and not isinstance(metric, LightningMetric):
         return NNLossAdapter(metric)
--- a/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
+++ b/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
@@ -245,23 +245,39 @@ class TorchMetricWrapper(Metric):
         return f"WrappedTorchmetric({repr(self.torchmetric)})"
 
 
-def convert_torchmetric_to_pytorch_forecasting_metric(
-    metric: LightningMetric,
-) -> Metric:
+def coerce_to_pytorch_forecasting_metric(
+    metric: LightningMetric | torch.nn.Module,
+) -> Metric | torch.nn.Module:
     """
-    If necessary, convert a torchmetric to a PyTorch Forecasting metric that
-    works with PyTorch Forecasting models.
+    Coerce a loss or metric into something usable by PyTorch Forecasting models.
 
-    Args:
-        metric (LightningMetric): metric to (potentially) convert
+    Accepts ptf metrics, plain ``torch.nn`` losses, and torchmetrics metrics.
 
-    Returns:
-        Metric: PyTorch Forecasting metric
+    * Already a ptf ``Metric`` / ``MultiLoss`` / ``CompositeMetric`` → unchanged
+    * Plain ``torch.nn`` loss module (e.g. ``nn.MSELoss()``) → ``NNLossAdapter``
+    * Other torchmetrics ``Metric`` → ``TorchMetricWrapper``
+
+    Parameters
+    ----------
+    metric :
+        Metric or loss to (potentially) adapt.
+
+    Returns
+    -------
+    Metric or torch.nn.Module
+        Loss/metric usable in ptf training (``forward(y_pred, y_actual)``).
     """
-    if not isinstance(metric, Metric | MultiLoss | CompositeMetric):
-        return TorchMetricWrapper(metric)
-    else:
+    if isinstance(metric, Metric | MultiLoss | CompositeMetric):
         return metric
+
+    from pytorch_forecasting.metrics.nn_loss_adapter import NNLossAdapter
+
+    if isinstance(metric, NNLossAdapter):
+        return metric
+    # bare torch.nn loss (nn.Module, but not a torchmetrics Metric)
+    if isinstance(metric, torch.nn.Module) and not isinstance(metric, LightningMetric):
+        return NNLossAdapter(metric)
+    return TorchMetricWrapper(metric)
 
 
 class MultiLoss(LightningMetric):
@@ -286,9 +302,7 @@ class MultiLoss(LightningMetric):
             metrics
         ), "Number of weights has to match number of metrics"
 
-        self.metrics = [
-            convert_torchmetric_to_pytorch_forecasting_metric(m) for m in metrics
-        ]
+        self.metrics = [coerce_to_pytorch_forecasting_metric(m) for m in metrics]
         self.weights = weights
 
         super().__init__()
--- a/pytorch_forecasting/metrics/nn_loss_adapter.py
+++ b/pytorch_forecasting/metrics/nn_loss_adapter.py
@@ -1,5 +1,3 @@
-from typing import List, Optional, Tuple, Union
-
 import torch
 import torch.nn as nn
 
@@ -21,23 +19,27 @@ class NNLossAdapter(nn.Module):
 
     def forward(
         self,
-        y_pred: Union[torch.Tensor, List[torch.Tensor]],
-        y_actual: Union[torch.Tensor, Tuple[Union[torch.Tensor, List[torch.Tensor]], torch.Tensor]],
+        y_pred: torch.Tensor | list[torch.Tensor],
+        y_actual: torch.Tensor | tuple[torch.Tensor | list[torch.Tensor], torch.Tensor],
     ) -> torch.Tensor:
         """
         Forward pass of the adapter.
 
         Args:
-            y_pred (Union[torch.Tensor, List[torch.Tensor]]): Model predictions.
-                Expected to be [B, T, N] for multi-target or [B, T, 1] / [B, T] for single target.
-            y_actual (Union[torch.Tensor, Tuple]): Actual values and optionally weights.
-                Can be a tensor, or a tuple (target, weight), where target can be a list of tensors.
+            y_pred (torch.Tensor | list[torch.Tensor]): Model predictions.
+                Expected to be [B, T, N] for multi-target or
+                [B, T, 1] / [B, T] for single target.
+            y_actual (torch.Tensor | tuple): Actual values and optionally weights.
+                Can be a tensor, or a tuple (target, weight), where target
+                can be a list of tensors.
 
         Returns:
             torch.Tensor: The computed and reduced loss.
         """
         # Handle y_actual as (target, weight) or just target
-        if isinstance(y_actual, (list, tuple)) and not isinstance(y_actual, torch.Tensor):
+        if isinstance(y_actual, (list, tuple)) and not isinstance(
+            y_actual, torch.Tensor
+        ):
             if len(y_actual) == 2:
                 target, weight = y_actual
             else:
@@ -50,9 +52,9 @@ class NNLossAdapter(nn.Module):
             # Multi-target scenario
             if not isinstance(y_pred, torch.Tensor):
                 raise ValueError(
-                    f"NNLossAdapter expected y_pred to be a torch.Tensor for multi-target, "
-                    f"but got {type(y_pred)}. Standard multi-target in ptf-v2 expects "
-                    f"y_pred of shape [B, T, N]."
+                    f"NNLossAdapter expected y_pred to be a torch.Tensor for "
+                    f"multi-target, but got {type(y_pred)}. Standard multi-target "
+                    f"in ptf-v2 expects y_pred of shape [B, T, N]."
                 )
 
             # y_pred is [B, T, N], split along last dimension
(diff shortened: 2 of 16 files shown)


phoeenniixx (changes requested):
Can we try this adpater with some already implemented methods (optimally all of them, but that is hard, ig we need to update our test framework so that it loops over the losses like we do for v1)

Faakhir30:
> Can we try this adpater with some already implemented methods (optimally all of them

I think you are referring to `get_test_train_params`?

> , but that is hard, ig we need to update our test framework so that it loops over the losses like we do for v1

v1 has `info:pred_type` tag that lets us get right metric/loss objects and them loop over them, I see that most v2 model pkgs are not specifying this tag (though the extension template suggests to add), looks intentional?

Even if I add that tag, are you suggesting to maintain all set of nn.losses like we had sets of metrics at `_loss_mapping` and fetch compatible ones, and loop over them for each model? 
That would mean: `num_of_nn_losses x len(get_test_train_params())` fixtures for each model_pkg.

I think it'll be better to just add a few nn losses to each `get_test_train_params`.

phoeenniixx:
Sorry I meant models, not methods

pytorch_forecasting/models/tide/_tide_dsipts/_tide_v2_pkg.py
```diff
@@ -55,6 +57,7 @@ def get_test_train_params(cls):
                 n_add_enc=1,
                 n_add_dec=1,
                 dropout_rate=0.1,
+                loss=nn.MSELoss(),
```
phoeenniixx, line 60: I have a question: as much as I remember, the `nn` losses expect different shapes right? Are all the `nn` losses supported by the adapter? Can you please add a test where you use `pytest.parameterize` to move through all the losses to see if they work correctly? I think the standlone test file `test_nn_loss_adapter` would be a good place for such a test, you can try 1-2 models to see if they work

Faakhir30: oh yeah, not all nn losses are supported. The adapter assumes a two-arg call `loss(y_pred, target)` where, after squeezing the last dim, both tensors are the same shape `[B, T]`

I'll try to improve the adapter to widen support of nn losses and have some parameterized tests

Faakhir30: wait a sec, im not sure if i've understood this correct.

> the nn losses expect different shapes right?

All point losses expect shape to be `[B, T]` i think. Above comment was supposed to refer class/non-point losses? like CrossEntropy?
I've added support of class based losses + GaussianNLLLoss in last commit.
Also added a standalone test for TFTModel to test with CE in `test_nn_loss_adapter`.

phoeenniixx: yes, i mean ALL `nn` losses, not just the point prediction ones. Ideally we should be able to handle this using jsut one adapter no? Having mulitple adpaters for a single module feels counterintuitive.

phoeenniixx: Although if it is hard to keep everything in one adpater, then we should create mixin classes or wrappers for the losses - not adapters

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def _prepare_inputs(
+        self,
+        y_pred: torch.Tensor,
+        target: torch.Tensor,
+        mode: _Mode,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        if mode == "point":
+            if y_pred.ndim == 3:
+                if y_pred.size(-1) != 1:
```
phoeenniixx: this too much indentation - these conditions could be in one `if`, no?

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,262 @@
...
+                    raise ValueError(
+                        f"NNLossAdapter only supports point predictions (H=1). "
+                        f"Got y_pred shape {list(y_pred.shape)} with "
+                        f"H={y_pred.size(-1)}. "
+                        "For multi-horizon losses, use a ptf metrics loss instead."
+                    )
+                y_pred = y_pred.squeeze(-1)
+            return y_pred, target
+
+        if mode == "class":
```
phoeenniixx: should we use `elif` here?

phoeenniixx (changes requested):
Can you please add docstrings in the adapter - without it, it is hard to review!
Also, what if a user uses this: `loss = Multiloss([MAPE,nn.MSELoss])`?
Should we add some way so that if we are even using a `ptf` loss with `nn` loss, this should work

pytorch_forecasting/metrics/nn_loss_adapter.py
Faakhir30: na, single  `if` is impossible imo, reduced it to two though ;)

Faakhir30:
> Also, what if a user uses this: loss = Multiloss([MAPE,nn.MSELoss])?
>Should we add some way so that if we are even using a ptf loss with nn loss, this should work

Ahh, I did think about this at start,
above call of `Multiloss([MAPE,nn.MSELoss])` suggests that multiloss should internally pass nn loss via adapter.

I was thinking if instead of taking this adapter's base from `nn.Module` and handling in metrics then, we base it from `MultiHorizonMetric`.
I do somewhat foresee the challanges it'll create, wrapping a stateless thing in a statefull metric API might be tough.

But that design automatically makes a nn loss interchangeable with any Metric losses without much friction.
User should be able to call `Metric.from_nn_loss(nn.MSELoss)` at anywhere a `Metric` is supported. 

I should have shared this intuition in PR description as alternative before ;) I did not point out initially, because on @Nidhicodes PR, I dont see any maintainers comment pointing above problem. Is there something I'm missing? is there some major issue with `Metric` approach?

phoeenniixx:
sorry, I dont exactly understand what your question is here, can you please rephrase it?

Faakhir30:
> sorry, I dont exactly understand what your question is here, can you please rephrase it?


I was suggesting an alternative design.

**Design alternative I'm inclined towards:** instead of `class NNLossAdapter(nn.Module)`, make it a real ptf metric, e.g. `class NNLossAdapter(MultiHorizonMetric)`

**Question:**
is there a strong reason to keep the adapter as a plain `nn.Module`, or treating nn losses as a first-class `Metric` route also viable?

I have pushed this newer design in a single commit for review, we can revert back to keeping adapter a `nn.Module` if needed.

pytorch_forecasting/metrics/base_metrics/_base_metrics.py
```diff
@@ -245,23 +245,39 @@ def __repr__(self):
...
-    if not isinstance(metric, Metric | MultiLoss | CompositeMetric):
-        return TorchMetricWrapper(metric)
-    else:
+    if isinstance(metric, Metric | MultiLoss | CompositeMetric):
+        return metric
+
+    from pytorch_forecasting.metrics.nn_loss_adapter import NNLossAdapter
+
+    if isinstance(metric, NNLossAdapter):
         return metric
```
phoeenniixx: these can be clubbed into one?

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,294 @@
...
+            return per_elem.view(batch_size, time_idx)
+        if per_elem.ndim == 0:
+            # defensive: some losses ignore reduction="none"
+            return per_elem.expand(batch_size, time_idx)
+        return per_elem
+
+    def update(self, y_pred, target):
+        """Update metric state; reject multi-target lists with a clear error."""
```
phoeenniixx: please improve the docstrings and add parameters for the public interfaces ATLEAST

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,294 @@
...
+        """Map a native loss class to the adapter reshape/call mode.
+
+        Parameters
+        ----------
+        loss :
+            Wrapped ``torch.nn`` loss module.
+
+        Returns
+        -------
+        {"point", "class", "gaussian_nll"}
```
phoeenniixx: it is one of these no? the return should be `str`?

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,294 @@
...
+                )
+            if target.ndim != 2:
+                raise ValueError(
+                    "Classification targets must have shape (batch, time), "
+                    f"got {tuple(target.shape)}."
+                )
+            return y_pred.reshape(-1, y_pred.size(-1)), target.reshape(-1).long()
+
+        # gaussian_nll
+        if y_pred.ndim != 3 or y_pred.size(-1) != 2:
```
phoeenniixx: should we add `else` here?

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,294 @@
...
+        -------
+        torch.Tensor
+            Loss as returned by the wrapped module (honoring its reduction,
+            unless temporarily overridden by :meth:`_compute_loss`).
+        """
+        if mode == "gaussian_nll":
+            mean = y_pred[..., 0]
+            var = F.softplus(y_pred[..., 1]) + 1e-6
+            return self._loss(mean, target, var)
+        if mode == "class" and isinstance(self._loss, nn.NLLLoss):
```
phoeenniixx: should we add `else` here?

phoeenniixx (changes requested):
Added some comments
- Please use `if-else` rather than multiple `if` statements if possible. This is not only useful to prevent looking at unnecessary `if` statements but also makes it clear for the reader that the statement below is related to the above `if` and not a totally different condition. In short, makes it easier to understand code

pytorch_forecasting/metrics/nn_loss_adapter.py
Faakhir30: it does nothing beside avoiding multi-target, replien on MHM update. sure, documented better. thanks for pointing out.

pytorch_forecasting/metrics/nn_loss_adapter.py
```diff
@@ -0,0 +1,329 @@
...
+        if "reduction" not in kwargs:
+            kwargs["reduction"] = "none" if nn_reduction == "none" else "mean"
+        if "name" not in kwargs or kwargs["name"] is None:
+            kwargs["name"] = f"NNLossAdapter({type(loss).__name__})"
+
+        super().__init__(**kwargs)
+
+        self._loss = copy.deepcopy(loss)
+        if hasattr(self._loss, "reduction"):
+            self._loss.reduction = "none"
```
phoeenniixx, line 57: why are we doing this?

pytorch_forecasting/models/base/_base_model_v2.py
```diff
@@ -63,15 +66,17 @@ class BaseModel(LightningModule):
...
         optimizer: Optimizer | str | None = "adam",
         optimizer_params: dict | None = None,
         lr_scheduler: str | None = None,
         lr_scheduler_params: dict | None = None,
     ):
         super().__init__()
-        self.loss = loss
+
+        # wrap bare nn losses; recurse into MultiLoss children via the converter
+        self.loss = coerce_to_pytorch_forecasting_metric(loss)
```
phoeenniixx: first collect in `self.loss` and then the changes should be made to `self._loss` similar to `sktime`

pytorch_forecasting/tests/test_nn_loss_adapter.py
```diff
@@ -0,0 +1,323 @@
...
+    elif isinstance(loss_fn, nn.PoissonNLLLoss):
+        y_pred = torch.randn(batch, time, 1)
+        target = torch.randint(0, 5, (batch, time), dtype=torch.float32)
+    else:
+        y_pred = torch.randn(batch, time, 1)
+        target = torch.randn(batch, time)
+    return y_pred, target
+
+
+def test_nn_loss_adapter_is_metric():
```
phoeenniixx: Is this needed? The adapter is child of these classes, so it is understood no?

phoeenniixx (changes requested):
Nice work!
I think it is almost ready - just a few suggestions (see above)

Also, as the adapter is a `Metric` child, should we add this to the tag system as well? As this is also "kinda" a metric?

pytorch_forecasting/metrics/nn_loss_adapter.py
Faakhir30: Because reduction is being handeled in MHM itself, I'm passing it in `super().__init__(**kwargs)`. Keeping default reduction (mean) attached to nn loss will result in error as loss func will return a scaler for classification losses too.

So, `self._loss.reduction="none"` keeps the wrapped `nn` loss in **elementwise** form so reshape, masking, and weighting can run, while metric base takes care of reduction.

Faakhir30:
> Also, as the adapter is a Metric child, should we add this to the tag system as well? As this is also "kinda" a metric?

Adapter is too generlized, so that will be a wrong abstraction, as we cant tell a single `metric_type`, `requires:data_type`, etc values. 
One option could be to break adapter into multiple classes such that each subclass can be tagged, but seems redundent to me. 
If testing is the main motivation behind this, standalone tests looks enough. lmk about your thoughts.

phoeenniixx (approved):
Thanks!

PR #2334, opened 2026-07-06
[ENH] add _tags.py to registry

### LINKED ISSUE #2333: [ENH] Add `_tags.py` to registry
Currently ptf documents tags at indivisual modules documentation. e.g.,`extension_templates/v2/README.md`

- This makes harder for searching of tags documentation (somewhat)
- breif documentations exist
- same tags being documented multiple places

Sktime repo maintains detailed documentation of tags in https://github.com/sktime/sktime/blob/main/sktime/registry/_tags.py.

Same approach is also used in skpro https://github.com/sktime/skpro/blob/main/skpro/registry/_tags.py

Tasks:
- It will be nice to follow same approach in ptf
- document all existing tags, copy descriptions from sktime/readme files in ptf where possible

~good first issue ;)~ not a good first issue

pyarchana:
#### Reference Issues/PRs

Fixes #2333

adds a class-based tag registry to pytorch-forecasting, following the same
pattern already used in sktime (`sktime/registry/_tags.py`) and skpro
(`skpro/registry/_tags.py`).

- new `pytorch_forecasting/_registry/_tags.py`: `_BaseTag` base class plus one
  subclass per tag currently in use across the codebase (19 tags total, covering
  `object_type`, `capability:*`, `info:*`, `requires:data_type`, and metric-specific
  tags), a discovery loop building `OBJECT_TAG_REGISTER`/`_TABLE`/`_LIST`, and
  `check_tag_is_valid`.
- `all_tags()` added to `_registry/_lookup.py` , this closes a gap where
  `all_objects()`'s own docstring already referenced a `registry.all_tags` utility
  that didn't exist yet.
- exported `all_tags` from `_registry/__init__.py`.
- trimmed the duplicated inline tag descriptions in
  `extension_templates/v2/README.md` down to a short summary plus a pointer to
  the new registry as the single source of truth.

- whether the 19 tags and their `short_descr` text accurately reflect intended
  usage (descriptions were written from reading existing `_pkg.py` files, not
  guessed).
- whether `parent_type` values (plain `object_type` strings like
  `forecaster_pytorch_v1`/`forecaster_pytorch_v2`/`metric`) are the right
  granularity, given pytorch-forecasting doesn't have a formal scitype-class
  registry like sktime/skpro do.

yes, `pytorch_forecasting/tests/test_registry_tags.py` (4 tests, all passing):
registry shape, `check_tag_is_valid` behavior, `all_tags` filtering, and an
AST-based drift check asserting every tag actually used anywhere in the
package is registered in the new module.

#### Any other comments?

 verified locally: new tests pass, plus existing `test_class_register.py` and
`test_base_pkg.py` still pass. Did not run `pre-commit` locally (not installed
in this environment), worth a check in CI.

#### PR checklist

- [x] Added/modified tests
- [ ] Used pre-commit hooks when committing to ensure that code is compliant with hooks.

### DIFF
--- a/pytorch_forecasting/_registry/_tags.py
+++ b/pytorch_forecasting/_registry/_tags.py
@@ -1,0 +1,396 @@
+"""Register of estimator and object tags.
+
+Note for extenders: new tags should be entered in OBJECT_TAG_REGISTER.
+No other place is necessary to add new tags.
+
+This module exports the following:
+
+---
+OBJECT_TAG_REGISTER - list of tuples
+
+each tuple corresponds to a tag, elements as follows:
+    0 : string - name of the tag as used in the _tags dictionary
+    1 : string - name of the object_type this tag applies to,
+                 e.g., "forecaster_pytorch_v1", "forecaster_pytorch_v2", "metric",
+                 or "object" if the tag applies to all objects
+    2 : string - expected type of the tag value
+        should be one of:
+            "bool" - valid values are True/False
+            "int" - valid values are all integers
+            "str" - valid values are all strings
+            "list" - valid values are all lists of arbitrary elements
+            ("str", list_of_string) - any string in list_of_string is valid
+            ("list", list_of_string) - any individual string and sub-list is valid
+            ("list", "str") - any individual string or list of strings is valid
+        validity can be checked by check_tag_is_valid (see below)
+    3 : string - plain English description of the tag
+
+---
+
+OBJECT_TAG_TABLE - pd.DataFrame
+    OBJECT_TAG_REGISTER in table form, as pd.DataFrame
+        rows of OBJECT_TAG_TABLE correspond to elements in OBJECT_TAG_REGISTER
+
+OBJECT_TAG_LIST - list of string
+    elements are 0-th entries of OBJECT_TAG_REGISTER, in same order
+
+---
+
+check_tag_is_valid(tag_name, tag_value) - checks whether tag_value is valid for tag_name
+"""
+
+import inspect
+import sys
+
+import pandas as pd
+
+from pytorch_forecasting.base._base_object import _BaseObject
+
+# ---------------------------------------------------------
+# Tag Class Definitions
+# ---------------------------------------------------------
+
+
+class _BaseTag(_BaseObject):
+    """Base class for all tags in pytorch-forecasting.
+
+    This follows the class-based tag registry pattern used in sktime and skpro,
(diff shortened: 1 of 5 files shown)


pyarchana:
heads up: the 2 failing checks (ubuntu-latest 3.13 and macos-latest 3.10) aren't related to this PR. both hit a 429 (too many requests) while downloading the stallion demo dataset in a session-scoped fixture, which then cascades into 57 errors across everything that depends on it. nothing here touches that code path, so a re-run should clear it whenever you get a chance.

pytorch_forecasting/_registry/_tags.py
```diff
@@ -0,0 +1,396 @@
...
+        "tag_name": "capability:quantile_generation",
+        "parent_type": _FORECASTER_TYPES,
+        "tag_type": "bool",
+        "short_descr": "whether the model supports quantile output generation",
+    }
+
+
+# --------------------------
+# Metrics
+# --------------------------
```
phoeenniixx: Can you please remove these agentic type comments

pyarchana: on it.

pytorch_forecasting/_registry/_tags.py
```diff
@@ -0,0 +1,372 @@
...
+class capability_flexible_history_length(_BaseTag):
+    """Whether the model works with variable-length input history."""
+
+    _tags = {
+        "tag_name": "capability:flexible_history_length",
+        "parent_type": _FORECASTER_TYPES,
+        "tag_type": "bool",
+        "short_descr": "whether the model works with variable-length input history",
+    }
+
```
Faakhir30, line 193: Point of having `_tags.py` was to have enough well documented tags like in sktime/sktime.
Tag should elobarate atleast on:
- possible values
- examples (if needed)
- default value (if any)
- a more descriptive description instead of having a one-liner almost saying tag name but in a few more words
- explain indivual cases, If the tag is ``False``/``True``, what methods behave in what ways?

Comment applies for all tags, reason for pointing this tag specifically is that I have been trying to understand it for past 15 minutes locally, but I cant understand how it plays in v2... ;) so problem is real.

Now, I also think this is not a good first contribution, appologies for mentioning "good first issue" in issue, updated.

It'll be better to have someone with deep understanding of what these tags reflect in ptf across v1/v2 beyond a one-liner. 

cc @phoeenniixx

PR #2335, opened 2026-07-07
[ENH] Implement a unified hyperparameter tuning interface decoupled from model classes for v2

### LINKED ISSUE #2332: [ENH] Implement a unified hyperparameter tuning interface decoupled from model classes for v2
## Is your feature request related to a problem? Please describe.

Currently in v1, hyperparameter optimization is implemented as a model-specific method (e.g., `optimize_hyperparameters` in the Temporal Fusion Transformer). This tightly couples the optimization logic with the model implementation, leading to:
* Duplicated code across different models.
* A lack of a unified tuning interface for the user.

In frameworks like DSIPTS, optimization is decoupled and happens entirely outside the model. This is a much cleaner architectural pattern that we should adopt for our v2 release.

## Describe the solution you'd like

I want to implement a generalized hyperparameter optimization method for v2 that sits completely outside the model classes. 

To handle the fact that hyperparameters and their valid search ranges are inherently model-specific, we propose the following architecture:

1. **Standardized Hook:** Add a standardized hook to the `BaseModel` (e.g., a `get_tuneable_hyperparameters()` method).
2. **Centralized Optimizer:** Create a centralized optimizer that will query this hook to dynamically fetch the list of parameters and their default search ranges for any given model.
3. **User Overrides:** Allow users to easily pass in custom search ranges to override the model's defaults.

Muhammad-Rebaal:
Fixes : #2332 

This PR implements the hyperparameter optimization interface for v2 models. 

### Implemented Architecture

1. **`SearchRange` & Global Registry** (`tuning/search_range.py`, `tuning/global_registry.py`):
   - A typed container for search spaces and a global dictionary defining standard ranges for common parameters (e.g., `hidden_size`, `dropout`, `n_heads`).
   
2. **Auto-Discovery Hook** (`models/base/_base_model_v2.py`):
   - A `get_tuneable_hyperparameters()` class method on `BaseModel` that uses `inspect.signature(cls.__init__)` to detect what parameters the subclass accepts and maps them to the global registry.

3. **`HyperparameterTuner`** (`tuning/hyperparameter_tuner.py`):
   - A standalone optimizer living outside model modules. It queries the base hook, merges user overrides, and runs sequential Optuna trials using `Base_pkg.fit()`.

**User Flow:**

```
from pytorch_forecasting.models.timexer import TimeXer_pkg_v2
from pytorch_forecasting.models.dlinear import DLinear_pkg_v2
from pytorch_forecasting.tuning.hyperparameter_tuner import HyperparameterTuner
from pytorch_forecasting.tuning.search_range import SearchRange

tuner = HyperparameterTuner(
    pkg_cls=TimeXer_pkg_v2,
    data=my_timeseries_data,
    base_datamodule_cfg={"context_length": 96, "prediction_length": 24},
)
study = tuner.optimize(n_trials=50, max_epochs=10)
print(f"Best params: {study.best_params}")
print(f"Best val_loss: {study.best_value}")
```


phoeenniixx (changes requested):
Why aren't we using `optuna` here? Also, in the param grid the values are hard coded

Muhammad-Rebaal:
> Why aren't we using optuna here?

Because of these 2 reason :

**User Experience:** It's much easier for a user to inspect, modify, and pass a simple SearchRange dataclass in custom_ranges than it is to write and pass custom Optuna lambda functions.

**Decoupling:** By keeping Optuna logic out of the base model layer and registry, the models themselves remain completely framework-agnostic. The SearchRange dataclass acts as a clean bridge.

> param grid the values are hard coded

In v1's **optimize_hyperparameters** for TFT, the default search spaces were hardcoded as tuple arguments (e.g., hidden_size_range=(16, 265), dropout_range=(0.1, 0.3)). The global_registry simply takes those v1 hardcoded defaults and standardizes them centrally so they can apply to all models automatically. However, they aren't strictly locked users can completely override these default "hardcoded" values by passing their own custom_ranges to the HyperparameterTuner.

fkiraly (changes requested):
Hm, ok start. I would make all classes and variables private for now, so we can later change. This increases separation of concerns which is good.

Some questions in the design:

* how would this get used?
* This designs custom search range classes. Do we want to do that, instead of relying on one of the standards? (e.g., even `optuna` itself)
* optuna is one choice of what we could do. Have a look at the `hyperactive` tuning class for neural networks, or the `OptCV` class in `sktime`. Could this be useful?

Muhammad-Rebaal:
Thanks for the review.

> Hm, ok start. I would make all classes and variables private for now, so we can later change. This increases separation of concerns which is good.

I've implemented this.

> * how would this get used?

```
from pytorch_forecasting.tuning import ForecastingSearchCV, _SearchRange
from pytorch_forecasting.models.dlinear import DLinear_pkg_v2
from pytorch_forecasting.metrics import SMAPE


datasets = DLinear_pkg_v2.get_test_dataset_from()

search = ForecastingSearchCV(
    pkg_cls=DLinear_pkg_v2,
    param_grid={
        "moving_avg": _SearchRange(param_type="categorical", choices=[3, 5, 7, 11]),
        "dropout": _SearchRange(param_type="float", low=0.1, high=0.5)
    },
    n_trials=3,
    base_model_cfg={"loss": SMAPE()},  
    base_datamodule_cfg={
        "context_length": 8,        
        "prediction_length": 2,
        "batch_size": 2,                  
        "train_val_test_split": (0.5, 0.5)
    },  
    base_trainer_cfg={
        "max_epochs": 2,
        "limit_train_batches": 2,         
        "limit_val_batches": 1            
    },      
)

search.fit(datasets["train"])

predictions = search.predict(datasets["predict"])

print("Best parameters found:", search.best_params_)
```

> * This designs custom search range classes. Do we want to do that, instead of relying on one of the standards? (e.g., even `optuna` itself)

`_SearchRange` is intentionally a thin adapter that maps to **Optuna** internally see suggest() method. If we ever want to support a second backend, e.g., Ray Tune, we only change suggest()  no user-facing code changes. Using raw `optuna.distributions` would hard-couple every model registry entry to **Optuna**.

> * optuna is one choice of what we could do. Have a look at the `hyperactive` tuning class for neural networks, or the `OptCV` class in `sktime`. Could this be useful?

Yes, it is preciously useful, helped me a lot and implemented the wrapper class `ForecastingSearchCV` like it is in `sktime`. Also with the help of the `param_grid`  we don't have to create hooks in the `base_model` class as well.

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,125 @@
...
+        -------
+        optuna.Study
+            The completed study with results.
+        """
+
+        try:
+            import optuna
+        except ImportError:
+            raise ImportError(
+                "Optuna is required for hyperparameter tuning. "
```
phoeenniixx: why aren't we using `_safe_import` or `_check_soft_dependencies` here?

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,125 @@
...
+        n_trials : int
+            Number of Optuna trials.
+        timeout : float
+            Maximum time in seconds.
+        max_epochs : int
+            Max training epochs per trial.
+        custom_ranges : dict[str, SearchRange], optional
+            Override or add to auto-discovered ranges.
+        study : optuna.Study, optional
+            Existing study to resume.
```
phoeenniixx: what if there is no existing study and the user wants to create a new study here?

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,125 @@
...
+                "Optuna is required for hyperparameter tuning. "
+                "Please install it with `pip install optuna`"
+            )
+
+        search_ranges = {}
+
+        if custom_ranges:
+            search_ranges.update(custom_ranges)
+
+        def objective(trial):
```
phoeenniixx: should this be a private method?

phoeenniixx (changes requested):
- Please use soft-dep isolation when importing soft dependencies 
See https://www.sktime.net/docs/developer-guide/dependencies/#isolating-soft-dependencies-to-estimators
- I think `lightning.tuner` can also be very helpful here, no?
- Please add tests

phoeenniixx:
> from pytorch_forecasting.tuning import ForecastingSearchCV, _SearchRange
> from pytorch_forecasting.models.dlinear import DLinear_pkg_v2
> from pytorch_forecasting.metrics import SMAPE
> 
> 
> datasets = DLinear_pkg_v2.get_test_dataset_from()
> 
> search = ForecastingSearchCV(
>     pkg_cls=DLinear_pkg_v2,
>     param_grid={
>         "moving_avg": _SearchRange(param_type="categorical", choices=[3, 5, 7, 11]),
>         "dropout": _SearchRange(param_type="float", low=0.1, high=0.5)
>     },
>     n_trials=3,
>     base_model_cfg={"loss": SMAPE()},  
>     base_datamodule_cfg={
>         "context_length": 8,        
>         "prediction_length": 2,
>         "batch_size": 2,                  
>         "train_val_test_split": (0.5, 0.5)
>     },  
>     base_trainer_cfg={
>         "max_epochs": 2,
>         "limit_train_batches": 2,         
>         "limit_val_batches": 1            
>     },      
> )
> 
> search.fit(datasets["train"])
> 
> predictions = search.predict(datasets["predict"])
> 
> print("Best parameters found:", search.best_params_)

based on this vignette:
Why the user has to pass a private `_SearchRange` class? Why dont we create this inside the wrapper itself and the user passes the params more organically?

Also, why do we need to pass the cfgs again to this class, I thought we could simply pass the `pkg` class and it would handle the params itself?

pytorch_forecasting/tuning/hyperparameter_tuner.py
phoeenniixx: I am talking about the method not the file?
And this is not even the file that I have added this comment on?

Muhammad-Rebaal:
> > from pytorch_forecasting.tuning import ForecastingSearchCV, _SearchRange
> > from pytorch_forecasting.models.dlinear import DLinear_pkg_v2
> > from pytorch_forecasting.metrics import SMAPE
> > datasets = DLinear_pkg_v2.get_test_dataset_from()
> > search = ForecastingSearchCV(
> > pkg_cls=DLinear_pkg_v2,
> > param_grid={
> > "moving_avg": _SearchRange(param_type="categorical", choices=[3, 5, 7, 11]),
> > "dropout": _SearchRange(param_type="float", low=0.1, high=0.5)
> > },
> > n_trials=3,
> > base_model_cfg={"loss": SMAPE()},
> > base_datamodule_cfg={
> > "context_length": 8,
> > "prediction_length": 2,
> > "batch_size": 2,
> > "train_val_test_split": (0.5, 0.5)
> > },
> > base_trainer_cfg={
> > "max_epochs": 2,
> > "limit_train_batches": 2,
> > "limit_val_batches": 1
> > },
> > )
> > search.fit(datasets["train"])
> > predictions = search.predict(datasets["predict"])
> > print("Best parameters found:", search.best_params_)
> 
> based on this vignette: Why the user has to pass a private `_SearchRange` class? Why dont we create this inside the wrapper itself and the user passes the params more organically?
> 
> Also, why do we need to pass the cfgs again to this class, I thought we could simply pass the `pkg` class and it would handle the params itself?

Nice Idea I'll try to implement that.

pytorch_forecasting/tuning/hyperparameter_tuner.py
Muhammad-Rebaal: I just got a little confused, thanks. I'll make `objective` method private as its an internal one.

Muhammad-Rebaal:
> * I think `lightning.tuner` can also be very helpful here, no?

Yes, 
What I found is :
* `Tuner.lr_find()` —> finds optimal learning rate
* `Tuner.scale_batch_size()` —> finds max batch size that fits in memory
Ref : https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.tuner.tuning.Tuner.html

These are the methods I found that can be helpful. Any suggestions from your side.. @phoeenniixx

pytorch_forecasting/tuning/hyperparameter_tuner.py
phoeenniixx: Please dont use AI blindly, this creates a lot of confusion and wastes time for all of us

phoeenniixx (changes requested):
I think the design still has a lot of flaws, we should discuss this in one of the meetings to make it clear

pytorch_forecasting/models/base/_base_object.py
```diff
@@ -117,4 +118,34 @@ class _BasePtForecasterV2(_BasePtForecaster_Common):
 
     _tags = {
         "object_type": "forecaster_pytorch_v2",
+        "tunable_params": {
+            "optimizer": _SearchRange(
```
phoeenniixx: why do we need to use this private class in tags? Please see my prev reviews, this should be used inside the wrapper, here we should just pass the param names (and the choices).

Also, I think `optimizer`, `optimizer_params` are common params which will be present in all teh model params no? This `tunable_params` should be a part of each model `pkg` class, not the base class. Also, here you are using the Common base object class used in both v1 and v2, this is not the right place for this. It should be a part of `Base_pkg` instead.

pytorch_forecasting/base/_base_pkg.py
```diff
@@ -256,8 +256,15 @@ def fit(
         trainer_init_cfg = self.trainer_cfg.copy()
         trainer_init_cfg.pop("callbacks", None)
 
+        auto_lr_find = trainer_init_cfg.pop("auto_lr_find", False)
```
phoeenniixx: should this be a `cfg` for `trainer`?

pytorch_forecasting/models/base/_base_object.py
```diff
@@ -117,4 +118,34 @@ class _BasePtForecasterV2(_BasePtForecaster_Common):
...
+                param_type="categorical", choices=["adam", "adamw"]
+            ),
+            "optimizer_params.lr": _SearchRange(
+                param_type="float", low=1e-5, high=1e-1, log=True
```
phoeenniixx: i think for this we decided to use `inspect.signature`?

pytorch_forecasting/tests/test_tuning.py
phoeenniixx, line 1: I think these tests are too verbose - do we really need them? All of them?

Also please add complete docstrings for the tests

pytorch_forecasting/tuning/forecasting_search_cv.py
```diff
@@ -0,0 +1,211 @@
+"""ForecastingSearchCV: Hyperparameter search with fit/predict interface."""
+
+import copy
+import inspect
+
+from pytorch_forecasting.tuning.hyperparameter_tuner import _HyperparameterTuner
+
+
+class ForecastingSearchCV:
```
phoeenniixx: How is this realted to `tuner` from `lightning`? Or is this a separate functionality?

phoeenniixx (changes requested):
I think the design is still not very clear to me

- How does the `tuner` gets the params to be tuned?
- What is the role of `ForecastingSearchCV` here?
   - I dont remember any vignette even using it?
- I remeber we thought of using `inspect` to see the signature of the models - i can still see the long tags which are not needed now?

Other than that, few comments:
- Please add the clear docstrings to the tests
- I think rn they are too much? And feels fully vibe-coded as it even checks if the params are correctly passed - which you literally pass few lines above? 


Can you please add some vignettes of how this will be used - i am still not very clear on this

Muhammad-Rebaal:
> I think the design is still not very clear to me
> 
> * How does the `tuner` gets the params to be tuned?
> * What is the role of `ForecastingSearchCV` here?
>   
>   * I dont remember any vignette even using it?
> * I remeber we thought of using `inspect` to see the signature of the models - i can still see the long tags which are not needed now?
> 
> Other than that, few comments:
> 
> * Please add the clear docstrings to the tests
> * I think rn they are too much? And feels fully vibe-coded as it even checks if the params are correctly passed - which you literally pass few lines above?
> 
> Can you please add some vignettes of how this will be used - i am still not very clear on this

Yes, I'm currently finishing up the design doc right now which we discussed about which I'd send you tonight or tomorrow morning that's why it is pending right now and yes this PR needs refactoring which I'd do once the design is approved and everything is clear.

Muhammad-Rebaal:
Hi @phoeenniixx,

Now the Hyperparameters can be tuned for any v2 model like this :

```
tuner = HyperparameterTuner(
    model="TFT",
    data=my_datamodule, 
)
```
Although the user can provide their `trainer_cfg`, and `datamodule_cfg` if passed the timeseries data.

Muhammad-Rebaal:
Hi @phoeenniixx,
I've made the following changes:
- Removed `datamodule_cfg`, `trainer_cfg`.

The user can now pass raw timeseries and datamodule as well if passed raw timeseries datamodule would be made to processed and if lightening datamodule is passed then the user can also specify the train-test-split  that we are working on another PR. 

Lastly, every model has a get_datamodule_cls method implemented in their respective pkg coming from base_pkg, and since we are moving towards deprecation of the base_pkg. I've moved the get_datamodule_cls in their respective model_cls such as 
- BaseModel has `EncoderDecoderTimeSeriesDataModule`
- TslibBaseModel override the BaseModel get_datamodule_cls method `TslibDataModule`

The purpose of this is that the user don't have to specify which datamodule is for which model and converts their data into the specific datamodule to be processed also added the edge case if the user passes the wrong datamodule it would raise a `TypeError`

Muhammad-Rebaal:
Hi @phoeenniixx,
I've made the following change:
- Added a variable `datamodule_cls`, now the user have to define the datamodule he has to use for the model

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,268 @@
...
+            The model class directly (e.g., DLinear, TFT).
+        data : TimeSeries or LightningDataModule
+            Dataset to split and reuse across all trials. If passing a
+            TimeSeries, ``datamodule_cls`` must be provided.
+        datamodule_cls : type[LightningDataModule], optional
+            The DataModule class to construct when ``data`` is a TimeSeries,
+            or against which to validate when ``data`` is a prebuilt DataModule.
+        **fixed_hparams
+            Any model parameter that should stay constant, e.g.
+            ``hidden_size=128``.
```
phoeenniixx, line 33: please add some examples here to how to use this

phoeenniixx (changes requested):
Thanks
Can you please clean up the test file? it has a lot of "param checks" like if i pass a, it is collected in `x.param` etc. It checks if `something` is implemented or not - we should only test if the things are working correctly... Not that if the tests accpet some param or not, it is understood that the class will accept `datamodule` and assign it to `tuner.datamoudle`. 
(There are multiple tests like that here)

Also, pls add some examples (maybe even an example nb!)

Muhammad-Rebaal:
Hi @phoeenniixx,
I've cleaned up the test file and also added an example.

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,306 @@
...
+        """
+        self.model_cls = model_cls
+        self.fixed_hparams = fixed_hparams
+        self._validate_fixed_hparams()
+
+        # 1. Validate datamodule_cls if provided
+        if datamodule_cls is not None:
+            if not (
+                isinstance(datamodule_cls, type)
+                and issubclass(datamodule_cls, LightningDataModule)
```
phoeenniixx, line 81: do we need this validation? if we add type hints etc, it would be understood that it is a `LightningDataModule` no?

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,306 @@
...
+                max_epochs=max_epochs,
+                accelerator="auto",
+                enable_progress_bar=False,
+                enable_model_summary=False,
+            )
+
+            trainer.fit(model, datamodule=self.datamodule)
+
+            metrics = trainer.callback_metrics
+            for key in ("val_loss", "train_loss_epoch", "train_loss"):
```
phoeenniixx, line 293: should we optimize for train losses? that might lead to overfitting?

pytorch_forecasting/tuning/hyperparameter_tuner.py
```diff
@@ -0,0 +1,306 @@
...
+        from pytorch_forecasting.models.base._base_model_v2 import BaseModel
+
+        base_params = set(inspect.signature(BaseModel.__init__).parameters.keys())
+        sig = inspect.signature(self.model_cls.__init__)
+
+        sampled_cfg = {}
+        custom_ranges = custom_ranges or {}
+
+        for param_name, param_obj in sig.parameters.items():
+            if param_name in base_params or param_name == "self":
```
phoeenniixx, line 162: Here what will happen to the params which are sent to the`BaseModel`? Like `Samformer` sends optimizer etc directly to base model:
https://github.com/sktime/pytorch-forecasting/blob/0158d1779cedb68a4b5778ccc73d36ac41572f7f/pytorch_forecasting/models/samformer/_samformer_v2.py#L61-L69

phoeenniixx (changes requested):
Thanks!
Will this work for a model checkpoint etc as well?
Also, should we also add a nb etc  - as rn we cant verify that the complete workflow works as expected

Issue #2338, opened 2026-07-10
[ENH]  Enable model caching irrespective of data being passed at refit in v2


Faakhir30:
**Describe the bug**

I think we are caching model irrespective of data being passed at `fit`:
```python
# pytorch-forecasting/pytorch_forecasting/base/_base_pkg.py
        if self.model is None: # caching irrespective of `data`
            if not self.model_cfg:
                raise RuntimeError(
                    "`model_cfg` must be provided to train from scratch."
                )
            metadata = self.datamodule.metadata
            self._build_model(metadata)
```

I think new datamodules should always create new model, otherwise pipeline will be using model with stale `metadata` and other attributes.

**To Reproduce**

```python
import numpy as np
import pandas as pd

from pytorch_forecasting.data import TimeSeries
from pytorch_forecasting.metrics import SMAPE
from pytorch_forecasting.models.temporal_fusion_transformer._tft_pkg_v2 import (
    TFT_pkg_v2,
)


def make_ts(n_cont):
    rng = np.random.default_rng(0)
    rows = []
    for g in range(2):
        for t in range(20):
            row = {
                "time_idx": t,
                "group": g,
                "target": float(rng.normal()),
                "weight": 1.0,
            }
            for i in range(n_cont):
                row[f"x{i}"] = float(rng.normal())
            rows.append(row)

    return TimeSeries(
        data=pd.DataFrame(rows),
        time="time_idx",
        target=["target"],
        group=["group"],
        weight="weight",
        num=[f"x{i}" for i in range(n_cont)],
        known=[f"x{i}" for i in range(n_cont)],
        unknown=[],
        static=["group"],
    )


pkg = TFT_pkg_v2(
    model_cfg={
        "loss": SMAPE(),
        "hidden_size": 16,
        "attention_head_size": 2,
    },
    datamodule_cfg={
        "max_encoder_length": 8,
        "max_prediction_length": 3,
        "train_val_test_split": (0.8, 0.2),
        "batch_size": 2,
    },
    trainer_cfg={
        "max_epochs": 1,
        "limit_train_batches": 1,
        "limit_val_batches": 1,
        "logger": False,
        "enable_checkpointing": False,
        "enable_progress_bar": False,
        "accelerator": "cpu",
    },
)

pkg.fit(make_ts(2), save_ckpt=False)  # initializes model from metadata A
pkg.fit(make_ts(5), save_ckpt=False)  # RuntimeError: stale model reused

```
**Output**

[...]

PR #2340, opened 2026-07-10
[ENH] Add xLSTMTime model in v2

### LINKED ISSUE #2039: [ENH] Add v2 interface support for `xLSTMTime`
Referencing #1992 

The xLSTMTime model currently only supports the v1 interface. With the introduction of the v2 architecture using `TslibBaseModel` and `TslibDataModule`, xLSTMTime should be migrated to support the new interface for consistency with other v2 models and to leverage improvements in the v2 design.

Implement `xLSTMTime_v2` with the following components:

1. **Model class** (`_xlstm_v2.py`): Extend `TslibBaseModel` with proper handling of:
   - Dynamic `input_size` and `output_size` derived from metadata
   - Both sLSTM and mLSTM variants
   - Proper dimension handling for BatchNorm and LSTM layers

2. **Package class** (`_xlstm_v2_pkg_v2.py`): Extend `Base_pkg` with:
   - Auto-computation of `input_size` (from continuous features) and `output_size` (prediction_length × n_targets)

3. **Integration**: Ensure compatibility with:
   - `TslibDataModule` for data handling
   - Standard v2 test suite (integration, checkpointing, prediction modes)

N/A

**Additional context**
N/A

CC : @fkiraly

Faakhir30:
#### Reference Issues/PRs

Fixes #2039

- Implements  `xLSTMTime` for v2
- adds package wrapper `xLSTMTime_pkg_v2` 

Nope

#### PR checklist

- [X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [X] Added/modified tests
- [X] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.

### DIFF
--- a/pytorch_forecasting/models/xlstm/__init__.py
+++ b/pytorch_forecasting/models/xlstm/__init__.py
@@ -2,5 +2,7 @@
 
 from pytorch_forecasting.models.xlstm._xlstm import xLSTMTime
 from pytorch_forecasting.models.xlstm._xlstm_pkg import xLSTMTime_pkg
+from pytorch_forecasting.models.xlstm._xlstm_pkg_v2 import xLSTMTime_pkg_v2
+from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime as xLSTMTimeV2
 
-__all__ = ["xLSTMTime", "xLSTMTime_pkg"]
+__all__ = ["xLSTMTime", "xLSTMTimeV2", "xLSTMTime_pkg", "xLSTMTime_pkg_v2"]
--- a/pytorch_forecasting/models/xlstm/__init__.py
+++ b/pytorch_forecasting/models/xlstm/__init__.py
@@ -3,6 +3,6 @@
 from pytorch_forecasting.models.xlstm._xlstm import xLSTMTime
 from pytorch_forecasting.models.xlstm._xlstm_pkg import xLSTMTime_pkg
 from pytorch_forecasting.models.xlstm._xlstm_pkg_v2 import xLSTMTime_pkg_v2
-from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime as xLSTMTimeV2
+from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime_v2
 
-__all__ = ["xLSTMTime", "xLSTMTimeV2", "xLSTMTime_pkg", "xLSTMTime_pkg_v2"]
+__all__ = ["xLSTMTime", "xLSTMTime_v2", "xLSTMTime_pkg", "xLSTMTime_pkg_v2"]
--- a/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
+++ b/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
@@ -1,0 +1,73 @@
+"""xLSTMTime v2 package container."""
+
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class xLSTMTime_pkg_v2(Base_pkg):
+    """xLSTMTime v2 package container."""
+
+    _tags = {
+        "info:name": "xLSTMTime",
+        "info:compute": 3,
+        "info:y_type": ["numeric"],
+        "authors": ["muslehal", "phoeenniixx"],
+        "capability:exogenous": True,
+        "capability:multivariate": False,
+        "capability:pred_int": True,
+        "capability:flexible_history_length": False,
+        "capability:cold_start": False,
+    }
+
+    @classmethod
+    def get_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime
+
+        return xLSTMTime
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data.data_module import (
+            EncoderDecoderTimeSeriesDataModule,
+        )
+
+        return EncoderDecoderTimeSeriesDataModule
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Return testing parameter settings for the trainer."""
+        from pytorch_forecasting.metrics import QuantileLoss
+
+        params = [
+            {},
+            {"xlstm_type": "mlstm"},
+            {"num_layers": 2},
+            {"xlstm_type": "slstm", "input_projection_size": 32},
+            {
+                "xlstm_type": "mlstm",
+                "decomposition_kernel": 3,
+                "dropout": 0.2,
+            },
+            {
+                "loss": QuantileLoss(quantiles=[0.1, 0.5, 0.9]),
+                "hidden_size": 16,
+            },
+            {
+                "optimizer": "adamw",
+                "lr_scheduler": "cosine_annealing",
+                "lr_scheduler_params": {"T_max": 5},
+            },
+        ]
+
+        default_dm_cfg = {
+            "max_encoder_length": 8,
+            "max_prediction_length": 3,
+        }
+
+        for param in params:
+            current_dm_cfg = param.get("datamodule_cfg", {})
+            default_dm_cfg.update(current_dm_cfg)
+            param["datamodule_cfg"] = default_dm_cfg.copy()
+
+        return params
--- a/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
+++ b/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
@@ -10,7 +10,7 @@ class xLSTMTime_pkg_v2(Base_pkg):
         "info:name": "xLSTMTime",
         "info:compute": 3,
         "info:y_type": ["numeric"],
-        "authors": ["Faakhir30"],
+        "authors": ["muslehal", "phoeenniixx", "Faakhir30"],
         "capability:exogenous": True,
         "capability:multivariate": False,
         "capability:pred_int": False,
@@ -21,9 +21,9 @@ class xLSTMTime_pkg_v2(Base_pkg):
     @classmethod
     def get_cls(cls):
         """Get model class."""
-        from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime
+        from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime_v2
 
-        return xLSTMTime
+        return xLSTMTime_v2
 
     @classmethod
     def get_datamodule_cls(cls):
@@ -37,6 +37,7 @@ class xLSTMTime_pkg_v2(Base_pkg):
     @classmethod
     def get_test_train_params(cls):
         """Return testing parameter settings for the trainer."""
+        from pytorch_forecasting.metrics import MAE, MAPE
 
         params = [
             {},
@@ -47,6 +48,10 @@ class xLSTMTime_pkg_v2(Base_pkg):
                 "xlstm_type": "mlstm",
                 "decomposition_kernel": 3,
                 "dropout": 0.2,
+                "loss": MAE(),
+            },
+            {
+                "loss": MAPE(),
             },
             {
                 "optimizer": "adamw",
--- a/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
+++ b/pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
@@ -10,7 +10,7 @@ class xLSTMTime_pkg_v2(Base_pkg):
         "info:name": "xLSTMTime",
         "info:compute": 3,
         "info:y_type": ["numeric"],
-        "authors": ["muslehal", "phoeenniixx"],
+        "authors": ["Faakhir30"],
         "capability:exogenous": True,
         "capability:multivariate": False,
         "capability:pred_int": True,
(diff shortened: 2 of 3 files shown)


pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
```diff
@@ -0,0 +1,73 @@
...
+
+    _tags = {
+        "info:name": "xLSTMTime",
+        "info:compute": 3,
+        "info:y_type": ["numeric"],
+        "authors": ["Faakhir30"],
+        "capability:exogenous": True,
+        "capability:multivariate": False,
+        "capability:pred_int": True,
+        "capability:flexible_history_length": False,
```
Faakhir30: two changes from v1:
- `pred_int` should be `True`, right? because we can give quantile support in v2
- `capability:flexible_history_length` means history_length can be variable across `fit` calls? if thats the case then it should be False, refer #2338

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
```diff
@@ -0,0 +1,73 @@
...
+            {"xlstm_type": "mlstm"},
+            {"num_layers": 2},
+            {"xlstm_type": "slstm", "input_projection_size": 32},
+            {
+                "xlstm_type": "mlstm",
+                "decomposition_kernel": 3,
+                "dropout": 0.2,
+            },
+            {
+                "loss": QuantileLoss(quantiles=[0.1, 0.5, 0.9]),
```
phoeenniixx: I remember v1 implementation didnt support quantile loss? I think you have added some improvements!?

phoeenniixx (changes requested):
before a deep review - i have a question out of curiosity - is this implementation similar to what was present in v1? Or have you implemented this from scratch?

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
Faakhir30: yes, as mentioned in above comment regarding changes from v1:

>two changes from v1:
>pred_int should be True, right? because we can give quantile support in v2
>capability:flexible_history_length means history_length can be variable across fit calls? if thats the case then it should be False, refer https://github.com/sktime/pytorch-forecasting/issues/2338

Faakhir30:
> is this implementation similar to what was present in v1

yeah, apart from 2 things mentioned above, and ofcourse other minor things like:
- reading data from metadata
- conforming output format to v2 API in `forward`

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
phoeenniixx: what i mean is, although v2 pipeline support he quantile loss, does the model implementation supports it?

phoeenniixx:
> > is this implementation similar to what was present in v1
> 
> yeah, apart from 2 things mentioned above, and ofcourse other minor things like:
> 
> * reading data from metadata
> * conforming output format to v2 API in `forward`

Nice, happy to hear this!

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
Faakhir30: hmm,
My current quantile support implementation has been adapted from `DLiner` and `Samformer`.
Those v2 models seems to define output dim as:
`output_dim = self.prediction_length * self.n_quantiles` 

https://github.com/sktime/pytorch-forecasting/blob/e7c7965cc08a52a566eb24661f31c06b49c88395/pytorch_forecasting/models/dlinear/_dlinear_v2.py#L244-L252

https://github.com/sktime/pytorch-forecasting/blob/e7c7965cc08a52a566eb24661f31c06b49c88395/pytorch_forecasting/models/samformer/_samformer_v2.py#L179-L186

So, idea currently implemented in all of these is to extend head dimention, as `QuantileLoss` expects predictions shaped (batch, horizon, n_quantiles) and then it applies pinball loss per quantile index / prediction_length.

Final statement:  the head supports it. the original core networks did not uniquely “know” about quantiles. Would you consider this as quantile support? 

head is part of the "model implementation", but personally, I'm unsure whether backpropagation in above senario will make models learn anything about quantiles. And even if it is intended/required.

phoeenniixx: > the original core networks did not uniquely “know” about quantiles. Would you consider this as quantile support?

Idts that is quantile support, I think it is more about if the network knows the quantiles? I think we should look at the original implementation (the paper and the repo), does that support it? I think it may be a possibility that I might have made this network simpler while implementing it in v1 as at that time I was still understanding the interface and I might have reduced the complexity... I dont remember exactly :)

Faakhir30: > I think we should look at the original implementation (the paper and the repo), does that support it?

- [xsltmtime paper](https://arxiv.org/pdf/2407.10240) dont mention `quantile` anywhere. I'll remove it from xLSTMTime.

- Should we also remove it from [dlinear](https://arxiv.org/pdf/2205.13504) and [samformer](https://arxiv.org/pdf/2402.10198)?? those papers also dont mention `quantile` anywhere.

phoeenniixx: > * Should we also remove it from [dlinear](https://arxiv.org/pdf/2205.13504) and [samformer](https://arxiv.org/pdf/2402.10198)?? those papers also dont mention `quantile` anywhere.

I agree, we should check if the models actually are aware of the quantiles or not

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class xLSTMTime_pkg_v2(Base_pkg):
+    """xLSTMTime v2 package container."""
+
+    _tags = {
+        "info:name": "xLSTMTime",
+        "info:compute": 3,
+        "info:y_type": ["numeric"],
+        "authors": ["Faakhir30"],
```
phoeenniixx: we should add the authors from original implementation as well (and if the code is quite same as v1, then v1 authors as well)

pytorch_forecasting/models/xlstm/__init__.py
```diff
@@ -2,5 +2,7 @@
 
 from pytorch_forecasting.models.xlstm._xlstm import xLSTMTime
 from pytorch_forecasting.models.xlstm._xlstm_pkg import xLSTMTime_pkg
+from pytorch_forecasting.models.xlstm._xlstm_pkg_v2 import xLSTMTime_pkg_v2
+from pytorch_forecasting.models.xlstm._xlstm_v2 import xLSTMTime as xLSTMTimeV2
```
phoeenniixx: I think if we need to do this:
```python
from pytorch_forecasting.models.xlstm import xLSTMTimeV2
```
Then why not just rename the class itself? We can change these once we have a fully stable v2

pytorch_forecasting/models/xlstm/_xlstm_pkg_v2.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data.data_module import (
+            EncoderDecoderTimeSeriesDataModule,
+        )
+
+        return EncoderDecoderTimeSeriesDataModule
+
+    @classmethod
+    def get_test_train_params(cls):
```
phoeenniixx, line 38: it would be good to add some losses in the params as well - they will be just point preditction ones now right?

pytorch_forecasting/models/xlstm/_xlstm_v2.py
```diff
@@ -0,0 +1,199 @@
...
+                "Error in xLSTMTime: xlstm_type must be either 'slstm' or 'mlstm'"
+            )
+
+        self.max_encoder_length = self.metadata["max_encoder_length"]
+        self.max_prediction_length = self.metadata["max_prediction_length"]
+        self.input_size = self.metadata["encoder_cont"] + 1
+
+        self.n_quantiles = 1
+        if hasattr(loss, "quantiles") and loss.quantiles is not None:
+            self.n_quantiles = len(loss.quantiles)
```
phoeenniixx, line 114: O thought we removed quantiles?

pytorch_forecasting/models/xlstm/_xlstm_v2.py
```diff
@@ -0,0 +1,199 @@
...
+        # output_size ~= forecast horizon; extend by n_quantiles when needed
+        self.output_size = self.max_prediction_length * self.n_quantiles
+
+        self.xlstm_type = xlstm_type
+        self.input_projection_size = input_projection_size or hidden_size
+
+        # clamp kernel so short test contexts still work (v1 assumed long history)
+        # kernel = min(decomposition_kernel, self.max_encoder_length)
+        # if kernel % 2 == 0:
+        #     kernel = max(1, kernel - 1)
```
phoeenniixx: can you please explain this block?

pytorch_forecasting/models/xlstm/_xlstm_v2.py
```diff
@@ -0,0 +1,199 @@
...
+        metadata: dict | None = None,
+        **kwargs,
+    ):
+        super().__init__(
+            loss=loss,
+            logging_metrics=logging_metrics,
+            optimizer=optimizer,
+            optimizer_params=optimizer_params,
+            lr_scheduler=lr_scheduler,
+            lr_scheduler_params=lr_scheduler_params,
```
phoeenniixx, line 92: a suggestion:
We should first collect all the params passed to `__init__` in `self` and then we should use them and not directly.

pytorch_forecasting/models/xlstm/_xlstm_v2.py
```diff
@@ -0,0 +1,199 @@
...
+    stabilized LSTM (sLSTM) or the matrix-memory mLSTM variant.
+
+    Parameters
+    ----------
+    loss : nn.Module
+        Loss (and evaluation metric) used during training.
+    hidden_size : int, default 32
+        Hidden size of the xLSTM network; also used by batch norm / LSTM internals.
+    xlstm_type : {"slstm", "mlstm"}, default "slstm"
+        Specifies which xLSTM variant to use:
```
phoeenniixx, line 30: we should add a blank line here so that the list is rendered correctly in the docs

pytorch_forecasting/models/xlstm/_xlstm_v2.py
```diff
@@ -0,0 +1,199 @@
...
+    lr_scheduler : str, optional
+        Learning rate scheduler name.
+    lr_scheduler_params : dict, optional
+        Parameters for the learning rate scheduler.
+    metadata : dict, optional
+        Metadata from the encoder-decoder datamodule. Used to derive
+        ``input_size`` (``encoder_cont + 1`` for ``target_past``) and
+        ``output_size`` (``max_prediction_length``).
+
+    Based on https://arxiv.org/pdf/2407.10240 and https://github.com/muslehal/xLSTMTime
```
phoeenniixx: this should be BEFORE `Parameters`

phoeenniixx (changes requested):
Nice!
Just added few comments above

pytorch_forecasting/models/xlstm/_xlstm_v2.py
Faakhir30: oh yeah, this is dead code, removed the `n_quantiles` in forward, updated, thanks

pytorch_forecasting/models/xlstm/__init__.py
Faakhir30: renamed to `xLSTMTime_v2` for now.
Can also be `xLSTMTimeModel` instead, or any other suggestion?

Renaming it to `xLSTMTimeV2` will require to make pkg name `xLSTMTimeV2_pkg_v2` (slightly awkward to have two v2 in a name) due to test `test_pkg_linkage`

pytorch_forecasting/models/xlstm/_xlstm_v2.py
Faakhir30: forgot to remove commented code, was originally facing error on some test, removed.

pytorch_forecasting/models/xlstm/__init__.py
phoeenniixx: > Renaming it to `xLSTMTimeV2` will require to make pkg name `xLSTMTimeV2_pkg_v2` (slightly awkward to have two v2 in a name) due to test `test_pkg_linkage`

I thought this issue was solved by @PalakB09 ? Can you pls recheck and lmk. See #2080

phoeenniixx: > renamed to `xLSTMTime_v2` for now.
> Can also be `xLSTMTimeModel` instead, or any other suggestion?

We can do this ofcourse, but think of the user who is using `ptf` and they wants to `xLSTMTime`, but as v2 is unstable, they cant use that in the production. So, to reduce the confusion, I think adding v2 at the end will be better so that the user knows which version of the model they are using!

Faakhir30: > I thought this issue was solved by @PalakB09 ? Can you pls recheck and lmk. See https://github.com/sktime/pytorch-forecasting/pull/2080

It shows relaxation for `_v2` that I have already used by renaming model to `xLSTMTime_v2`.
It doesnt handles `V2` as a special case. 

`_v2` also seems fine, does bring the clarity of v2. 
If `xLSTMTimeV2` is strictly necessary, we need to allow `V2` also as a special case, lmk

phoeenniixx (approved):
Thanks!

PR #2211, opened 2026-03-18
[ENH] Add optimized D1 layer categorical encoder for v2

Siddhazntx:
#### Reference Issues/PRs

Addresses the `label_encoders` task mentioned in the v2 roadmap tracking issue : #1974 

#### What does this implement/fix? Explain your changes.
This PR introduces an optimized `D1CategoricalEncoder` to the v2 data pipeline to handle categorical and text variables, preventing PyTorch tensor conversion crashes.

**Key Changes:**
* **New Encoder Class:** Created `_encoders_v2.py` featuring a `D1CategoricalEncoder` that strictly follows the `scikit-learn` API (`fit`, `transform`, `inverse_transform`).
* **C-Level Optimization:** Utilized `pd.factorize()` instead of native Python dictionaries to ensure the encoding process is efficient and scalable for large datasets.
* **Robust Edge-Case Handling:** Safely manages original `NaN` values without silently dropping them.
  * Handles unseen variables during the `transform` phase (defaulting to `0`).
  * Implemented a `_warned_cols` set to ensure warnings for unseen variables only trigger once per column, preventing terminal flooding during dataloader loops.
* **D1 Layer Integration:** Integrated the encoder into the `__init__` of `TimeSeries` inside `_timeseries_v2.py`. Columns specified in the `cat` argument are now automatically encoded.

#### What should a reviewer concentrate their feedback on?
* **Integration Point:** Please review the placement of the encoding logic within `_timeseries_v2.py`'s `__init__` method to ensure it aligns with the intended v2 data ingestion flow.
* **Unseen Variable Strategy:** I defaulted to `handle_unknown="assign_new"` (mapping to 0). Let me know if the core team prefers a different default behavior for the v2 release!

#### Did you add any tests for the change?
Yes. I added a comprehensive `pytest` suite in a new `test_encoders_v2.py` file. Tests include:
- `test_encoder_fit_transform`: Validates integer conversion and preservation of numeric columns using `pd.api.types.is_integer_dtype`.
- `test_encoder_inverse_transform`: Ensures perfect reverse translation, including restoring true `NaN` values.
- `test_unseen_variables_warning`: Confirms correct fallback assignment and verifies the custom warning triggers.
- `test_only_categorical_columns_selected`: Ensures the auto-detect feature properly ignores numeric columns when `columns=None`.

#### PR checklist

[...]

PR #2346, opened 2026-07-17
[ENH] Add optimized D1 layer categorical encoder for v2 (Follow up on #2211 Open)

### LINKED ISSUE #2347: [ENH] Add optimized D1 layer categorical encoder for v2
### Is your feature request related to a problem? Please describe.
Currently, the v2 data pipeline lacks robust native support for categorical and text variables. When passing datasets with string or categorical columns, it results in PyTorch tensor conversion crashes. There is a need for an efficient, standardized way to encode these variables before they hit the neural network layers.

### Describe the solution you'd like
We need to implement an optimized `D1CategoricalEncoder` specifically designed for the v2 data pipeline. 

Key requirements for the implementation:
* **API Compatibility:** Must strictly follow the scikit-learn API (`fit`, `transform`, `inverse_transform`).
* **Performance:** Should use C-level optimization (like `pd.factorize()`) rather than native Python dictionaries to ensure it scales well with large datasets.
* **Edge-Case Handling:** 
  * Safely manage original `NaN` values without silently dropping them.
  * Handle unseen variables during the `transform` phase (e.g., mapping to index `0`).
  * Implement a warning system for unseen variables that only triggers once per column, preventing terminal flooding during dataloader loops.
* **Integration:** Seamlessly integrate the encoder into the `__init__` of `TimeSeries` inside `_timeseries_v2.py`. Columns specified in a `cat` argument should automatically be encoded.

### Additional context
This addresses the `label_encoders` task mentioned in the v2 roadmap tracking issue: #1974.

Muhammad-Rebaal:
Fixes #2347 
Follow-up to #2211 (picking up stale work by @Siddhazntx).

This PR implements the `D1CategoricalEncoder` for `_timeseries_v2.py`, keeping the core logic of the original PR while fixing a critical data leakage bug and resolving CI warnings.

### Key Changes
* **Fixed Data Leakage:** Updated `TimeSeries` initialization to accept a pre-fitted `categorical_encoder`. Previously, `.fit()` was called unconditionally, which rebuilt vocabularies for validation/test sets and broke PyTorch Embedding indices.
* **Robust `<UNK>` Handling:** The encoder maps unknown categories to index `0`, with actual categories starting at `1`.
* **Pandas 3.0 Compatibility:** Resolved CI deprecation warnings in `_encoders_v2.py` by explicitly including `"string"`/`"str"` alongside `"object"` and `"category"` in `select_dtypes`.
* **Tests:** Verified all standalone encoder unit tests pass successfully.


phoeenniixx (changes requested):
I dont understand this PR, how is it different from #2082 ? We already have `NanLabelEncoder`, we can use that to perform all this on the complete df no? And `NanLabelEncoder` has better `torch` support. I am not able to understand why do we need to implement a new label encoder which doesnt even support `torch`?

And if we even want to add this support, why not add this as an extension to `NanLabelEncoder` ?

Can you please explain why do we need this and why not do this in #2082 (or in a child branch of #2082)?

Please dont overuse AI

Muhammad-Rebaal:
> I dont understand this PR, how is it different from #2082 ? We already have `NanLabelEncoder`, we can use that to perform all this on the complete df no? And `NanLabelEncoder` has better `torch` support. I am not able to understand why do we need to implement a new label encoder which doesnt even support `torch`?
> 
> And if we even want to add this support, why not add this as an extension to `NanLabelEncoder` ?
> 
> Can you please explain why do we need this and why not do this in #2082 (or in a child branch of #2082)?
> 
> Please dont overuse AI

I just adjusted a couple of lines and misread the previous code as it isn't an extension of `NanLabelEncoder`. I apologize for the trouble and the oversight on my part. I will close this issue & PR now and I'll take care next time.

Issue #2304, opened 2026-06-08
[ENH] Umbrella Issue for Documentation for `pytorch-forecasting` v2


phoeenniixx:
This is the umbrella issue for the `pytorch-forecasting` v2 documentation. 
We need to perform a complete migration of docs from v1 to v2 while still maintaining v1 documentation until it is completely deprecated

Issue #2288, opened 2026-05-22
[ENH] Create migration guide for v1 and v2


phoeenniixx:
We should create a migration guide for v1 to v2 
- Add what things to change and what things are same
- New models added
- etc

PR #2349, opened 2026-07-21
[DOC] Add a v1 to v2 migration guide

echo-xiao:
Adds docs/source/migration_v2.rst (a v1 to v2 migration guide) and registers it in the api_v2 toctree. Addresses #2288, part of the v2 docs umbrella #2304. Includes a v1-to-v2 concept mapping table, an end-to-end TFT before/after example (v1, v2 package wrapper, v2 explicit layers), a what-stays-the-same section, and a whats-new section with an auto-generated model-overview table. Example code runs on load_toydata and the docs build cleanly.


docs/source/api_v2.rst
```diff
@@ -52,3 +52,6 @@ Join the discussion and track our progress on GitHub:
     Metrics <metrics>
     Utils <utils>
     Tutorials <tutorials_v2>
+
+.. seealso::
+    Porting a v1 model implementation to v2? See :doc:`migration_v1_to_v2`.
```
phoeenniixx: this guide should be in the boxes at the top of every page which shows v2 is in works. WHat do you think?

PR #2355, opened 2026-07-26
[ENH] Add v2 implementation of DecoderMLP

### LINKED ISSUE #2076: Add v2 support for MLP
## Description

This issue tracks the migration of the `DecoderMLP` model to the new version 2 (v2) interface of pytorch-forecasting.

Currently, `DecoderMLP` exists only in the legacy v1 implementation (based on `BaseModelWithCovariates`) and does not have a native v2 implementation. A `_decodermlp_v2.py` file is missing, and therefore the model cannot be used directly within the new metadata-driven v2 pipeline and sktime integration framework.

---

## Proposed Changes

- Implement `DecoderMLP` inheriting from the v2 `BaseModel` in `pytorch_forecasting/models/mlp/_decodermlp_v2.py`.
- Follow the structure and conventions of existing v2 models (e.g., DLinear and TFT as described in [ENH] Work Items for ptf-v2 #1974).
- Port the core MLP architecture logic from the existing v1 implementation (`_decodermlp.py`) to the new v2 structure.
- Adapt the forward pass to consume standardized v2 input dictionaries instead of the legacy `TimeSeriesDataSet`-coupled inputs.
- Ensure compatibility with v2 loss handling and prediction output format (`{"prediction": ...}`).
- Add necessary tests in `tests/test_models/test_decodermlp_v2.py`.

---

## Checklist

- [ ] Create `_decodermlp_v2.py` in `pytorch_forecasting/models/mlp/`.
- [ ] Implement forward, step, and configuration methods compatible with the v2 API.
- [ ] Add unit tests for the v2 implementation.
- [ ] Verify backward compatibility or document breaking changes if any.

---

Please consider assigning this issue to me.``

echo-xiao:
#### Reference Issues/PRs
Closes #2076

Adds a v2 implementation of `DecoderMLP` as `DecoderMLP_v2`.

- Categoricals concatenated as numeric
- Single target
- No inverse transform inside `forward` 
- `QuantileLoss` → `n_quantiles` per step, otherwise 1
- All dims derived from datamodule metadata
- no overridden training/test steps

- Whether the decoder-only + static input handling is what you'd expect for v2

- Covered by the v2 test framework via `get_test_train_params`
- no separate test file

#### PR checklist
- [x] Title starts with [ENH]
- [x] Added/updated tests (via framework params) + docs API reference
- [x] Ran pre-commit hooks

### DIFF
--- a/pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
+++ b/pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
@@ -10,6 +10,9 @@ class DecoderMLP_pkg_v2(Base_pkg):
         "info:name": "DecoderMLP_v2",
         "info:compute": 1,
         "authors": ["jdb78", "echo-xiao"],
+        # TODO: the v2 datamodule supports categorical inputs but not
+        # categorical targets yet; add "categorical" to y_type once
+        # EncoderDecoderTimeSeriesDataModule supports categorical targets.
         "info:y_type": ["numeric"],
         "capability:exogenous": True,
         "capability:multivariate": False,
--- a/pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
+++ b/pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
@@ -1,0 +1,78 @@
+"""DecoderMLP v2 package container."""
+
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class DecoderMLP_pkg_v2(Base_pkg):
+    """DecoderMLP v2 package container."""
+
+    _tags = {
+        "info:name": "DecoderMLP_v2",
+        "info:compute": 1,
+        "authors": ["echo-xiao"],
+        "info:y_type": ["numeric"],
+        "capability:exogenous": True,
+        "capability:multivariate": False,
+        "capability:pred_int": True,
+        "capability:flexible_history_length": False,
+        "capability:cold_start": False,
+    }
+
+    @classmethod
+    def get_cls(cls):
+        """Get model class."""
+        from pytorch_forecasting.models.mlp._decodermlp_v2 import DecoderMLP_v2
+
+        return DecoderMLP_v2
+
+    @classmethod
+    def get_datamodule_cls(cls):
+        """Get the underlying DataModule class."""
+        from pytorch_forecasting.data.data_module import (
+            EncoderDecoderTimeSeriesDataModule,
+        )
+
+        return EncoderDecoderTimeSeriesDataModule
+
+    @classmethod
+    def get_test_train_params(cls):
+        """Return testing parameter settings for the trainer.
+
+        Returns
+        -------
+        params : list of dict
+            Parameters to create testing instances of the class. Each dict is passed
+            as ``model_cfg`` to the package constructor; the ``"datamodule_cfg"`` key
+            is forwarded to the datamodule constructor.
+        """
+        from pytorch_forecasting.metrics import MAE, RMSE, SMAPE, QuantileLoss
+
+        params = [
+            {},
+            dict(
+                hidden_size=64, n_hidden_layers=2, dropout=0.1, norm=True, loss=RMSE()
+            ),
+            dict(
+                hidden_size=128,
+                n_hidden_layers=1,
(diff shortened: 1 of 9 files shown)


pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
```diff
@@ -0,0 +1,78 @@
...
+from pytorch_forecasting.base._base_pkg import Base_pkg
+
+
+class DecoderMLP_pkg_v2(Base_pkg):
+    """DecoderMLP v2 package container."""
+
+    _tags = {
+        "info:name": "DecoderMLP_v2",
+        "info:compute": 1,
+        "authors": ["echo-xiao"],
```
phoeenniixx: i think we should add the authors from v1as well here,  if the code is very much common between the versions

pytorch_forecasting/models/mlp/_decodermlp_v2.py
```diff
@@ -0,0 +1,171 @@
...
+# experimental, please use with care.
+########################################################################################
+
+import torch
+import torch.nn as nn
+from torch.optim import Optimizer
+
+from pytorch_forecasting.metrics import QuantileLoss
+from pytorch_forecasting.models.base._base_model_v2 import BaseModel
+from pytorch_forecasting.models.mlp.submodules import FullyConnectedModule
```
phoeenniixx: is this a good idea to move the modules to `layers` ? Maybe this PR could do that!

echo-xiao: 1. added the v1 author to the v2 package's authors
2. moved FullyConnectedModule into layers, following the N-BEATS pattern

pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class DecoderMLP_pkg_v2(Base_pkg):
+    """DecoderMLP v2 package container."""
+
+    _tags = {
+        "info:name": "DecoderMLP_v2",
+        "info:compute": 1,
+        "authors": ["jdb78", "echo-xiao"],
+        "info:y_type": ["numeric"],
```
phoeenniixx, line 16: based on [v1](https://github.com/sktime/pytorch-forecasting/blob/e7c7965cc08a52a566eb24661f31c06b49c88395/pytorch_forecasting/models/mlp/_decodermlp_pkg.py#L13), it also supports categorical data

phoeenniixx (dismissed):
I think this is ready!
just one small doubt

pytorch_forecasting/models/mlp/submodules.py
```diff
@@ -1,52 +1,8 @@
...
-            )
-            if dropout is not None:
-                module_list.append(nn.Dropout(dropout))
-            if norm:
-                module_list.append(nn.LayerNorm(hidden_size))
-        # output layer
-        module_list.append(nn.Linear(hidden_size, output_size))
-
-        self.sequential = nn.Sequential(*module_list)
+# TODO v2: remove this file.
```
phoeenniixx: can you make it v2.0.0 instead (to mention exact minor and patch as well) will be useful while releasing!

pytorch_forecasting/models/mlp/_decodermlp_pkg_v2.py
echo-xiao: yes, v1 support categorical data. but it seems like v2 datamodule support only categorical input not categorical targets. am i right?

if the metadata doesn't support categorical target, i will have to modify shared datamodule file which is EncoderDecoderTimeSeriesDataModule. is my understanding correct?

phoeenniixx: Yep that is true, v2 doesn't support categorical target. We should add a comment here so that we can revisit once it does and we don't miss any coverage?

echo-xiao: done

- changed v2 to v2.0.0
- added todo comment above info:y_type

phoeenniixx (approved):
Thanks a lot!

PR #2356, opened 2026-07-27
[ENH] Add multiloss support for v2 by adding one more dimension to `x` and `y`

### LINKED ISSUE #2308: [ENH] Add `MultiLoss` and multi-target support to `ptf-v2`

phoeenniixx:
This PR adds support for MultiLoss to ptf v2

Fixes https://github.com/sktime/pytorch-forecasting/issues/2308

### Changes
- [x] Add `weights` to the `collate_fn` and `__getitem__` (of private dataset class of datamodule) return for `MultiLoss`. The return would be `(x, (y, weight))` in place of `(x,y)`.
    Done to:
    - [x] `EncoderDecoderTimeSeriesDataModule`
    - [x] `TslibDataModule`
- [ ]  Add support to `BaseModel`. 
- [ ] Add tests and update the test framework - maybe should happen in a new PR instead - along with fixes if needed?

### DIFF
--- a/pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
+++ b/pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
@@ -377,21 +377,24 @@ class EncoderDecoderTimeSeriesDataModule(LightningDataModule):
 
     def _coerce_sample(
         self, sample: dict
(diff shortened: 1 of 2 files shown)


phoeenniixx:
tests will fail as i still need to add support to `BaseModel`

PR #2357, opened 2026-07-27
[BUG] Fix quantile support in `Samformer`

Faakhir30:
#### Reference Issues/PRs

Refs #https://github.com/sktime/pytorch-forecasting/pull/2340#discussion_r3636729938

`Samformer`:

```python
# blocks of code from samformer
        self.linear_forecaster = nn.Linear(
            self.max_encoder_length, self.max_prediction_length
        )  # noqa: E501
...
       out = self.linear_forecaster(out)

        out = out.transpose(1, 2)

        target_predictions = out[:, :, -1]  # (batch_size, max_prediction_length)
...
        if self.n_quantiles > 1:
            target_predictions = target_predictions.unsqueeze(-1).expand(
                -1, -1, self.n_quantiles
            )
        elif self.n_quantiles == 1:
            target_predictions = target_predictions.unsqueeze(-1)
        return {"prediction": target_predictions}
```

This doesnt seem right to me, I think its just generating `max_prediction_length` single quantile and `.expand` is just replicating that single quantile to `n_quantile` times

each quantile should have its own parameters and receive its own gradient, allowing the model to learn different quantiles. i.e.,

```python
self.linear_forecaster = nn.Linear(
            self.max_encoder_length, self.max_prediction_length * n_quantiles
        ) 
```

#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/models/samformer/_samformer_v2.py
+++ b/pytorch_forecasting/models/samformer/_samformer_v2.py
@@ -97,9 +97,12 @@ class Samformer(BaseModel):
         self.compute_values = nn.Linear(
             self.max_encoder_length, self.max_encoder_length
         )  # noqa: E501
+        # Emit a separate forecast head per quantile so each has its own
+        # parameters and gradients (unlike tiling a point forecast with expand).
         self.linear_forecaster = nn.Linear(
-            self.max_encoder_length, self.max_prediction_length
-        )  # noqa: E501
+            self.max_encoder_length,
+            self.max_prediction_length * self.n_quantiles,
+        )
 
     def _scaled_dot_product_attention(
         self,
@@ -169,18 +172,15 @@ class Samformer(BaseModel):
 
         out = x_norm + att_score
         out = self.linear_forecaster(out)
+        # (batch, channels, prediction_length * n_quantiles)
+        batch_size, n_channels, _ = out.shape
+        out = out.view(
+            batch_size,
+            n_channels,
+            self.max_prediction_length,
+            self.n_quantiles,
+        )
+        # Target is the last channel: (batch, prediction_length, n_quantiles)
+        target_predictions = out[:, -1, :, :]
 
-        out = out.transpose(1, 2)
-
-        target_predictions = out[:, :, -1]  # (batch_size, max_prediction_length)
-
-        if target_predictions.ndim == 1:
-            target_predictions = target_predictions.unsqueeze(0)
-
-        if self.n_quantiles > 1:
-            target_predictions = target_predictions.unsqueeze(-1).expand(
-                -1, -1, self.n_quantiles
-            )
-        elif self.n_quantiles == 1:
-            target_predictions = target_predictions.unsqueeze(-1)
         return {"prediction": target_predictions}


fkiraly (changes requested):
I am confused - I think your initial claim (the models do not support quantiles) is incorrect.

If a quantile loss is used at the last/output layer, then the predictions indeed become quantile predictions.

This is because the argmin of a quantile loss expectation (true distribution vs variable in the armin) is the (generative) quantile. Similarly, if you take multiple quantiles in the quantile loss, you get multiple quantiles in the forecast.

fkiraly:
> These model arcitectures are not "quantile-aware".

All that needs to be done for that is to use a quantile loss (or a distributional loss such as log-loss) in the output layer.

Which, as far as I can see in the current implementation, is exactly what happens?

fkiraly:
Please correct me explicitly though if you think I am wrong.

phoeenniixx:
I think this is my mistake @fkiraly. See this comment: https://github.com/sktime/pytorch-forecasting/pull/2340#discussion_r3631703695

I had a misconception that for quantile support, just using loss at the output layer is not enough, it should be used somehow in the deeper layers as well. My bad

phoeenniixx:
I thought in the model, the gradients will be diluted by the time they reach the inner layers and it is not necessary that it is completely aware of the quantiles. But now when I think in more detail, it is an issue with extremely deep networks (where we might need to use auxiliary heads to compute the loss in the middle of the layers as well). I didnt think of it by the chain rule and realised these are not that deep models. And by thinking that some models just support point pred losses, It made me feel that my thinking is right. But now I understand, it is just the case of the output layer not supporting the loss

phoeenniixx:
> * Increasing dimention by `output_dim = self.prediction_length * self.n_quantiles` of linear head doesnt make model quantile-aware

This was my thinking :)

Faakhir30:
> Increasing dimention by output_dim = self.prediction_length * self.n_quantiles of linear head doesnt make model quantile-aware

Alright, now that we have agreed that above statement is false, I think `timexer` and `Dlinear` were already having correct support, reverted those.
I just have one final comment regarding `Samformer`:

```python
# blocks of code from samformer
        self.linear_forecaster = nn.Linear(
            self.max_encoder_length, self.max_prediction_length
        )  # noqa: E501
...
       out = self.linear_forecaster(out)

        out = out.transpose(1, 2)

        target_predictions = out[:, :, -1]  # (batch_size, max_prediction_length)
...
        if self.n_quantiles > 1:
            target_predictions = target_predictions.unsqueeze(-1).expand(
                -1, -1, self.n_quantiles
            )
        elif self.n_quantiles == 1:
            target_predictions = target_predictions.unsqueeze(-1)
        return {"prediction": target_predictions}
```

This still doesnt seem right to me, I think its just generating `max_prediction_length` single quantile and `.expand` is just replicating that single quantile to `n_quantile` times

each quantile should have its own parameters and receive its own gradient, allowing the model to learn different quantiles. i.e.,

```python
self.linear_forecaster = nn.Linear(
            self.max_encoder_length, self.max_prediction_length * n_quantiles
        )

fkiraly:
> I thought in the model, the gradients will be diluted by the time they reach the inner layers and it is not necessary that it is completely aware of the quantiles. But now when I think in more detail, it is an issue with extremely deep networks (where we might need to use auxiliary heads to compute the loss in the middle of the layers as well). I didnt think of it by the chain rule and realised these are not that deep models. And by thinking that some models just support point pred losses, It made me feel that my thinking is right. But now I understand, it is just the case of the output layer not supporting the loss

I see - it may make sense to use different networks or at least heads for different quantiles so the model is better. But it also does not feel strictly necessary, more of a "one can try".

However, I think all that matters from the API perspective is whether the last layer reasonably *claims* to produce quantiles (not whether the model is actually good). For this, having a multi-quantile-loss in the last layer is already sufficient - even if it may or may not produces the best quantile predictions that could be obtained.

In this, there is also an idea for an interesting wrapper or operation: split the NN at some point or add heads onto it for different quantiles. Though maybe a lower prio tangent compared to ptf2 release.

phoeenniixx (changes requested):
Thanks!

phoeenniixx (approved):
Thanks! will merge once everything passes

phoeenniixx:
FYI @fkiraly, are you satisfied with this?

PR #2365, opened 2026-08-01
[DOC] Recommend uv and venv for developer setup

### LINKED ISSUE #2364: [DOC] replace `conda` as recommended env manager with `uv` and/or `venv`
The current docs recommend using `conda` as environment manager.

This is a bit outdated, especially how `uv` has gained popularity, and the commercial liabilities that using `conda` incurs.

We should replace the developer docs in this respect - the `sktime` developer docs for setting up a developer install can be mostly copied here.

BLVCK-MAMBA-6:
#### Reference Issues/PRs

Fixes #2364.

Updates the developer environment setup documentation to:

* recommend `uv` as the primary environment manager;
* document virtual environment creation and activation on Unix and Windows;
* retain standard-library `venv` as a fallback;
* document editable installation with the `dev` dependencies;
* show how to install all optional dependencies when required.

The existing Conda package-installation guidance remains unchanged.

* Whether the `uv` and `venv` commands are accurate across supported platforms.
* Whether the distinction between the recommended workflow and fallback is clear.

No automated tests were added because this is a documentation-only change.

The updated RST document was parsed successfully with `docutils`, and `git diff --check` passed.

#### Any other comments?

None.

#### PR checklist

* [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG].
* [ ] Added/modified tests — not applicable because this is documentation-only.
* [x] Used pre-commit hooks when committing.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -131,30 +131,71 @@ Creating a fork and cloning the repository
 Setting up an editable virtual environment
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-1. Set up a new virtual environment. Our instructions will go through the commands to set up a ``conda`` environment which is recommended for ``pytorch-forecasting`` development.
-The process will be similar for ``venv`` or other virtual environment managers.
+Navigate to your local ``pytorch-forecasting`` repository:
 
-  .. warning::
-       Using ``conda`` via one of the commercial distributions such as Anaconda
-       is in general not free for commercial use and may incur significant costs or liabilities.
-       Consider using free distributions and channels for package management,
-       and be aware of applicable terms and conditions.
+.. code-block:: bash
 
-In the ``conda`` terminal:
+    cd pytorch-forecasting
 
-2. Navigate to your local pytorch-forecasting folder, :code:`cd pytorch-forecasting` or similar
+Using ``uv`` (recommended)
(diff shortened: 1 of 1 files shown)


phoeenniixx (changes requested):
Thanks! This is great
I have a doubt though - should we completely remove `conda` installation instructions? We can recommend `uv` but it should not mean that we remove installation instructions for `conda` ?
@fkiraly What do you think?

BLVCK-MAMBA-6:
Yes, I agree that Conda should remain available as an installation option.

This PR does not remove the general Conda installation instructions, the Conda command is still present near the top of `installation.rst`. I only replaced the Conda-first workflow in the developer setup section, based on #2364.

My understanding was that `uv` should become the recommended environment manager for contributors, while Conda remains available as an alternative installation method. I’ll wait for @fkiraly’s view before making any further changes, and I can make this distinction more explicit in the documentation if needed.

BLVCK-MAMBA-6:
I’ve now made the distinction explicit in the documentation: Conda remains documented as an installation option for the published package, while `uv` is recommended specifically for setting up a contributor development environment.

BLVCK-MAMBA-6:
Hi @phoeenniixx, just following up on this. I clarified the Conda/uv distinction in `072d2ec`: Conda remains available for package installation, while `uv` is recommended specifically for contributor development environments.

All checks are passing. When you have a chance, could you please take another look? Thanks!

PR #2366, opened 2026-08-01
[DOC] Update tslib v2 tutorial to use the package (pkg) classes

### LINKED ISSUE #2353: [DOC] Update `tslib` tutorial for v2
Currently, the [v2 tutorial](https://github.com/sktime/pytorch-forecasting/blob/main/docs/source/tutorials/tslib_v2_example.ipynb) for `tslib` models is not updated with the use of `pkg` classes and other recent changes in the lib. I think we should update this!

echo-xiao:
#### Reference Issues/PRs

Fixes #2353.

Updates the v2 `tslib` tutorial notebook
(`docs/source/tutorials/tslib_v2_example.ipynb`) to use the high-level
package (`pkg`) API for v2 models, as requested in #2353. The previous
version wired up the `TslibDataModule`, model and Lightning `Trainer` by
hand; the tutorial now drives the whole pipeline through the pkg classes.

Main change:
- Rewrite the tutorial around `TimeXer_pkg_v2` / `DLinear_pkg_v2`: configure
  with three dicts (`model_cfg` / `datamodule_cfg` / `trainer_cfg`) and run
  with `.fit()` / `.predict()`, instead of manually constructing the
  datamodule, model and `Trainer`.
- Import the pkg classes from their public module path
  (`pytorch_forecasting.models.timexer` / `pytorch_forecasting.models.dlinear`).
- Add a DLinear example to show the same pipeline works with another
  tslib model.

Correctness fixes made while updating the notebook:
- Fix the `future_known_feature` typo in the `TimeSeries` definition.
- Fix `nhead` -> `n_heads` (the old name was silently swallowed by
  `**kwargs`, so the model was running with the default number of heads).
- Make `context_length` divisible by `patch_length` to drop the
  "some time steps will not be used" warning.
- Remove `categorical_encoders`, which `TslibDataModule` does not accept.
- Leave `logging_metrics` off the `QuantileLoss` variant: point metrics
  (MAE/SMAPE) cannot be computed on quantile output and raise an assertion
  during `fit`.
- Use `accelerator="cpu"` so the small demo is reproducible on any machine.

Clarifying note added to the notebook:
- TimeXer v2 does not yet consume categorical features, and the datamodule
  does not encode them (every feature is cast to float and sliced by index).
  The `cat` columns are kept in the data definition to show the full API.

- Whether the pkg-based flow is the right canonical example for tslib v2
  models (still an experimental / unstable API, see #1836).
- Whether the note on categorical features accurately reflects current
  TimeXer v2 / `TslibDataModule` behavior.
- Whether covering TimeXer + DLinear is sufficient — these are the only two
  models currently built on `TslibBaseModelV2`.

No — this is a documentation-only change (a tutorial notebook); no library
code is touched. The notebook was executed end-to-end locally and all cells
run without errors. The docs build uses `nbsphinx_execute = "never"`, so the
committed outputs are what render.

#### Any other comments?

While updating the notebook I noticed a stray `print(self.model_cfg)` debug
statement in `pytorch_forecasting/base/_base_pkg.py` that fires on every pkg
construction and leaks into the output (the `{'loss': ...}` lines visible in
the notebook). I kept it out of this PR to keep the change documentation-only;
it could be removed in a separate [MNT] PR.

#### PR checklist

- [x] The PR title starts with [DOC].
- [ ] Added/modified tests — N/A, documentation-only change.
- [ ] Used pre-commit hooks when committing.

### DIFF
--- a/docs/source/tutorials/tslib_v2_example.ipynb
+++ b/docs/source/tutorials/tslib_v2_example.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "662cccd2",
+   "id": "66a17c35",
    "metadata": {},
    "source": [
     "# TSLib for v2 - Example notebook for full pipeline"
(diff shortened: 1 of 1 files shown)


phoeenniixx (changes requested):
Nice!
Few suggestions:
- In v2, you can perform the training in two ways - High level API (using `pkg` class) or skipping the `pkg` class and creating the data modules, models etc yourself. So, it would be great if we dont remove the 2nd way (which was I think already present?) And instead add this new High level API with it. Although, the cells for using `pkg` class should be BEFORE the cells where we dont use the `pkg` class. See the [other](https://github.com/sktime/pytorch-forecasting/blob/main/docs/source/tutorials/ptf_V2_example.ipynb) tutorial to see what I mean!
- For the toydataset, you can use the inbuilt method present in `data/examples` file. See [here](https://github.com/sktime/pytorch-forecasting/blob/8918eac8cd4384cc9ef6c913bfcb08fbee9dcdc3/pytorch_forecasting/data/examples.py#L114).
- For the note about API being unstable, It would be great if that can be highlighted so that, noone misses that!
- Please in the begining of the tutorial, also add a md cell explaining `tslib` models are taken from `thuml` and they are different from the enc-dec models using `EncoderDecoderDataModule`.

Issue #2359, opened 2026-07-29
[ENH] Add `target_scale` support to `EncoderDecoderDataModule`


phoeenniixx:
Currently `target_scale` is more of a placeholder in the datamodules (both datamodules, but for `tslib` dm, we first need to fix #2330 )

For `EncoderDecoderDataModule`, we should add this support as we have scalers and normalizers implemented in the datamodule already.

PR #2368, opened 2026-08-03
[ENH] Add preprocessing (scalers and normalizers) to tslib datamodule

### LINKED ISSUE #2330: [ENH] Add preprocessing to `tslib` datamodule
Currently, we dont have preprocessing support in `tslib` datamodule. We should add that!
We could add the scalers and normalizers in a similar way as in #2302, but before that, I think we should 
look at  [thuml's](https://github.com/thuml/Time-Series-Library) implementation of pre-processing steps. We should add the implementation inspired from that package (as `tslib` datamodule is inspired from that package).

### How should we do this?
- First we should look at the [Time-Series-Library](https://github.com/thuml/Time-Series-Library), and then write down the how the preprocessing steps (like scaler and normalizer use) is implemented there
- Try to replicate that in `tslib` data module while staying coherent with the v2 design.

echo-xiao:
#### Reference Issues/PRs

Fixes #2330.

`TslibDataModule` accepted `scalers` / `target_normalizer` but never used them
(stored and ignored). This makes them functional, following #2302 and reusing
`ScalerAdapter`:

- `__init__` wraps them in `ScalerAdapter`, adds `_*_fitted` flags and a
  per-series `_preprocess_cache`.
- `_fit_scalers` / `_fit_target_normalizer` fit on the **train split only**.
- `_normalize_features` / `_normalize_target` transform per series (no-op until
  fitted; only configured continuous columns).
- `_preprocess_data` applies the transforms and caches per series;
  `setup("fit")` fits on `_train_indices`.

1. no deepcopy of scalers
2. deferring `target_scale`** to a follow-up 
3. only continuous features are scaled, the models don't consume them yet
4. target_scale, collate_fn and the models are untouched

Yes — 7 tests in `pytorch_forecasting/data/tests/test_tslib_data_module.py`:
- adapter wrapping + flags; 
- fit-on-train (StandardScaler → ~0 mean / ~1 std);
- target-normalizer fit flag; 
- feature transform (only configured columns);
- target no-op-until-fitted; `_preprocess_data` scaling + cache identity;
- `setup("fit")` produces scaled samples;
- a no-scaler backward-compat guard (byte-identical output, no `target_scale`).

#### PR checklist

- [x] The PR title starts with [ENH].
- [x] Added/modified tests.
- [x] Used pre-commit hooks (`pre-commit run --files` on both changed files, clean).


echo-xiao:
hi, @phoeenniixx please review these and let me know if these decisions are right:
- no deepcopy of scalers
- deferring target_scale to a follow-up, collate_fn and the models are untouched
- only continuous features are scaled, the models don't consume them yet 

TODO:
- lift the shared preprocessing into the base datamodule
- implement target_scale for tslib (sibling of #2359)

pytorch_forecasting/data/tests/test_tslib_data_module.py
```diff
@@ -1,12 +1,39 @@
...
 import torch
 
+from pytorch_forecasting.adapters import ScalerAdapter
 from pytorch_forecasting.data.data_module import TslibDataModule
+from pytorch_forecasting.data.encoders import TorchNormalizer
 from pytorch_forecasting.data.timeseries import TimeSeries
 
 
+def _make_ts(n_series: int = 20, length: int = 40, offset: float = 100.0) -> TimeSeries:
+    """合成数据集:连续特征 ``x`` 远离 0(~offset),便于看出标准化;目标 ``y`` 是正弦。"""
```
phoeenniixx, line 14: please use english docstrings

phoeenniixx (changes requested):
Nice!
I have a few questions:
- Did you look at the `thuml` [implementation](https://github.com/thuml/Time-Series-Library)? I think we should try to keep it as close to this package as possible - when it comes to the preprocessing logic as this dm is based on this package
   - Currently it supports all the scalers and normalizers that `EncoderDecoderDataModule` supports, but is this true for the `tslib` implementation as well?

Also, please use english docstrings, to maintain uniformity across the documentation.

echo-xiao:
> Nice! I have a few questions:
> 
> * Did you look at the `thuml` [implementation](https://github.com/thuml/Time-Series-Library)? I think we should try to keep it as close to this package as possible - when it comes to the preprocessing logic as this dm is based on this package
>   
>   * Currently it supports all the scalers and normalizers that `EncoderDecoderDataModule` supports, but is this true for the `tslib` implementation as well?
> 
> Also, please use english docstrings, to maintain uniformity across the documentation.

In thuml's data layer, it's preprocessing is:
- one hardcoded sklearn standardscaler. no other way to pass a custom scaler
- features and target share the one scaler, there is no seperate target_normalizer
- it is fit strictly on the train segment. 
- inverse_transform is exposed on the datasets. metrics are computed in the original scale.
- per-window normalisation exists but lives in the model, not in the data layer. the normalizer in usa.py is per-sample and classification-only.

tslib does not support EncoderDecoderDataModule supports. 
- list/tuple of normalizers
- multivariate target + one normalizer
- groupnormalizer
- encodernormalizer

what this pr does:
- scalers are now really fits on the training split only. the split is computed once and cached, scalers are fit once, and fitting happens for every stage. previously setup() redrew the split on each call, so trainer.test() saw a split the scalers had already been fit on.
- only sklearn standardscaler is accepted. anything else raises not implemented error. 
- target normalization is off by default. 
- the tests now run in CI

docs/source/tutorials/tslib_v2_example.ipynb
```diff
@@ -304,7 +301,7 @@
     "    context_length=30,\n",
     "    prediction_length=1,\n",
     "    add_relative_time_idx=True,\n",
-    "    target_normalizer=TorchNormalizer(),\n",
+    "    target_normalizer=StandardScaler(),\n",
```
phoeenniixx, line 304: will `TorchNormalizer` not work?

phoeenniixx (changes requested):
I have a  doubt: why are we pinning `scalers`, `target_normalizers` to `StandardScaler` only? Is this how it happens in `thuml`? Would it be better if we relax it a bit to allow the same amount of normalizers, scalers etc as of `EncoderDecoderTimeSeriesDataModule`?

echo-xiao:
> * EncoderDecoderDataModule

Are these right?
1. what we take from thuml is the data shape, the window layout and what the models expect in a batch
2. what we align with EncoderDecoderTimeSeriesDataModule is the configurating surface, which scalers and normalizers a user can pass.
      -  sklearn transformer（StandardScaler / RobustScaler / MinMaxScaler / PowerTransformer / …）
      -  TorchNormalizer
      - EncoderNormalizer
      - GroupNormalizer
      - MultiNormalizer
      - list / tuple of normalizer
      -  NaNLabelEncoder (classified target)
      - "auto"
      - None
      - sktime transformer
3. And eventually extract a shared base class

I think I had applying thum to both, which is where the StandardScaler restraiction came from.

PR #2369, opened 2026-08-05
[BUG] fixes correct passing of `index` in v2

### LINKED ISSUE #2367: [BUG] `index` is not being returned correctly in v2
The time index is not being returned correctly in v2. For some reason `tslib` datamodule even comments this out 
https://github.com/sktime/pytorch-forecasting/blob/8918eac8cd4384cc9ef6c913bfcb08fbee9dcdc3/pytorch_forecasting/data/data_module/_tslib_data_module.py#L195-L196

We should diagnose exactly why this is, and how we can actually correct this index return.

FYI @PranavBhatP 

**NOTE:** This is a hard issue, so please comment with your proposed changes before raising a PR

phoeenniixx:
Fixes #2367 

Adds a new `prediction_start_idx` to the `collate_fn` that is returned when `return_info` has `index`.

### DIFF
--- a/docs/source/tutorials/ptf_V2_example.ipynb
+++ b/docs/source/tutorials/ptf_V2_example.ipynb
@@ -57,38 +57,59 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
    "metadata": {
-    "id": "RkgOT4kiy_RU"
+    "id": "RkgOT4kiy_RU",
+    "ExecuteTime": {
+     "end_time": "2026-08-05T07:39:12.499333648Z",
+     "start_time": "2026-08-05T07:39:12.347281372Z"
+    }
    },
-   "outputs": [],
    "source": [
     "from pytorch_forecasting.data.examples import load_toydata"
-   ]
+   ],
+   "outputs": [],
+   "execution_count": 18
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
    "metadata": {
     "colab": {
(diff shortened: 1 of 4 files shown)


phoeenniixx:
Has to remove `index` from the cell in the tutorial as previously it was just printing `y[1]` (a random placeholder, see [here](https://github.com/sktime/pytorch-forecasting/blob/5a35991c1fe7e569288ac20e96e28d30cbe17ca1/pytorch_forecasting/callbacks/predict.py#L84)). We still need to add some thing like `x_to_index` of [v1](https://github.com/sktime/pytorch-forecasting/blob/5a35991c1fe7e569288ac20e96e28d30cbe17ca1/pytorch_forecasting/data/timeseries/_timeseries.py#L2666), to make this index actually useful. But for now, it is similar to v1, matching `group` with `decoder_idx` (see `x_to_index`).

But this breaks what the cell printed before, as now `preds["index"]` is a 2d tensor and not 1d. So, i just removed it, it makes no sense there.

phoeenniixx:
The big diff in the tutorial is because I had to rerun the whole nb locally to update the outputs :)

fkiraly (changes requested):
The nb looks substantially different, can you kindly explain?

phoeenniixx:
> The nb looks substantially different, can you kindly explain?

As the nb had wrong indexes printed (which were just `y[1]`), I had to rerun the whole nb to remove that wrong values.

PR #2371, opened 2026-08-06
[ENH] optimize metric state updates to prevent O(N^2) reallocation

andersendsa:
Initialize states with empty lists instead of dummy tensors for reduction='none'. Append to list during `update` and only do a single `torch.cat` inside `compute`. This dramatically speeds up long evaluation loops for multi-horizon metrics.

#### PR checklist

- [ ] Added/modified tests

### DIFF
--- a/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
+++ b/pytorch_forecasting/metrics/base_metrics/_base_metrics.py
@@ -819,21 +819,21 @@ class MultiHorizonMetric(Metric):
     def __init__(self, reduction: str = "mean", **kwargs) -> None:
         super().__init__(reduction=reduction, **kwargs)
         if reduction == "none":
-            default_losses = default_lengths = []
+            default_losses, default_lengths = [], []
             dist_reduce_fx = "cat"
         else:
-            default_losses = 0.0
-            default_lengths = 0
+            default_losses = torch.tensor(0.0, dtype=torch.float)
+            default_lengths = torch.tensor(0, dtype=torch.long)
             dist_reduce_fx = "sum"
 
         self.add_state(
             "losses",
-            default=torch.tensor(default_losses, dtype=torch.float),
+            default=default_losses,
             dist_reduce_fx=dist_reduce_fx,
         )
         self.add_state(
             "lengths",
-            default=torch.tensor(default_lengths, dtype=torch.long),
+            default=default_lengths,
             dist_reduce_fx=dist_reduce_fx,
         )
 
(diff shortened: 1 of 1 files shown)


phoeenniixx (changes requested):
Thanks! Can you please share an example how this helps?
Here you check if `losses` is a `list`, when does that happen?

pytorch_forecasting/metrics/base_metrics/_base_metrics.py
```diff
@@ -897,12 +897,17 @@ def update(self, y_pred, target):
     def _update_losses_and_lengths(self, losses: torch.Tensor, lengths: torch.Tensor):
         losses = self.mask_losses(losses, lengths)
         if self.reduction == "none":
-            if self.losses.ndim == 0:
-                self.losses = losses
-                self.lengths = lengths
+            if isinstance(self.losses, list):
+                self.losses.append(losses)
```
Shriprasad-P, line 901: [P2] Add a multi-update regression test for this new list-accumulation path. Existing coverage invokes `MAE(reduction="none")` only once via `forward`; the parametrized `test_reduction_modes` leaves the metric at its default `mean`. Nothing asserts that multiple direct `update()` calls are appended and concatenated correctly, or that `reset()` clears the list state. Please add a direct `MAE(reduction="none")` test with uneven batch sizes that checks `compute()` against the expected concatenation, then `reset() -> update() -> compute()`. This is the core changed lifecycle and would catch a regression that assigns rather than appends.

Shriprasad-P (changes requested):
I verified numerical parity, state lifecycle, TorchMetrics list-state behavior, CPU movement, two-rank Gloo synchronization, and the performance improvement. The requested regression coverage below is the only blocking concern.

Shriprasad-P (changes requested):
Could we add a direct regression test for the new multi-update reduction="none" lifecycle here?
The current tests do not appear to cover multiple direct update() calls being accumulated into the list state and then concatenated in compute(), nor do they verify that reset() clears that accumulated state.

A focused MAE(reduction="none") test with uneven batch sizes that checks compute() against the expected concatenated result, followed by reset() -> update() -> compute(), would protect the core behavior introduced by this change.

Shriprasad-P (commented):
## Review

Accumulating `reduction="none"` updates in a list and concatenating once in `compute()` is a clean way to avoid repeated `torch.cat` O(N²) reallocations.

Fallback path when state is already a tensor is sensible for distributed concat behavior. Worth a quick check that Lightning’s state sync still works with list defaults, but the approach looks sound.

PR #2393, opened 2026-08-22
[BUG] Fix y not being normalized when using EncoderNormalizer

### LINKED ISSUE #2360: [BUG] `y` not being normalized when using `EncoderNormalizer` in `EncoderDecoderDataModule`
Currently, in `EncoderDecoderDataModule`, `y` is not normalized if we are using `EncoderNormalizer` as it is fitted per sequence (during `__getitem__`). In this case we only normalize `target_past` and forgot to normalize the decoder indices of the target. 

See here, we just normalize `target_past` and not the decoder indices (or `y`):
https://github.com/sktime/pytorch-forecasting/blob/e7c7965cc08a52a566eb24661f31c06b49c88395/pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py#L653-L658

`y` remains raw:
https://github.com/sktime/pytorch-forecasting/blob/e7c7965cc08a52a566eb24661f31c06b49c88395/pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py#L797-L804

### Solution
We should normalize `y` as well

AmjadAAYD:
Fixes #2360. When using EncoderNormalizer (fit per encoder window, not globally), target_past was normalized at __getitem__ time but the decoder target y was left raw, so the model was trained to predict the raw scale while being fed a normalized encoder input. Fixed by transforming y with the same per-sequence-fitted normalizer used for target_past. For the multi-target case, added a transform_sequence() method on ScalerAdapter mirroring fit_transform_sequence()'s per-column logic (only transform sub-normalizers that are fit_per_sequence, leave the globally-normalized ones untouched) instead of naively calling the top-level transform(), which broke GroupNormalizer sub-adapters that need group columns X. Added a regression test (test_encoder_normalizer_scales_decoder_target) verifying y now matches normalizer.transform() of the raw decoder target. Full test_data_module.py suite passes (43 passed).


phoeenniixx (changes requested):
can you please make sure that code quality is passing?

PR #2398, opened 2026-08-27
[DO NOT MERGE] Experimental Design for new v2 API

phoeenniixx:
This PR tries a new prototype for new design.
Created `_proto` to have the files with major changes and some minor changes (like to `EncoderDecoderDM`) has been made in place.

### NOTE
This is a prototype - not to be merged as here the idea is just to get a minimal working state of the design to test out how it feels. I have done somethings which will not happen in ACTUAL implementation. Like
* `BaseForecaster` is inheriting from `BasePkg` which will not happen in actuality as `BaseFroecaster` will REPLACE `BasePkg`.
* Special nb for this prototype - will replace the current tutorials
* `_proto` folder with `Forecaster`, `TimeSeries_datatype` etc files. which will be present in their designated folders and not in this folder
* `TimeSeries_datatype` will be renamed to `TimeSeries` if we decide to move on with this plan

### What should a reviewer concentrate their feedback on?

Just the notebook and try out the vignettes and play around with it. This PR is  not meant for the "code review" but just for the review of the vignettes and design

Try the notebook on colab (i have added a cell to install the branch as well)

**AI Generated, so it SHOULD NOT BE MERGED**
Only raised to try out the new interface and get feedback. If approved, will be implemented over a stack of PRs :)

### DIFF
--- a/docs/source/tutorials/new_v2_proto.ipynb
+++ b/docs/source/tutorials/new_v2_proto.ipynb
@@ -1,6 +1,22 @@
 {
  "cells": [
   {
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": "Install the branch",
+   "id": "dc4be0556b5e56ff"
+  },
+  {
+   "metadata": {},
+   "cell_type": "code",
+   "outputs": [],
+   "execution_count": null,
+   "source": [
+    "!pip install git+https://github.com/phoeenniixx/pytorch-forecasting.git@ptf-new-design"
+   ],
+   "id": "56363d9fc1c14965"
+  },
+  {
    "metadata": {
     "ExecuteTime": {
      "end_time": "2026-08-31T18:05:34.917923545Z",
--- a/docs/source/tutorials/new_v2_proto.ipynb
+++ b/docs/source/tutorials/new_v2_proto.ipynb
@@ -1,0 +1,1618 @@
+{
+ "cells": [
+  {
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2026-08-31T18:05:34.917923545Z",
+     "start_time": "2026-08-31T18:05:30.822063696Z"
+    }
+   },
+   "cell_type": "code",
+   "source": [
+    "from pytorch_forecasting.data.examples import load_toydata"
+   ],
+   "id": "3368c59d3c5a058c",
+   "outputs": [],
+   "execution_count": 1
+  },
+  {
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2026-08-31T18:05:34.981494335Z",
+     "start_time": "2026-08-31T18:05:34.919751851Z"
+    }
+   },
+   "cell_type": "code",
+   "source": [
+    "num_series = 100  # Number of individual time series to generate\n",
+    "seq_length = 50  # Length of each time series\n",
+    "data_df = load_toydata(num_series, seq_length)\n",
+    "data_df.head()"
+   ],
+   "id": "56f54a5bf101b4e3",
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "   series_id  time_idx         x         y  category  future_known_feature  \\\n",
+       "0          0         0  0.176004  0.286363         0              1.000000   \n",
+       "1          0         1  0.286363  0.465868         0              0.995004   \n",
+       "2          0         2  0.465868  0.571678         0              0.980067   \n",
+       "3          0         3  0.571678  0.586340         0              0.955336   \n",
+       "4          0         4  0.586340  0.856573         0              0.921061   \n",
+       "\n",
+       "   static_feature  static_feature_cat  \n",
+       "0        0.000406                   0  \n",
+       "1        0.000406                   0  \n",
+       "2        0.000406                   0  \n",
+       "3        0.000406                   0  \n",
+       "4        0.000406                   0  "
+      ],
+      "text/html": [
+       "<div>\n",
+       "<style scoped>\n",
+       "    .dataframe tbody tr th:only-of-type {\n",
+       "        vertical-align: middle;\n",
+       "    }\n",
+       "\n",
+       "    .dataframe tbody tr th {\n",
+       "        vertical-align: top;\n",
+       "    }\n",
+       "\n",
+       "    .dataframe thead th {\n",
+       "        text-align: right;\n",
+       "    }\n",
+       "</style>\n",
+       "<table border=\"1\" class=\"dataframe\">\n",
+       "  <thead>\n",
+       "    <tr style=\"text-align: right;\">\n",
+       "      <th></th>\n",
+       "      <th>series_id</th>\n",
+       "      <th>time_idx</th>\n",
+       "      <th>x</th>\n",
+       "      <th>y</th>\n",
+       "      <th>category</th>\n",
+       "      <th>future_known_feature</th>\n",
+       "      <th>static_feature</th>\n",
+       "      <th>static_feature_cat</th>\n",
+       "    </tr>\n",
+       "  </thead>\n",
+       "  <tbody>\n",
+       "    <tr>\n",
+       "      <th>0</th>\n",
+       "      <td>0</td>\n",
+       "      <td>0</td>\n",
+       "      <td>0.176004</td>\n",
+       "      <td>0.286363</td>\n",
+       "      <td>0</td>\n",
+       "      <td>1.000000</td>\n",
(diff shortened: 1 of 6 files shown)


docs/source/tutorials/new_v2_proto.ipynb
```diff
@@ -0,0 +1,1634 @@
...
+   "source": "Install the branch",
+   "id": "dc4be0556b5e56ff"
+  },
+  {
+   "metadata": {},
+   "cell_type": "code",
+   "outputs": [],
+   "execution_count": null,
+   "source": [
+    "!pip install git+https://github.com/phoeenniixx/pytorch-forecasting.git@ptf-new-design"
```
fkiraly, line 15: general best practice suggestion: jupyter notebooks should not modify the runtime environment...

phoeenniixx: yes, this was intentionally added so that people can run this notebook direclty in colab as well if they want to. As we are not going to merge this, I thought it provides the reviewer some flexibility (not setting up ptf in their env) and then pulling the branch.

fkiraly (changes requested):
Very nice! I think this is great!

I have one question though:
* why is `TimeSeries` no longer a `Dataset`? Is there anything that forces us to move away from standard APIs resp classes? I do not see anything that we could not do with a `dataset`

echo-xiao:
i think the layered architecture is great. and I found a missing part that could be a small problem. 

i think the trainer lifecycle and repeated-fit contract are undefined. i used a second fit with the same trainer, it silently yields an untrained model marked as fitted. here is the code:

```python
from lightning.pytorch import Trainer
from pytorch_forecasting.data.examples import load_toydata
from pytorch_forecasting._proto._tft_forecaster import TFTForecaster
from pytorch_forecasting._proto._timeseries_datatype import TimeSeries_datatype

df = load_toydata(20, 50)
data_january = TimeSeries_datatype(
    df[df["series_id"] < 10], time="time_idx", target="y", group=["series_id"]
)
data_february = TimeSeries_datatype(
    df[df["series_id"] >= 10], time="time_idx", target="y", group=["series_id"]
)

trainer = Trainer(max_epochs=1, logger=False, enable_checkpointing=False)
forecaster = TFTForecaster(hidden_size=16, num_layers=1, attention_head_size=2,
                           trainer=trainer)

forecaster.fit(data_january)
step_after_january = trainer.global_step
model_after_january = forecaster.model_

forecaster.fit(data_february)   # returns without error
step_after_february = trainer.global_step
model_after_february = forecaster.model_

print("optimization steps in second fit:", step_after_february - step_after_january)
print("same model object across fits:   ", model_after_january is model_after_february)
print("forecaster._is_fitted:           ", forecaster._is_fitted)
``` 

here is my output:
Screenshot 2026-09-21 at 3 47 58 PM

pytorch_forecasting/_proto/_timeseries_datatype.py
```diff
@@ -0,0 +1,459 @@
...
+        self.num = self._num
+        self.cat = self._cat
+        self.known = self._known
+        self.unknown = self._unknown
+        self.static = self._static
+
+    def _infer_columns(self):
+        """Fill in ``time``, ``target``, ``num`` and ``cat`` from the data frame.
+
+        Applies the defaults promised by the docstring: target is the last
```
CloseChoice, line 178: Tbh I am a fan of 
> Explicit is better than implicit.

from the Python Zen and would therefore require the user to specify `time` and `target`. All others can have `default=None`. Num and cat can be inferred from the dtype if not explicitly provided.

But this is a side discussion and would break code (but that is what major versions are good for).

pytorch_forecasting/_proto/_timeseries_datatype.py
```diff
@@ -0,0 +1,459 @@
...
+                "x": self.feature_cols,
+                "st": self._static,
+            },
+            col_type=col_type,
+            col_known=col_known,
+        )
+
+    def __len__(self) -> int:
+        """Return number of time series in the dataset."""
+        return len(self._group_ids)
```
CloseChoice, line 230: I would assume the length of a dataset is the number of observations. Maybe I should check how Autogluon and darts handle this.

pytorch_forecasting/_proto/_timeseries_datatype.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def __getitem__(self, index: int) -> dict[str, torch.Tensor]:
+        """Get time series data for given index.
+
+        Returns
+        -------
+        t : numpy.ndarray of shape (n_timepoints,)
+            Time index for each time point in the past or present. Aligned with `y`,
+            and `x` not ending in `f`.
```
CloseChoice, line 239: not ending in `f`? What does that mean

pytorch_forecasting/_proto/_timeseries_datatype.py
```diff
@@ -0,0 +1,459 @@
...
+                    else self._group
+                )
+                future_mask = self.data_future.groupby(group_arg).groups[group_id]
+                future_data = self.data_future.loc[future_mask]
+            else:
+                future_data = self.data_future
+
+            data_fut_vals = future_data[time].values
+
+            combined_times = np.concatenate([data_vals, data_fut_vals])
```
CloseChoice, line 313: Is this just the numpy version of 

```python
data["source"] = "data"
future_data["source"] = "future_data"
df_complete = pd.concat([data, future_data])
df_complete.duplicated(subset=["time"], keep="first")

data = df_complete[df_complete["source"] == "data"]
future_data = df_complete[df_complete["source"] == "future_data"]

...
```
?

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -215,13 +331,52 @@ def _build_cont_scalers(self):
             if name in self._scalers
         ]
 
+    def with_data(
+        self,
+        data: Union[TimeSeries, "TimeSeries_datatype"],  # noqa: F821
```
CloseChoice, line 336: when should the renaming of `TimeSeries_datatype` -> `TimeSeries` happen.

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
```diff
@@ -295,30 +444,32 @@ def _prepare_metadata(self):
...
+            max_encoder_length=self.max_encoder_length,
+            max_prediction_length=self.max_prediction_length,
+            min_encoder_length=self._min_encoder_length,
+            min_prediction_length=self._min_prediction_length,
         )
 
-        return metadata
-
     @property
     def metadata(self):
```
CloseChoice, line 466: return type annotation would be good.

CloseChoice (changes requested):

pytorch_forecasting/_proto/_timeseries_datatype.py
phoeenniixx: Yeah this is soemthing that we should discuss, i am not very sure which parts should be inferred directly. What are your preferences?

> But this is a side discussion and would break code (but that is what major versions are good for).

Yes. that is why we should settle it now, before we make this new design public?

pytorch_forecasting/_proto/_timeseries_datatype.py
phoeenniixx: Umm, the no of groups are kinda same thing, no? What do you suggest? it should be total no of data points? But in some cases no of groups might define the actual data correctly, mainly because we send the `__getitem__` here currently "per group" - meaning `dataset[idx]` would give `group` at that `idx`

pytorch_forecasting/_proto/_timeseries_datatype.py
phoeenniixx: future

phoeenniixx: I think the docstring is not clear enough, I will update this in #2410. 
Thanks!

pytorch_forecasting/_proto/_timeseries_datatype.py
phoeenniixx: kinda similar i think, we do concatenate past and future, but we also sort them for each unique time stamp

pytorch_forecasting/data/data_module/_encoder_decoder_data_module.py
phoeenniixx: I think this should happen in #2423

Issue #2407, opened 2026-09-03
[ENH] New API design for `pytorch-forecasting` v2


phoeenniixx:
## Background
After some user feedback on the current API of `pytorch-forecasting`, we came to know there are some issues with it. 

### API 

- They hit many errors, and the messages did not say what went wrong or how to fix it
- Errors surfaced late, deep in the pipeline, rather than at the call that took the bad input
- Configs are plain dicts; users coming from `sklearn` and `torch` found this unintuitive
- Constructors accept invalid configs without complaint
- `fit` / `predict` do not check their inputs
- Users found the 3-stage pipeline (not using `pkg` class) easier to follow than the current entry point
- `predict` output and other returned objects are hard to inspect and do not behave as expected

### Docs

- Users could not find a copy-pasteable example of the v2 pipeline
- Docs do not describe the shape of the config dicts
- Users could not tell which parameters `model_cfg` accepts and got stuck there
- Online docs were hard to navigate and examples hard to find

## New v2 API

So, based on this input, we are revamping and making some (a LOT of) changes to the v2 API.
To keep the changes separate from the current version and to not break the current code, **we will be doing it on a separate branch (`v2-dev`) instead of `main`.**

### Changes

See #2398 for prototype.

- Make `TimeSeries` a datatype and not D1 layer - this will be the standardised datatype of `ptf-v2`. This means the input and final preds will be a `TimeSeries` obj.
- The metadata objects will now be dataclasses
- The `pkg` class will be renamed to `forecaster` (like in place of `TFT_pkg_v2`, it will be `TFTForecaster`)
  - The `model_cfg` will be replaced by the flattened params - meaning, the `forecaster` class will NOT accept the `model_cfg`, instead we will flatten the the cfg and the params will be passed directly to the class
  - It will now accept the datamodule and `trainer` class directly (not cfgs)
    - datamodule will be   passed as `__init__` argument
    - `trainer` can be passed as an `__init__` arg as well as as `fit` arg. 
      - Here, the `trainer` passed to `fit` will override the `trainer` passed to `__init__`.

### Vignettes
**Minimal Vignette**
```python
data = TimeSeries(
    df, 
    num=['x'], 
    ...,
) 
forecaster = TFTForecaster()
forecaster.fit(data)
```

**Maximal Vignette**
```python
trainer_init = Trainer(
    max_epochs = 5, 
    accelerator = "auto", 
    ...
)
trainer_fit = Trainer(
[...]

PR #2410, opened 2026-09-06
[ENH] Add `TimeSeries` datatype as per new API

phoeenniixx:
This PR adds the new Add `TimeSeries` datatype and update datamodule with the new design.
See #2407, #2398

- `TimeSeries` is now a standarised datatype. Updated the tests and renamed the file to `test_timeseries_v2` as this is no more a D1 layer!

- Also, we now have `metadata` as a dataclass, instead of the `dict` to make it more user friendly.
Updated tests also updates some old ones (and removes some tests like `init` checks - which are not needed i think?)

This PR is aimed at `v2-dev` instead of `main`, so that we dont break current user code until this design is complete and stable. See #2407 for details. So I had to update the branch target in the workflows for the tests to run (which might fail due to the changes)

## Note
The tutorials will fail for sure, will update the notebook in one of the branchs stacked over this PR.
Still need to add the `Forecaster` and tests - will add the notebook and docs changes on the upcoming branches as the vignette is incomplete without `forecaster` which is not in this PR.

## Changes made
- Changed the test file of `TimeSeries` from `test_d1` tro `test_timeseries_v2`
- updates to `TimeSeries`
   - Now uses `TimeSeriesMetadata` as metadata inplace of a `dict`
   - Not a `Dataset` anymore
   - Added some auto-detection of cols
     - `target` is the last column of `data`.
     - `time` is the first column not already used as `group`, `target`, `static` or `weight`. If no such column exists, `ValueError` is raised - pass `time` explicitly.
     - `num` and `cat` are the dtype split over all columns that are not `group`, `time` or `weight`
   - data validation
   - `to_pandas` to create a df
   - `from_tensors` to create a `TimeSeries` class for predicitons 
     - which lack a lot of metadata, so need special handling otherwise will interfere with the input data validation

### DIFF
--- a/pytorch_forecasting/data/timeseries/_timeseries_v2.py
+++ b/pytorch_forecasting/data/timeseries/_timeseries_v2.py
@@ -78,24 +78,16 @@ class TimeSeries:
 
     Notes
     -----
-    This is a datatype, not a ``torch.utils.data.Dataset``. It implements
-    ``__len__`` and ``__getitem__`` so that a data module can iterate series out
-    of it, but it must not be handed to a ``DataLoader``.
-
     Columns that are not passed are inferred:
 
     * ``target`` is the last column of ``data``.
     * ``time`` is the first column not already used as ``group``, ``target``,
       ``static`` or ``weight``. If no such column exists, ``ValueError``
       is raised - pass ``time`` explicitly.
-    * ``num`` and ``cat`` are the dtype split (``"fi"`` numerical,
-      ``"Obc"`` categorical) over all columns that are not ``group``, ``time``
-      or ``weight``. Each is inferred only if not passed, so passing ``cat=[]``
-      still infers ``num``.
+    * ``num`` and ``cat`` are the dtype split over all columns that are not ``group``,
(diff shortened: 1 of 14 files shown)


pytorch_forecasting/models/samformer/_samformer_v2.py
```diff
@@ -67,7 +67,9 @@ def __init__(
             lr_scheduler_params=lr_scheduler_params,
         )
 
-        self.save_hyperparameters(ignore=["loss", "logging_metrics", "optimizer"])
+        self.save_hyperparameters(
+            ignore=["loss", "logging_metrics", "optimizer", "metadata"]
```
phoeenniixx, line 71: Had to add `metadata` (which is a dataclass now) to ignore as otherwise it causes issue of `UnpicklingError` of `torch`.

pytorch_forecasting/data/timeseries/_timeseries_v2.py
phoeenniixx, line 291: need to make this docstring better and clearer - also check other docstrings which might not be clear enough

PR #2412, opened 2026-09-06
[ENH] Rename `*_v2_pkg.py` to `*_pkg_v2.py`

### LINKED ISSUE #2339: [ENH] Rename `*_v2_pkg.py` to `*_pkg_v2.py`
Rename pkg files to make file names consistent:
- `tide_v2_pkg`
- `samformer_v2_pkg`

very good first issue

dchaudhari7177:
Renames the two files that spelled the suffix the other way round:

```
models/samformer/_samformer_v2_pkg.py          -> _samformer_pkg_v2.py
models/tide/_tide_dsipts/_tide_v2_pkg.py       -> _tide_pkg_v2.py
```

Every other v2 package file already uses `_pkg_v2` — dlinear, frets, decodermlp, patch_tst, scinet, softs, tft, timexer, tsmixer, xlstm — so these two were the only ones left.

**The class names do not change.** `Samformer_pkg_v2` and `TIDE_pkg_v2` are untouched, so nothing that resolves packages by class name moves. Worth noting explicitly because `test_all_estimators_v2.py:122` derives the expected package name from the *class* (`class_name + "_pkg_v2"`), not from the filename — the rename is invisible to it either way.

Updated with the rename:

- `models/samformer/__init__.py`
- `models/samformer/_samformer_v2.py` (a deferred import inside the model)
- `models/tide/_tide_dsipts/__init__.py`
- `docs/source/pkg_v2.rst` (two autosummary entries)

One consequence worth flagging so it doesn't look like unrelated churn: the import order in both `__init__.py` files flips. `_samformer_pkg_v2` sorts *before* `_samformer_v2`, where the old `_samformer_v2_pkg` sorted after it, so ruff's isort rule moves those two lines. That is the whole reason those hunks are in the diff.

```
grep -rn "_samformer_v2_pkg\|_tide_v2_pkg" .   (nothing)
ruff check pytorch_forecasting docs             All checks passed!
```

I could not run the test suite locally — this machine has a numpy/scikit-learn ABI mismatch that breaks `import sklearn` before any project code runs — so the verification above is static plus CI's. Given the change is a file rename with no behavioural surface, I'd rather say that than imply a green suite I did not see.

Fixes #2339


phoeenniixx (approved):
Thanks a lot!

PR #2413, opened 2026-09-06
[BUG] Normalize `y` when using a per-sequence target normalizer

### LINKED ISSUE #2360: [BUG] `y` not being normalized when using `EncoderNormalizer` in `EncoderDecoderDataModule`

dchaudhari7177:
Fixes #2360.

## The bug

`EncoderNormalizer` is fitted per sequence at `__getitem__` time, and only `target_past` was transformed. `y` came straight out of `preprocessed_data["target"]` — which is **raw** for exactly this case, because `_fit_target_normalizer` returns early for a single per-sequence normalizer rather than fitting globally.

The encoder therefore saw normalized values while the loss was computed against unnormalized ones. On a series with mean ≈ 100:

```
before:  target_past mean=0.0000 std=1.0000 | y mean=131.4266 std=2.1250
after:   target_past mean=0.0000 std=1.0000 | y mean=  2.0674 std=0.5471
```

## The fix

Adds `ScalerAdapter.transform_sequence`, the companion to the existing `fit_transform_sequence`, and calls it on `y` behind the same `fit_per_sequence` check.

**`transform`, not `fit_transform` — this is the part worth reviewing.** Refitting on the decoder window would scale the target by statistics of the very values being predicted, which is the leak `EncoderNormalizer` exists to avoid. `transform_sequence` reuses the parameters the encoder window fitted moments earlier in the same `__getitem__` call. I checked that nothing refits the target normalizer in between: the covariate loop touches `_cont_scalers`, not `_target_normalizer`.

**Multi-target mirrors `fit_transform_sequence`.** Only sub-normalizers with `fit_per_sequence` are applied; the rest were already applied during preprocessing, and transforming again would scale them twice.

## Tests

Seven new tests. **Three fail on current `main`** — I reverted the two source files and reran to confirm they bite:

```
FAILED test_encoder_normalizer_normalizes_y
  AssertionError: y looks unnormalized: mean 131.43. Raw targets in this fixture are around 100.

FAILED test_encoder_normalizer_y_matches_the_encoder_fitted_transform
  expected [1.383, 2.183, 2.203, 1.735, 2.833]
  got      [118.835, 121.401, 121.466, 119.964, 123.488]

FAILED test_multivariate_target_normalizes_only_the_per_sequence_column
```

The other four are guards rather than regressions, and are the ones I'd want if someone later "simplifies" this:

- `test_encoder_normalizer_is_not_refitted_on_the_decoder_window` — asserts `y` does **not** have mean 0, which is what a refit would produce. This is the leak check, and it passes on `main` too (raw `y` also isn't centred), so it earns its place only against future changes.
- `test_non_per_sequence_normalizers_are_unchanged[None|TorchNormalizer|GroupNormalizer]` — those normalize during preprocessing, so `y` must stay byte-identical. This is what fails if the new call escapes the `fit_per_sequence` check and starts double-scaling.

The exact-transform test derives the expected values from the module's own raw target and window bounds rather than from a second data module, so there is no RNG or split drift between the two sides.

```
pytest tests/test_data/           184 passed
ruff check pytorch_forecasting tests   All checks passed!
```

## One thing I did not change

`target_scale` is still computed from `target_original`, independently of the normalizer. That looked deliberate (it is the denormalization scale the model reports predictions in), so I left it alone — but if the intent was for it to track the encoder normalizer's parameters when `fit_per_sequence` is set, that would be a second, separate change and I'm happy to open it.


tests/test_data/test_data_module.py
```diff
@@ -726,3 +726,163 @@ def test_group_normalizer_uses_groups():
         mean1 = target1["target"].mean().abs()
         assert mean0 < 1.0, "Group 0 target should be normalized near 0"
         assert mean1 < 1.0, "Group 1 target should be normalized near 0"
+
+
+def _trending_series(n_groups=4, seq_length=60, offset=100.0, slope=0.5):
```
phoeenniixx: can you please explain: why not use the existing series? and datamodule?

phoeenniixx (changes requested):
Thanks!
Can you please simplify the comments and the docstrings in the datamodule and the adapter? That is too verbose, no? Over-explaining things

dchaudhari7177:
Both fair — done.

**Verbosity.** Cut the `transform_sequence` docstring and the data module
comment down to the one thing that isn't obvious from the code: why it's
`transform` and not `fit_transform`. The rest was restating the diff.

**Reusing the fixtures.** Switched all three tests to `sample_timeseries_data`
and dropped `_trending_series`. Its only reason to exist was the offset — raw
targets near 100 so they couldn't be mistaken for normalized ones — but that
was propping up magnitude-threshold assertions like `abs(y.mean()) < 20`.
Comparing `y` against the encoder-fitted transform instead pins the behaviour
exactly, which is both a stronger assertion and scale-independent, so the
offset stopped being necessary. That also let the two threshold tests collapse
into one; net −128 lines in the test file.

`data_module` I kept out of, since it pins no `target_normalizer` and these
tests have to vary it (`EncoderNormalizer`, `TorchNormalizer`,
`GroupNormalizer`, `None`). The remaining `_module` helper is just that one
kwarg, plus `_window_targets` so the three tests read the raw window from the
same place rather than each recomputing the slice.

Verified: all 48 tests in `tests/test_data/test_data_module.py` pass, and
`test_encoder_normalizer_normalizes_y` still fails without the fix (it was
passing vacuously nowhere — reverting the `transform_sequence` call reproduces
it). Formatted with the ruff v0.6.9 the pre-commit config pins.

tests/test_data/test_data_module.py
dchaudhari7177: No good reason, so I switched. All three tests now use the shared `sample_timeseries_data` fixture and its data module, and `_trending_series` is gone (commit "Simplify the comments and docstrings, and reuse the shared fixture"). I explained the details in the conversation tab: https://github.com/sktime/pytorch-forecasting/pull/2413#issuecomment-5747843597

PR #2420, opened 2026-09-09
[ENH] Adding Support of Distribution Loss in v2 Models

### LINKED ISSUE #2389: [ENH] Adding Support of Distribution Loss in v2 Models.
Currently there isn't any support of Distribution Loss in v2 models

The V2 BaseModel `_base_model_v2.py` should detect when the passed loss is a `DistributionLoss`  from metrics/distributions.py and automatically infer `output_size` from `len(loss.distribution_arguments)`, call `loss.rescale_parameters()` in the training loop, and ensure target_scale is passed through from the V2 `DataModule` batch the full `DistributionLoss` contract `map_x_to_distribution`, `to_quantiles`, sample is already implemented and only needs to be wired into the V2 model layer.

Muhammad-Rebaal:
Fixes #2389 

Hi @phoeenniixx, 
In order to add support I've done the following changes :

- Added `transform_output` in `BaseModel` to scale predictions back to their real-world numbers.
- Added `output_size` in `BaseModel` to tell the model how many numbers to predict.
- Added `target_normalizer` to `__init__` in `TslibBaseModel` to pass the data scaler to the model.


phoeenniixx (changes requested):
Can you please add some tests here? To see if the shapes are correct etc? 

Maybe we can update the test framework even here - to loop over the losses (like in  v1). That can happen in a separate PR stacked over this PR.

Muhammad-Rebaal:
Hi @phoeenniixx,
I've added the tests

tests/test_models/test_base_model_v2.py
```diff
@@ -122,3 +122,50 @@ def test_optimizer_instance():
...
+    target_scale = {"center": center, "scale": scale}
+
+    result = model.transform_output(raw, target_scale)
+
+    assert result.shape == raw.shape
+    assert torch.allclose(result[0], raw[0] * 2.0 + 10.0)
+    assert torch.allclose(result[3], raw[3] * 5.0 + 40.0)
+
+
+def test_transform_output_plain_tensor_target_scale():
```
phoeenniixx, line 161: how are these two tests useful? Why not direclty use the `DistributionLoss`?

phoeenniixx (changes requested):
I think we are still lacking the tests that check If `DistributionLoss` works as intended with v2.

Also, this should be present in the test framework? why the base model?

By adding "support" we should also update the models which have this todo, no?

tests/test_models/test_base_model_v2.py
Muhammad-Rebaal: These two tests only validate the affine denormalization path (pred * scale + center), which is the straightforward case.
I'll add those distribution loss tests as well

Muhammad-Rebaal:
> I think we are still lacking the tests that check If `DistributionLoss` works as intended with v2.

Ok I'll add that specific tests as well

> Also, this should be present in the test framework? why the base model?

Yeah I got your point up here so that each time it would run it would also check it .

> By adding "support" we should also update the models which have this todo, no?

I thought for the model we'd open an umbrella issue that would adding support for each. Although If that's your vision to add in all models I'll update that as well.

phoeenniixx:
> I thought for the model we'd open an umbrella issue that would adding support for each. Although If that's your vision to add in all models I'll update that as well.

We should try with atleast 1-2 models otherwise how would we know if this is working or not?

tests/test_models/test_base_model_v2.py
phoeenniixx: My doubt is - `DistributionLoss` does somethign similar no?
(i have not looked at the math of the losses, so i may be wrong)

PR #2423, opened 2026-09-11
&Faakhir30 [ENH] Update Datamodules according to new API design

phoeenniixx:
Stacks on #2410 (and #2313 in coming commits FYI @Faakhir30 )
Also see #2407, #2398

This PR is aimed at `v2-dev` instead of `main`, so that we dont break current user code until this design is complete and stable. See #2407 for details. So I had to update the branch target in the workflows for the tests to run (which might fail due to the changes)

## Note
The tutorials will fail for sure, will update the notebook in one of the branchs stacked over this PR.
Still need to add the `Forecaster` and tests - will add the notebook and docs changes on the upcoming branches as the vignette is incomplete without `forecaster` which is not in this PR.

## Changes
- All metadata dataclasses (like `EncDecDataModuleMetadata`, `TimeSeriesMetadata` etc) are added to `_metadata.py`
- a new util for validation checks in `utils/_validation`
- Update the tests (needed to make some changes like updating the param name) in standalone model tests for v2 as well
   - Also had to add `metadata` in `ignore` of `save_hyperparameters` of `Samformer`. See [this comment](https://github.com/sktime/pytorch-forecasting/pull/2423#discussion_r4045987106)
- Updates to datamodules (currently duplicate in both classes)
  - Uses `EncDecDataModuleMetadata` (in case of `EncoderDecoderTImeSeriesDatamodule`) and tslib dm has its own metadata dataclass.
  - Data validation.
  - `with_data` method to create a new dm instance using the already present params (like already fitted scalers - only to create the transformed data).
     - Like creating a datamodule for prediction where we need the scalers from the datamodule we used for fitting. 
     - Also useful to create the class "without data" which arrives at `forecaster.fit()` while we accept the dm class in `__init__` of the `forecaster`. So, the class is actually just empty at this point without data.

### DIFF
--- a/pytorch_forecasting/models/samformer/_samformer_v2.py
+++ b/pytorch_forecasting/models/samformer/_samformer_v2.py
@@ -67,7 +67,9 @@ class Samformer(BaseModel):
             lr_scheduler_params=lr_scheduler_params,
         )
 
-        self.save_hyperparameters(ignore=["loss", "logging_metrics", "optimizer"])
+        self.save_hyperparameters(
+            ignore=["loss", "logging_metrics", "optimizer", "metadata"]
+        )
         self.metadata = metadata
         self.n_quantiles = 1
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,9 @@ name: Test
 
 on:
   push:
-    branches: [main]
+    branches: [v2-dev]
   pull_request:
-    branches: [main]
+    branches: [v2-dev]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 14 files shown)


pytorch_forecasting/models/samformer/_samformer_v2.py
```diff
@@ -67,7 +67,9 @@ def __init__(
             lr_scheduler_params=lr_scheduler_params,
         )
 
-        self.save_hyperparameters(ignore=["loss", "logging_metrics", "optimizer"])
+        self.save_hyperparameters(
+            ignore=["loss", "logging_metrics", "optimizer", "metadata"]
```
phoeenniixx, line 71: need to add `metadata` to be ignored to prevent the `torch` `UnpicklingError` as metadata is now a dataclass

Issue #2377, opened 2026-08-09
[DOC] Add usage examples to all models in v1 and v2


phoeenniixx:
We currently have no usage examples for v1 and v2 models. We should do that!

As v2 is still WiP, I would suggest we start with adding examples to v1.

### v1 models 

- [ ] [`DeepAR`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/deepar/_deepar.py) 
- [ ] [`DecoderMLP` ](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/mlp/_decodermlp.py)
- [ ] [`NBeats` ](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nbeats/_nbeats.py)
- [ ] [`NBeatsKAN`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nbeats/_nbeatskan.py)
- [ ] [`NHiTS`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/nhits/_nhits.py)
- [ ] [`RNN`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/rnn/_rnn.py)
- [ ] [`TFT`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/temporal_fusion_transformer/_tft.py)
- [ ] [`TiDE`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/tide/_tide.py)
- [ ] [`TimeXer`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/timexer/_timexer.py)
- [ ] [`xLSTMTime`](https://github.com/sktime/pytorch-forecasting/blob/main/pytorch_forecasting/models/xlstm/_xlstm.py)

### v2 models
 To be done once the design is finalised

phoeenniixx:
Please add examples to 1-2 models in one PR

ZhangStudyLife:
I’ll add a focused usage example for `DecoderMLP` only. I have checked that it is not covered by #2378 or another open PR.

dinesh-patel45:
Hi, I'd like to contribute to this issue by adding a v1 usage example for the TiDE model. Is TiDE still available to work on?

PR #2425, opened 2026-09-13
[DOC] Add usage example for RecurrentNetwork

shanta0338:
Markdown
#### Reference Issues/PRs
Addresses part of #2377.

Adds a runnable docstring usage example for `RecurrentNetwork` in `pytorch_forecasting/models/rnn/_rnn.py`.
- Tested locally with `python -m doctest pytorch_forecasting/models/rnn/_rnn.py -v` (12 tests passed).
- Configured with `fast_dev_run=True`, CPU accelerator, and silent logging flags to prevent stdout noise during test execution.
- Verified with `pre-commit` hooks.


phoeenniixx (approved):
Thanks!

PR #2430, opened 2026-09-17
[MNT] Pin GitHub Actions to full commit SHAs

### LINKED ISSUE #2399: [MNT] Pin github actions to full commit SHA
https://github.com/sktime/pytorch-forecasting/blob/main/.github/workflows/test.yml#L22

currently we're using mutable tags in our workflow, i think we can pin full commit SHA (as it guards supply chain attacks).

@fkiraly gentle ping, we can have this improvement in all other projects of sktime, creating issue, here for approval, and then i can just submit PR for fix.

And then as a follow up we can implement Zizmor in our workflow https://github.com/zizmorcore/zizmor

Please let me know your thoughts on this. Thanks!

emirhan-karaca:
#### Reference Issues/PRs

Fixes #2399.

The Test, CodeQL, and PyPI Release workflows execute actions through mutable tags. Pin all 26 action references to full commit SHAs using [action-pin v2.0.0](https://github.com/emirhan-karaca/action-pin/tree/v2.0.0), keeping the original tag beside each SHA for review and future updates.

This covers 13 references in `test.yml`, 4 in `codeql.yml`, and 9 in `pypi_release.yml`, including the CodeQL subpath actions and `pypa/gh-action-pypi-publish@release/v1`.

The selected SHAs correspond to the existing refs, verified directly against all eight upstream action repositories with `git ls-remote`. Workflow triggers, jobs, permissions, inputs, scripts, and the existing GitHub Actions Dependabot configuration are preserved.

Validation completed:

- `pre-commit run --files .github/workflows/test.yml .github/workflows/codeql.yml .github/workflows/pypi_release.yml` passes all applicable hooks.
- `action-pin --check --dir .github/workflows` passes; `action-pin --diff --dir .github/workflows` produces no further changes.
- Parsed all three YAML files and compared them with `main`, verifying that only `uses` values changed. Reversing the generated SHA/version annotations reproduces the original text.
- `git diff --check` passes.

No Python application code or tests changed. The full test matrix was not run locally. The [Test](https://github.com/sktime/pytorch-forecasting/actions/runs/35185553127) and [CodeQL](https://github.com/sktime/pytorch-forecasting/actions/runs/35185553129) PR workflows are awaiting maintainer approval.

#### Any other comments?

Zizmor integration remains the separate follow-up proposed in #2399.

#### PR checklist

- [x] The PR title starts with `[MNT]`.
- [x] Ran the configured pre-commit hooks on all changed files.
- [x] Verified every pin against the original upstream ref and checked that workflow behavior is otherwise unchanged.


phoeenniixx (changes requested):
Thanks!
I have a doubt though: Why should we actually pin to SHAs? this will even prevent dependabots to add new updates no?

